PR #1, opened 2022-01-19
Fix EPL version and point to EPL 2.0 directly

fredg02:


akurtakov:
This is not yet the official git repo used. It's a test bed for releng work until we are sure we can switch. Please open gerrit against https://git.eclipse.org/c/platform/eclipse.platform.releng.aggregator.git/ so it's not lost when we sync up.

akurtakov:
Pushed to gerrit as https://git.eclipse.org/r/c/platform/eclipse.platform.releng.aggregator/+/189787

fredg02:
Thanks Alex!

Issue #29, opened 2022-02-17
Update centos 8 docker files to use redhat base image 8


sravanlakkimsetti:
Support for centos has been dropped by Dec 31, 2021. Need to migrate these to use redhat base image version 8.

Code is available here 

1. https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/tree/master/cje-production/dockerfiles/centos-gtk3-metacity/8-gtk3
2. https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/tree/master/cje-production/dockerfiles/centos-gtk3-metacity/8-swtBuild

SDawley:
I created a PR for this that updates to ubi 8.5-226. I included registry.access.redhat.com because when I tested locally it defaulted to registry.redhat.io which needs a login. 

Also @sravanlakkimsetti since this is now UBI not CentOS should I rename the build and push scripts?

PR #33, opened 2022-02-17
Updating centos 8 dockerfiles to use the latest 8.5 universal base image

SDawley:
For [issue 29](https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/issues/29).

### DIFF
--- a/cje-production/dockerfiles/centos-gtk3-metacity/8-gtk3/Dockerfile
+++ b/cje-production/dockerfiles/centos-gtk3-metacity/8-gtk3/Dockerfile
@@ -1,5 +1,5 @@
-ARG FROM_TAG="8"
-FROM centos:${FROM_TAG}
+ARG FROM_TAG="8.5-226"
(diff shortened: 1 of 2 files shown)


sravanlakkimsetti (changes requested):
Can we use https://hub.docker.com/r/redhat/ubi8/tags. At eclipse foundation we may not have access to Red Hat's image repository. Also we will be uploading this to docker.io. So we may have legal issues. 

Can you please try with the image available in docker hub?

sravanlakkimsetti:
I am getting this when I try to build
`Updating Subscription Management repositories.
Unable to read consumer identity

This system is not registered with an entitlement server. You can use subscription-manager to register.
`

sravanlakkimsetti:
I think we can use quay.io/centos/centos:stream8. This should be fine for our work.

sravanlakkimsetti (approved):

PR #47, opened 2022-03-01
fix: update folder names to reflect that centos9 installs mutter not …

SDawley:
…metacity

Since I had to switch to mutter to get them to build.


SDawley:
I only moved the gtk4/centos9 files since the 7 & 8 dockerfiles still install and use metacity, but if you'd like me to switch them all to mutter for consistency I can do that.

sravanlakkimsetti:
I asked because the folder names for changed for centos7 and centos 8 in this commit.  centos-gtk3-metacity got changed to centos-gtk3-mutter. I don't think that is required. Can you please check?

PR #92, opened 2022-03-08
Attempt to add illegal dependency

mickaelistria:


mickaelistria:
@akurtakov Looks like the license-check doesn't check the license for this addition. I suspect it's a regression in Tycho 2.7 when it comes to listing dependencies for an eclipse-repository. I'll investigate further and bring this to Tycho if it's the right issue.

mickaelistria:
So it looks like, unfortunately, `mvn dependency:list` has never worked for eclipse-repository. I've opened https://github.com/eclipse/tycho/issues/744

mickaelistria:
This showcase that -if we use Tycho 3.0.0-SNAPSHOT that better lists dependencies- we get a complete workflow that involves IP validation of dependencies on every PR.
@sravanlakkimsetti @akurtakov I think it's worth using Tycho 3.0.0-SNAPSHOT. Do you agree? If we I can submit a PR.

akurtakov:
I think it's fine to go for 3.0.0-SNAPSHOT once we remove the deprecated packaging types in Tycho as this is probably the biggest removal which I don't want to hurt us cause we forgot to change smth in releng.aggregator build.

sravanlakkimsetti:
> I think it's fine to go for 3.0.0-SNAPSHOT once we remove the deprecated packaging types in Tycho as this is probably the biggest removal which I don't want to hurt us cause we forgot to change smth in releng.aggregator build.

I think we can go with 3.0.0-SNAPSHOT. We will have to move to that in 15 days time when Java 19 development work starts. it might be better if we can move now

akurtakov:
No longer needed as a test.

PR #104, opened 2022-03-14
Sign all repositories (GH#32)

mickaelistria:


mickaelistria:
Let's abandon this one. I think the issue is more that we should just avoid having Platform publishing multiple p2 repositories.

PR #120, opened 2022-03-16
Bug 578004 - Eclipse 4.23 prerequisites

akurtakov:
Test commit to see whether bugzilla linking works


akurtakov:
Not working so let's not keep it to polute the queue.

Issue #102, opened 2022-03-14
Migrate p2 to GitHub


mickaelistria:

mickaelistria:
Opened https://gitlab.eclipse.org/eclipsefdn/helpdesk/-/issues/984

mickaelistria:
Submitted PR #122 to continue the migration

merks:
I think this is done now too then, right?

mickaelistria:
That's not complete yet, some more steps of the workflow defined in #24 needs to be done/verified.
At the moment, i'm waiting for a happy build result from CI to merge the change in aggregator.

mickaelistria:
An infra issue is actually blocking: https://gitlab.eclipse.org/eclipsefdn/helpdesk/-/issues/984#note_616989

mickaelistria:
We can now close, everything is working as expected and the only remaining details are better tracked directly in p2 tracker.

PR #122, opened 2022-03-17
Issue #102 - Migrate p2 to GitHub

mickaelistria:


mickaelistria:
Build failure is not related. The simple fact that everything is fetched and starts resolving is enough to validate this change about submodule remote.

Issue #130, opened 2022-03-18
Update to Jetty 10.0.8


akurtakov:
There will no longer be intermediate p2 repository created by releng but use Maven artifacts directly.

akurtakov:
Build was successful and all jetty bundles show proper pgp signatures in the about dialog.

akurtakov:
Old I-build updated to latest one:
![Screenshot from 2022-03-18 21-54-45](https://user-images.githubusercontent.com/574788/159074700-554b71c9-f8a6-44a2-a790-49dbff456e42.png)

jmcc0nn3ll:
@akurtakov what is the recommended process to update versions of jetty for this? we released 10.0.9 this week

akurtakov:
It's already done via #187 - as you can see with it only trivial commits . Easier for us to do although not yet as easy as I wish but we are getting there :)

PR #131, opened 2022-03-18
Update to Jetty 10.0.8 in prereqs file

akurtakov:
Refers to Maven artifacts directly instead of using them from
intermediate p2 repo created by us.
Part of #130 .


merks:
Does this imply it will only be PGP signed?

akurtakov:
Yes

PR #134, opened 2022-03-22
Fix CentOS Stream9 containers

akurtakov:
* use curl-minimal (provides only http/https/ftp)
* use dnf command
* use new webkit package name webkit2gtk3
* use Java 17 not Java 1.8
* drop useless GConf2 and cmake packages


sravanlakkimsetti (approved):

akurtakov:
Merged without waiting for verification build as it will not verify these parts.

PR #137, opened 2022-03-22
Bug 579215 - IES423: The hebrew translation will not be available in

ktatavarthi:
JDK17

Adding the vm argument -Djava.locale.useOldISOCodes=true to ini files.

Signed-off-by: Kalyan Prasad Tatavarthi <kalyan_prasad@in.ibm.com>


sravanlakkimsetti (commented):
I dont think this is required now. since the bug 579215 is verified. Closing

Issue #140, opened 2022-03-24
Update to ecj 3.30.0 version with java 18 support


sravanlakkimsetti:
Need to update compiler for 4.24 to use ecj with java 18 support

PR #141, opened 2022-03-24
Update to ecj  #140

sravanlakkimsetti:
Update to ecj 3.30.0 version having java 18 support


sravanlakkimsetti:
the new compiler has been picked up properly. other problem needs to be verified is comparator errors that can be done only in official build. Merging this one and starting a new I-build

PR #192, opened 2022-04-04
Add sat4j 2.3.6 to target

mickaelistria:
Older sat4j will be removed later when dependent features and bundles
have migrated to use 2.3.6.

eclipse-equinox/p2#28


vogella:
So our plan is to move libraries from Orbit usage to direct Maven usage? I very much like that, as we would have then one middle-man less.

ruspl-afed:
> So our plan is to move libraries from Orbit usage to direct Maven usage? I very much like that, as we would have then one middle-man less.

It would be good to announce such an intention for cross-project people, if it is really the case.

akurtakov:
It was discussed multiple times when the PGP story was discussed but there is explicit mail now https://www.eclipse.org/lists/cross-project-issues-dev/msg19066.html

PR #197, opened 2022-04-06
Explicit block for Jetty in target definition fil

vogella:
Helps to easily identify the Jetty dependencies which are frequently
updated.

### DIFF
--- a/eclipse.platform.releng.prereqs.sdk/eclipse-sdk-prereqs.target
+++ b/eclipse.platform.releng.prereqs.sdk/eclipse-sdk-prereqs.target
@@ -235,41 +235,9 @@
     <!-- uncomment 'eclipse_home' location, with text editor, for use in Eclipse IDE
     <location path="${eclipse_home}" type="Profile"/>
     -->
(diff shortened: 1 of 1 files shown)


vogella:
Looks like this:

image

@akurtakov and @mickaelistria WDYT?

mickaelistria:
I think it's indeed a bit better to have such groups. However, we need to ensure there are no extra constraints and drawbacks with that. @laeubi would there be any reason to *not* use groups (performance drawback, resolution inconsistencies...)?

laeubi:
All items of a group share the same settings for 

- includeDependencyScope
- includeSource
- missingManifest

and probably default BND settings. So if thats fine to apply to all there is no drawback in grouping them.

laeubi:
> which are frequently updated

Just in case not noted that functionality, selecting a dependency and click the Update button will look for the latest dependency in the maven repository.

mickaelistria:
The question is more the other way round (although I didn't express it clearly): is there any drawback in turning 1 single group into several ones?

PR #226, opened 2022-04-20
Update mockito to 4.5.0

akurtakov:


akurtakov:
Test feature is touched so should be good to go.

PR #230, opened 2022-04-23
Bug 534128 - Provide a home for Platform.setup

merks:


merks:
This has no impact on the build, so I aborted the build.

PR #235, opened 2022-04-25
Add osgi-services, -app, -coordinator, -log bundles from Maven-Central

HannesWell:
Add more osgi bundles to the Eclipse-SDK prerequisite target, which are required for:
- https://github.com/eclipse-equinox/equinox.framework/pull/44
- https://github.com/eclipse-equinox/equinox.bundles/pull/26

@tjwatson are you again fine with the versions? I selected the latest ones again.

### DIFF
--- a/eclipse.platform.releng.prereqs.sdk/eclipse-sdk-prereqs.target
+++ b/eclipse.platform.releng.prereqs.sdk/eclipse-sdk-prereqs.target
@@ -343,17 +343,101 @@
         </dependency>
       </dependencies>
     </location>
-    <location includeDependencyDepth="infinite" includeDependencyScopes="compile" includeSource="true" label="OSGi-util" missingManifest="error" type="Maven">
+    <location includeDependencyDepth="infinite" includeDependencyScopes="compile" includeSource="true" label="OSGi" missingManifest="error" type="Maven">
       <dependencies>
         <dependency>
           <groupId>org.osgi</groupId>
-          <artifactId>org.osgi.util.function</artifactId>
+          <artifactId>org.osgi.service.application</artifactId>
+          <version>1.1.0</version>
+          <type>jar</type>
+        </dependency>
+        <dependency>
+          <groupId>org.osgi</groupId>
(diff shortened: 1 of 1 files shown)


HannesWell:
Created https://gitlab.eclipse.org/eclipsefdn/emo-team/iplab/-/issues/2418 to check the not vetted licenses.

eclipse.platform.releng.prereqs.sdk/eclipse-sdk-prereqs.target
```diff
@@ -343,17 +343,107 @@
...
+        </dependency>
+        <dependency>
+          <groupId>org.osgi</groupId>
+          <artifactId>org.osgi.service.log.stream</artifactId>
+          <version>1.0.0</version>
+          <type>jar</type>
+        </dependency>
+        <dependency>
+          <groupId>org.osgi</groupId>
+          <artifactId>org.osgi.service.log</artifactId>
```
tjwatson, line 410: This is the right version, but it shows another complication.  If we pull in this bundle, it will provide yet another version of `org.osgi.service.log` which is also included and exported by the Equinox framework (`org.eclipse.osgi`).  I'm not sure what the best coarse of action is for this one.  We really want to avoid installing the OSGi WG `org.osgi.service.log` bundle because I suspect it only exports without importing the log packages.

What happens today with the `org.eclipse.osgi.services` bundle is that it ends up always importing  the `org.osgi.service.log` package from the framework and never gets selected as an exporter of the package.

This is another package in `org.eclipse.osgi.services` (like `component.annotations`) that we may just want to keep as-is until we can deprecate and remove the bundle altogether.

HannesWell: Thank you for that hint. I did not notice that.
I'm about to remove that bundle from the TP.

Lets discuss in https://github.com/eclipse-equinox/equinox.framework/pull/44 how to handle this further.

HannesWell:
@tjwatson do you agree that this is ready now?

tjwatson (approved):

PR #238, opened 2022-04-27
Eclipse 4.24 M2 Prerequisites: Orbit (fixup jackson)

jonahgraham:
Orbit 2022-06 M2 includes newer version of Jackson. Instead of
pointing at that newer version, consume Jackson directly from
Maven central as it is a proper OSGi bundle.


jonahgraham:
I made sure the target platform resolves, but I don't really know how to test if the build will work more than that.

sravanlakkimsetti:
> I made sure the target platform resolves, but I don't really know how to test if the build will work more than that.

Thanks @jonahgraham. This should be sufficient now. Rebuilding platform now

PR #244, opened 2022-04-30
Fix path in publish.xml so the links in performanceResultsTable.html …

SDawley:
…should actually work and take you to the results. Also fix typo in performance.php template (I don't know if perfTestNotes.html is actually getting generated but if it is now it will be found.


akurtakov:
Let's push things to see for next build.

PR #246, opened 2022-05-02
Eliminate pde.build from the platform SDK setup.

merks:


merks:
This doesn't affect the build, so I aborted it.

PR #248, opened 2022-05-03
Remove pde.build module from aggregator

akurtakov:
It's merged into pde. See
https://github.com/eclipse-pde/eclipse.pde/pull/39


akurtakov:
As this let's the maven build to at least start I push it in order to get one step ahead.

PR #255, opened 2022-05-05
Switch some Orbit deps to latest I-build

akurtakov:
They were locked at the old CVS repo as they were not available in the
new recipes way until very recently.


akurtakov:
Opened helpdesk issue https://gitlab.eclipse.org/eclipsefdn/emo-team/iplab/-/issues/2497 as it's not normal for the tool to complain for Orbit content - looks like database inconsistency to me.

PR #259, opened 2022-05-10
Make $HOME open to the 'root' group

mickaelistria:
So Jenkins/OpenShift user can write there, see
 https://docs.openshift.com/container-platform/3.9/creating_images/guidelines.html#use-uid

Should fix
https://github.com/eclipse-platform/eclipse.platform.swt/issues/108 as
 newer WebKitGTK transitively needs to write in HOME folder.


sravanlakkimsetti:
There was a small typo in the change update the PR with the fix

Issue #171, opened 2022-03-30
Eclipse 4.24 Prerequisites: EMF


SDawley:
Platform's M1 is on April 8th. Last build for M1 is 18:00 ET on April 6th.
@merks is there a new version for M1?

merks:
I've contributed the following to 2022-06 M1:

https://download.eclipse.org/modeling/emf/emf/builds/milestone/S202204030421

merks:
I'm unavailable the next two days, but I've kicked off a build that I will contribute for M2.

When the job https://ci.eclipse.org/emf/job/integration/634/ completes, this site will be updated with the location of the latest build for M2:

https://download.eclipse.org/modeling/emf/emf/builds/milestone/latest/index.html

It's currently references S202204030421 but will be updates I believe to S202204260307.

sravanlakkimsetti:
Updated for 4.24 M2

sravanlakkimsetti:
Any update for 4.24 M3

sravanlakkimsetti:
We are in RC phase now. Please let us know once the the final release version is available.
Last build for RC2 is at June 1 at 6PM eastern. It would be helpful if we get the release version by May 31.

merks:
The final release build is available and has been contributed to SimRel:

https://download.eclipse.org/modeling/emf/emf/builds/release/2.30

It has the same bits as the previous milestone build.

PR #264, opened 2022-05-16
EMF Update for 4.24 M3

SDawley:
Updates #171 

I noticed org.eclipse.emf.databinding.feature.group and org.eclipse.emf.databinding.feature.group were still set to 1.8 so I've updated them to 1.9.0.v20220516-1117 but asking for review from @merks to make sure they weren't left at 1.8 for a reason.


sravanlakkimsetti (approved):

sravanlakkimsetti:
In eclipse projects we do not increment versions unless there is a change in that particular plugin/bundle. So you may not see version increments for couple of releases. In this case there must be a fix in databinding feature, causing the version increment.

PR #206, opened 2022-04-08
[RelEng] Add reusable verifyFreezePeriode workflow

HannesWell:
This PR moves the Freeze-Period verification from the Jenkins build to a dedicated reusable workflow to get more detailed/clear PR checks as I have suggested on the [Mailing-Lists](https://www.eclipse.org/lists/platform-releng-dev/msg38299.html).

Each repository that requires the code-freeze check can add the following workflow-file
named `.github/workflows/verifyFreezePeriod.yml`:
```
# This workflow calls the Code-Freeze-Period check

name: Check Code Freeze Period

on:
  pull_request:
    branches: 
     - 'master'

jobs:
  check-freeze-period:
    uses: eclipse-platform/eclipse.platform.releng.aggregator/.github/workflows/verifyFreezePeriod.yml@master
```

This PR was originally developed at https://github.com/eclipse-platform/eclipse.platform.releng/pull/24

PR #272, opened 2022-05-23
Improve verifyFreezePeriod workflow

HannesWell:
As requested in https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/pull/206#issuecomment-1129257714 this PR implements a first minor improvement of the freezePeriod check workflow, by printing the queried calendar on the console.
So in case of a freeze period one can check manually when it has started and when it ends.

Additionally the script could be enhanced to print the end of the freeze period. This could be determined by querying the calendar in a loop until the first day is encountered that has no "stabilization" entry. Technically this would be feasible but I'm not sure if this is wanted. Usually the re-opening of the master is announced via the mailing-list, so I'm not sure what's the source of truth here. On the other hand, if this check is triggered on the first day after a freeze-period it succeeds anyway. 

Besides that we should consider to have dedicated "freeze-period" dates into the calendar. At the moment any date that contains the string `stabilization` is interpreted as freeze period but this check seems not be reliable anymore since there is no such date for the current days although the master branch is only open for PMC/Project lead approved changes (so it is not fully frozen).
More aliases for `stabilization` like "endgame", "signoff" or "promotion" could be added, but IHMO having dedicated freeze-periode dates would make it more clear.

This already caused confusion in https://github.com/eclipse-equinox/p2/pull/64 and https://github.com/eclipse-pde/eclipse.pde/pull/122

### DIFF
--- a/.github/workflows/verifyFreezePeriod.yml
+++ b/.github/workflows/verifyFreezePeriod.yml
@@ -9,12 +9,13 @@ jobs:
   verify-freeze-period:
     runs-on: ubuntu-latest
     steps:
-      - name: Checking whether Eclipse project is currently in stabilitzation/code-freeze period ...
+      - name: Checking whether Eclipse project is currently in stabilization/code-freeze period ...
         run: |
           today=$(TZ=UTC date "+%Y%m%d")
           tomorrow=$(TZ=UTC date -d "+1 days" "+%Y%m%d")
-          calID="cHJmazI2ZmRtcHJ1MW1wdGxiMDZwMGpoNHNAZ3JvdXAuY2FsZW5kYXIuZ29vZ2xlLmNvbQ"
-          curl "https://calendar.google.com/calendar/u/0/htmlembed?src=${calID}&mode=AGENDA&ctz=UTC&dates=${today}/${tomorrow}" | grep -i -q -e "stabilization"
+          calURL="https://calendar.google.com/calendar/u/0/htmlembed?src=cHJmazI2ZmRtcHJ1MW1wdGxiMDZwMGpoNHNAZ3JvdXAuY2FsZW5kYXIuZ29vZ2xlLmNvbQ"
+          echo "Querying calendar ${calURL}"
+          curl "${calURL}&mode=AGENDA&ctz=UTC&dates=${today}/${tomorrow}" | grep -i -q -e "stabilization"
           if [[ $? == 0 ]]; then
             echo "::error::Today is a freeze day"
             exit 1 #Exiting with non-0 makes this workflow fail


.github/workflows/verifyFreezePeriod.yml
```diff
@@ -13,8 +13,9 @@ jobs:
         run: |
           today=$(TZ=UTC date "+%Y%m%d")
           tomorrow=$(TZ=UTC date -d "+1 days" "+%Y%m%d")
-          calID="cHJmazI2ZmRtcHJ1MW1wdGxiMDZwMGpoNHNAZ3JvdXAuY2FsZW5kYXIuZ29vZ2xlLmNvbQ"
-          curl "https://calendar.google.com/calendar/u/0/htmlembed?src=${calID}&mode=AGENDA&ctz=UTC&dates=${today}/${tomorrow}" | grep -i -q -e "stabilization"
+          calURL="https://calendar.google.com/calendar/u/0/htmlembed?src=cHJmazI2ZmRtcHJ1MW1wdGxiMDZwMGpoNHNAZ3JvdXAuY2FsZW5kYXIuZ29vZ2xlLmNvbQ"
+          echo "Query calender ${calURL}"
+          curl "${calURL}&mode=AGENDA&ctz=UTC&dates=${today}/${tomorrow}" | grep -i -q -e "stabilization"
```
iloveeclipse, line 18: I haven't checked whhat calendar returns here, but wouldn't that be interesting too? 

So I would save the curl received data to variable, enhance check to also check for "endgame", and print the variable to the console too.

sravanlakkimsetti:
I updated the calendar to have an entry with "stabilization" key word on actual freeze days(tuesday and wednesday). So the verifyFreezePeriod job does fail only on actual freeze days. not on other days. 

If there is a specific keyword. that needs be used for freeze we can use that and I will update the calendar

.github/workflows/verifyFreezePeriod.yml
```diff
@@ -13,8 +13,9 @@ jobs:
         run: |
           today=$(TZ=UTC date "+%Y%m%d")
           tomorrow=$(TZ=UTC date -d "+1 days" "+%Y%m%d")
-          calID="cHJmazI2ZmRtcHJ1MW1wdGxiMDZwMGpoNHNAZ3JvdXAuY2FsZW5kYXIuZ29vZ2xlLmNvbQ"
-          curl "https://calendar.google.com/calendar/u/0/htmlembed?src=${calID}&mode=AGENDA&ctz=UTC&dates=${today}/${tomorrow}" | grep -i -q -e "stabilization"
+          calURL="https://calendar.google.com/calendar/u/0/htmlembed?src=cHJmazI2ZmRtcHJ1MW1wdGxiMDZwMGpoNHNAZ3JvdXAuY2FsZW5kYXIuZ29vZ2xlLmNvbQ"
+          echo "Query calender ${calURL}"
```
pbodnar, line 17: I would fix a typo and possibly also rephrase this to:

```suggestion
          echo "Querying calendar ${calURL}"
```

I don't know local conventions though, so maybe `Query` is just fine as well...

While at it, there is another typo on a line above: `stabilitzation`.

pbodnar (commented):
@HannesWell, thanks for this PR, I think it could be sufficient for a start - I think the link to the calendar helps a lot on its own. Maybe also some other keywords like `endgame` could be checked within this PR, but of course, I myself am totally new to this area, so I leave this question to others... :)

.github/workflows/verifyFreezePeriod.yml
HannesWell: Thank you. I applied it as suggested. Since I'm not a native speaker I'm not sure which one is better. Probably both is possible in the end.

HannesWell:
> I updated the calendar to have an entry with "stabilization" key word on actual freeze days(tuesday and wednesday). So the verifyFreezePeriod job does fail only on actual freeze days. not on other days.
> 
> If there is a specific keyword. that needs be used for freeze we can use that and I will update the calendar

When thinking about it a bit more, I think it is actually a good idea to have the code-freeze check failing a few days before the actual freeze period starts (as announced via the mailing list) and also to fail in the end before a release when changes require PMC/project lead approval. 
The failure then gives a heads-up there is a freeze period soon so don't wait too long (some PRs are not submitted immediately) or that at the moment changes need special requirements. Therefore I think it would be better to fail one time too often to indicate special circumstances at the moment/soon. This failure is not preventing a submission on a PR it is just a red-cross. Unfortunately as far as I know it is not possible to have a warning as build-result in GH.

> @HannesWell, thanks for this PR, I think it could be sufficient for a start - I think the link to the calendar helps a lot on its own. Maybe also some other keywords like `endgame` could be checked within this PR, but of course, I myself am totally new to this area, so I leave this question to others... :)

Great.

>> I haven't checked whhat calendar returns here, but wouldn't that be interesting too?
>It returns a full HTML page, you can check for yourself, for example for [today](https://calendar.google.com/calendar/u/0/htmlembed?src=cHJmazI2ZmRtcHJ1MW1wdGxiMDZwMGpoNHNAZ3JvdXAuY2FsZW5kYXIuZ29vZ2xlLmNvbQ&mode=AGENDA&mode=AGENDA&ctz=UTC&dates=20220524/20220525). So we would have to probably extract & show just the relevant part of the page, or use some other Calendar API method, provided there is some usable.

Exactly. Displaying that html-source code is probably not very helpful.
If we want to search for other words too "stabilization\|endgame\|signoff\|promotion" should do the job.
But it depends when we want this job to fail exactly.

Using the CalenderAPI would be nice, I tried that, but it requires an authentication token that we don't want to distribute over the repos because of the management overhead.

.github/workflows/verifyFreezePeriod.yml
HannesWell: Exactly. Displaying that html-source code is probably not very helpful.
If we want to search for other words too "stabilization\|endgame\|signoff\|promotion" should do the job.
But it depends when we want this job to fail exactly.

Using the CalenderAPI would be nice, I tried that, but it requires an authentication token that we don't want to distribute over the repos because of the management overhead.

sravanlakkimsetti:
> When thinking about it a bit more, I think it is actually a good idea to have the code-freeze check failing a few days before the actual freeze period starts (as announced via the mailing list) and also to fail in the end before a release when changes require PMC/project lead approval.

That was the initial idea. but the code freeze doesn't start till Monday evening though. it doesn't make any sense to fail the freeze job over the weekend. For this reason I went ahead modified the freeze dates to Tuesday and Wednesday. Still the job will fail from Monday till Thursday as we check for "stabilization" in today and tomorrows canendar.

sravanlakkimsetti (approved):

sravanlakkimsetti:
This does not impact the code freeze. Can be merged.

PR #282, opened 2022-05-27
Add felix.scr and felix.gogo.command/runtime/shell from Maven-Central

HannesWell:
Use the latest Jars from Maven-Central for the following Apache Felix bundles:
- org.apache.felix.scr
- org.apache.felix.gogo.runtime
- org.apache.felix.gogo.command
- org.apache.felix.gogo.shell

This updates the version of the first two. For completeness I also added the latter two, but I'm a bit concerned if that can cause problems due to different build qualifiers respectively I expect that the previous variants will not be updated if available (but that should actually not be a problem then).

This is not intended to be merged after the master for the next Eclipse release cycle has opened and not before the upcoming Apache-Felix release. 
Although the versions are not yet final, we can discuss the change itself already.

### DIFF
--- a/eclipse.platform.releng.prereqs.sdk/eclipse-sdk-prereqs.target
+++ b/eclipse.platform.releng.prereqs.sdk/eclipse-sdk-prereqs.target
@@ -28,12 +28,8 @@
       <unit id="org.apache.commons.logging.source" version="1.2.0.v20180409-1502"/>
       <unit id="org.apache.felix.gogo.command" version="1.1.2.v20210111-1007"/>
       <unit id="org.apache.felix.gogo.command.source" version="1.1.2.v20210111-1007"/>
-      <unit id="org.apache.felix.gogo.runtime" version="1.1.4.v20210111-1007"/>
-      <unit id="org.apache.felix.gogo.runtime.source" version="1.1.4.v20210111-1007"/>
       <unit id="org.apache.felix.gogo.shell" version="1.1.4.v20210111-1007"/>
       <unit id="org.apache.felix.gogo.shell.source" version="1.1.4.v20210111-1007"/>
-      <unit id="org.apache.felix.scr" version="2.1.24.v20200924-1939"/>
-      <unit id="org.apache.felix.scr.source" version="2.1.24.v20200924-1939"/>
 
       <unit id="org.junit" version="4.13.2.v20211018-1956"/>
       <unit id="org.junit.source" version="4.13.2.v20211018-1956"/>
@@ -501,6 +497,22 @@
           <groupId>com.fasterxml.jackson.core</groupId>
           <artifactId>jackson-databind</artifactId>
           <version>2.13.2.2</version>
+          <type>jar</type>
+        </dependency>
+      </dependencies>
+    </location>
+	  <location includeDependencyDepth="none" includeSource="true" label="Apache Felix" missingManifest="error" type="Maven">
+      <dependencies>
+        <dependency>
+          <groupId>org.apache.felix</groupId>
+          <artifactId>org.apache.felix.gogo.runtime</artifactId>
+          <version>1.1.6</version>
+          <type>jar</type>
+        </dependency>
+        <dependency>
+          <groupId>org.apache.felix</groupId>
+          <artifactId>org.apache.felix.scr</artifactId>
+          <version>2.2.2</version>
           <type>jar</type>
         </dependency>
       </dependencies>


HannesWell:
This is required for https://github.com/eclipse-equinox/equinox.framework/pull/44

When this is merged the following features that contain these bundles have to be touched in order to pick the new versions:
- `org.eclipse.e4.rcp`
- `org.eclipse.equinox.compendium.sdk`
- `org.eclipse.equinox.core.feature`
- `org.eclipse.equinox.server.core`

akurtakov:
If there is new version of felix-gogo-runtime and we are not waiting on any change in it - it might be best to push update of it right after opening the stream. felix-scr update can wait for the change in it.

HannesWell:
> If there is new version of felix-gogo-runtime and we are not waiting on any change in it - it might be best to push update of it right after opening the stream. felix-scr update can wait for the change in it.

Sure, I can split that. But if everything goes as planned/expected the new release of felix.scr should be available before the next Eclipse release cycle starts.

Btw. since this updates dependencies that are referenced in features in sub-projects, which are replaced with their baseline due to same version and build-qualifier and therefore don't automatically pick-up the change: Is there a way to have a clean successful build here? Or should we ignore the build failure and submit this once ready and touch the corresponding features immediately after it so that the subsequent I build succeeds.

(Referencing https://github.com/eclipse/tycho/issues/926 to link this as an contra-argument dependency inclusion.)

HannesWell:
> This is required for [eclipse-equinox/equinox.framework#44](https://github.com/eclipse-equinox/equinox.framework/pull/44)
> 
> When this is merged the following features that contain these bundles have to be touched in order to pick the new versions:
> 
>     * `org.eclipse.e4.rcp`
> 
>     * `org.eclipse.equinox.compendium.sdk`
> 
>     * `org.eclipse.equinox.core.feature`
> 
>     * `org.eclipse.equinox.server.core`

Only those features that are not changed as a consequence of [eclipse-equinox/equinox.framework#44](https://github.com/eclipse-equinox/equinox.framework/pull/44) need a force-qualifier update change. Which is only the compendium.sdk. I will do that as part of https://github.com/eclipse-equinox/equinox.bundles/pull/29.

HannesWell:
@tjwatson do you know an ETA for the required new felix.scr version already?

tjwatson:
> @tjwatson do you know an ETA for the required new felix.scr version already?

I've been distracted with work outside of Eclipse dev that is taking higher priority.  I'll try to get the artifacts built and staged for vote this week.  If that happens they can be in maven central by end of next week.

tjwatson:
I did manage to start this release vote for SCR 2.2.2 (see https://www.mail-archive.com/dev%40felix.apache.org/msg53793.html)

If you are able to it would be nice to confirm we can use it from the stage repo just as a test: https://repository.apache.org/content/repositories/orgapachefelix-1432/

HannesWell:
> I did manage to start this release vote for SCR 2.2.2 (see https://www.mail-archive.com/dev%40felix.apache.org/msg53793.html)

Great! Thank you.

> If you are able to it would be nice to confirm we can use it from the stage repo just as a test: https://repository.apache.org/content/repositories/orgapachefelix-1432/


Yes that is possible. I just added a additional repository element to the .target file that points to the staging repo.

HannesWell:
The build failure is:

```
[ERROR] Cannot resolve project dependencies:
[ERROR]   Software being installed: org.eclipse.equinox.compendium.sdk.feature.group 3.22.300.qualifier
[ERROR]   Missing requirement: org.eclipse.equinox.compendium.sdk.feature.group 3.22.300.qualifier requires 'org.eclipse.equinox.p2.iu; org.apache.felix.scr.source 0.0.0' but it could not be found
```

I wonder why this happens. The target-location is configured with `includeSource="true"` and the staging repo seems to contain sources.
@laeubi do you have any idea?

laeubi:
> I wonder why this happens. The target-location is configured with `includeSource="true"` and the staging repo seems to contain sources.

You could open the target in PDE and check if it is maybe a different name, beside that it might be that sources are disabled in the pom (then the setting from the target is ignored).

HannesWell:
> > I wonder why this happens. The target-location is configured with `includeSource="true"` and the staging repo seems to contain sources.
> 
> You could open the target in PDE and check if it is maybe a different name, beside that it might be that sources are disabled in the pom (then the setting from the target is ignored).

Thanks for that hint. It looks like the felix.scr source is not found. The content location shows this:

grafik

Also I also checked my local maven repos and `.m2\repository\org\apache\felix\org.apache.felix.scr\2.2.2` did not contain the sources-jar.
I also created a default Maven-project that has a dependency  `org.apache.felix:org.apache.felix.scr:2.2.2` and the additional repo specified and then the sources-jar where downloaded by m2e. Therefore I think the sources are available in general but Maven-targets fail to download them.

What is interesting is that if `o.a.felix.scr/2.2.2` directory is empty/does not exist and I resolve the Maven-target (i.e. by opening the editor) the directory looks like the following afterwards:

grafik


M2E seems to attempt to download the sources but does not find them, from the content of the `org.apache.felix.scr-2.2.2-sources.jar.lastUpdated` file I strongly suspect that the additional repository is not available when searching sources:
```
#NOTE: This is a Maven Resolver internal implementation file, its format can be changed without prior notice.
#Fri Jun 10 10:25:19 CEST 2022
https\://repo.maven.apache.org/maven2/.lastUpdated=1654849519413
https\://repo.maven.apache.org/maven2/.error=
```
`org.apache.felix.scr-2.2.2.jar.lastUpdated` and `org.apache.felix.scr-2.2.2.pom.lastUpdated` have the following content (only slightly different timestamps)

```
#NOTE: This is a Maven Resolver internal implementation file, its format can be changed without prior notice.
#Fri Jun 10 10:25:19 CEST 2022
https\://repo.maven.apache.org/maven2/.lastUpdated=1654849519186
https\://repository.apache.org/content/repositories/orgapachefelix-1432/.lastUpdated=1654849519338
https\://repo.maven.apache.org/maven2/.error=
```

Is it possible that the additional repositories have to be added to the resolution context separately to fetch sources?

In general fetching and including sources of Maven-artifact works. For example I have already included many of the osgi-bundles from Maven Central including their sources in features..

laeubi:
> Is it possible that the additional repositories have to be added to the resolution context separately to fetch sources?
> 
> In general fetching and including sources of Maven-artifact works. For example I have already included many of the osgi-bundles from Maven Central including their sources in features..

Probably, as you has a test-case now it should be just a matter of putting a breakpoint to find out :-)
The maven repository system is quite a mystery...

HannesWell:
> > Is it possible that the additional repositories have to be added to the resolution context separately to fetch sources?
> > In general fetching and including sources of Maven-artifact works. For example I have already included many of the osgi-bundles from Maven Central including their sources in features..
> 
> Probably, as you has a test-case now it should be just a matter of putting a breakpoint to find out :-) The maven repository system is quite a mystery...

Yes I can do that, but probably not before sunday/the beginning of next week. But I will continue on that.
If you have any immediate hints, don't hesitate to let me know. :)

HannesWell:
Although the build failed in the end every relevant parts look good now.
Tycho and M2E now also properly fetch source bundles from additional repositories. And the felix.scr-2.2.2 from the staging repo also resolves well (@tjwatson that's probably the part interesting for you).
So we are ready for submission as soon as felix.scr-2.2.2 has been officially released and is available from Maven central.

eclipse.platform.releng.prereqs.sdk/eclipse-sdk-prereqs.target
```diff
@@ -28,12 +28,8 @@
       <unit id="org.apache.commons.logging.source" version="1.2.0.v20180409-1502"/>
       <unit id="org.apache.felix.gogo.command" version="1.1.2.v20210111-1007"/>
```
HannesWell, line 29: @akurtakov do you think we should also switch to `org.apache.felix.gogo.command` and `org.apache.felix.gogo.shell` from Maven-Central?
Unfortunately there is no more recent version available and the available versions have a build-qualifier that is older than the variant from Orbit that is currently used. Do you think this could cause problems?

akurtakov: If the qualifier is older than the Orbit one , updated installations will just keep the old one so we would better not go that path.

HannesWell: Acknowledge. Then lets migrate them to Maven-dependencies when new versions are available. 👍🏽
Thanks.

laeubi: Qualifiers only change mean its irrelevant if you use A or B ... just from OSGi perspective...

HannesWell:
@tjwatson do you have any updates on the release of felix.scr 2.2.2?
The answers to https://www.mail-archive.com/dev%40felix.apache.org/msg53793.html look promising, but I could not found a release yet.
Sorry to annoy you with that, I just would like to complete this and all depending PRs. :)

tjwatson:
Thanks for the reminder.  I'll close out the vote tomorrow and publish.

HannesWell:
Do I have to update [publish-to-maven-central/SDK4Mvn.aggr](https://github.com/eclipse-platform/eclipse.platform.releng/blob/master/publish-to-maven-central/SDK4Mvn.aggr) once this is merged like in https://github.com/eclipse-platform/eclipse.platform.releng/pull/48/? Because the the felix.scr artifact is not covered by the current mappings.

On the other hand felix.scr should not be used on as direct dependency, so I wonder if this is necessary.

tjwatson:
Felix SCR 2.2.2 should now be published to maven central.  https://repo1.maven.org/maven2/org/apache/felix/org.apache.felix.scr/2.2.2/

HannesWell:
> Felix SCR 2.2.2 should now be published to maven central. https://repo1.maven.org/maven2/org/apache/felix/org.apache.felix.scr/2.2.2/

Yes it is. The commit I just pushed uses Maven-Central already.
Thanks for completing the release.

If you can review and approve https://github.com/eclipse-equinox/equinox/pull/30 I would like to submit that and this PR tomorrow at the same day to avoid forced qualifier updates.

Issue #293, opened 2022-06-02
Switch builds to use Java 17 JVM 


akurtakov:
Tycho is working on new major version which will switch to Java 17 requirement https://github.com/eclipse/tycho/issues/983 . As this version is needed for latest fixes and for Java 19 work our builds have to switch.

akurtakov:
Note that one needs 17.0.3 at least as older versions have a bug that breaks jdt.core.
See https://github.com/eclipse-jdt/eclipse.jdt.core/issues/147 and https://github.com/eclipse-jdt/eclipse.jdt.core/issues/114

akurtakov:
I consider this one done as the main I-build is handled using Java 17. If I missed it somewhere it would be handled separately.

PR #331, opened 2022-06-09
Switch aggregator I and Y builds to use Java 17

SDawley:
Tracked in #293


akurtakov:
Verification build will prove nothing on this patch but it looks good. Merging.

PR #342, opened 2022-06-14
Fix NOTICE file

akurtakov:
* Point to correct git urls.
* Remove platform.runtime as it's merged in platform one


akurtakov:
No need to wait for verification build as it's just documentation file.

PR #345, opened 2022-06-16
Change aggregator build to use merged equinox repo

akurtakov:
Switches aggregator build to use the new merged equinox repository
instead of the old split ones.
See https://github.com/eclipse-equinox/equinox.bundles/issues/18


akurtakov:
I'm merging and starting new I-build to verify as with platform.runtime verification build didn't work but full build did.

PR #346, opened 2022-06-16
Fix I-build cloning to use new merged equinox repo

akurtakov:
See eclipse-equinox/equinox.bundles#18


akurtakov:
Verification build will not work here.

PR #351, opened 2022-06-17
Add the bundle and bnd annotations to the target platform

laeubi:
Recently there [was a demand to use the bnd annotations](https://github.com/eclipse-equinox/equinox/issues/27) and there is a [open PDE issue to support the bundle annotations](https://github.com/eclipse-pde/eclipse.pde/issues/69), for both use cases we would need the bnd/osgi annotation available at the targetplatform.

### DIFF
--- a/eclipse.platform.releng.prereqs.sdk/eclipse-sdk-prereqs.target
+++ b/eclipse.platform.releng.prereqs.sdk/eclipse-sdk-prereqs.target
@@ -354,7 +354,19 @@
     -->
 
     <location includeDependencyDepth="infinite" includeDependencyScopes="compile" includeSource="true" label="OSGi" missingManifest="error" type="Maven">
-      <dependencies>
+        <dependencies>
+		  <dependency>
+		  <groupId>biz.aQute.bnd</groupId>
+		  <artifactId>annotation</artifactId>
+		  <version>2.4.0</version>
+		  <type>jar</type>
+        </dependency>
+        <dependency>
+		  <groupId>org.osgi</groupId>
+		  <artifactId>org.osgi.annotation.bundle</artifactId>
+		  <version>2.0.0</version>
+		  <type>jar</type>
+        </dependency>
         <dependency>
           <groupId>org.osgi</groupId>
           <artifactId>org.osgi.service.cm</artifactId>


HannesWell:
Given that in the long run PDE should support bundle annotations without adding corresponding dependencies to the Manifest (e.g. optional Import-Package/Require-Bundle) I think it is better to not add the annotation bundles to the target-platform and to not add such dependencies to the Manifest because it has the risk that somebody starts to rely on this somehow and we later have trouble to remove them when PDE has implemented https://github.com/eclipse-pde/eclipse.pde/issues/69.

As a workaround to be able to use those annotations immediately I suggest instead to embedded the corresponding jars into the using plug-ins and reference them via `jars.extra.classpath` entries in the build.properties just like for example `org.eclipse.osgi.services` is doing it at the moment:
https://github.com/eclipse-equinox/equinox/blob/776c5262fb0a4ece90d6abc464cbb053231ec796/bundles/org.eclipse.osgi.services/build.properties#L24

This way nothing is exposed and we can easily revert the workaround once PDE supports bundle and/or bnd annotations.

laeubi:
> Given that in the long run PDE should support bundle annotations without adding corresponding dependencies to the Manifest (e.g. optional Import-Package/Require-Bundle) I think it is better to not add the annotation bundles to the target-platform

How will PDE then consume these? :-)



> As a workaround to be able to use those annotations immediately I suggest instead to embedded the corresponding jars into the using plug-ins and reference them via `jars.extra.classpath`

I don't think we should copy (binary) artifacts to PDE and embed them, as we need them in the target anyways, for PDE to add them to the classpath and for tycho to find those later on for compile-time computation.

> This way nothing is exposed and we can easily revert the workaround

Why is a bundle a workaround? What is exposed is the whole decision of the bundle author, and what is consumed a decision of the consumer so we just provide the required bits here.

HannesWell:
> > Given that in the long run PDE should support bundle annotations without adding corresponding dependencies to the Manifest (e.g. optional Import-Package/Require-Bundle) I think it is better to not add the annotation bundles to the target-platform
> 
> How will PDE then consume these if they are not there? :-)

Actually my first attempt would have been to embed the annotation-jar into the corresponding PDE bundle so that it can be placed on the compile-time classpath of the consuming bundle (basically like it is done now for the ds.component annotation but probably all jars in one bundle). I would have fetched those jars directly from Maven-Central in the pom as part of the build process.
I think the reason that this approach is favored is the same like for the `org.osgi.annotation.versioning` and `org.osgi.service.component.annotations` as it was discussed here:
- https://github.com/eclipse-equinox/equinox.framework/pull/44#discussion_r858748246
- https://github.com/eclipse-equinox/equinox/issues/18#issuecomment-1157433294

Compile-time only dependencies should not be part of the Manifest since it actually dedicated to runtime.
But indeed of course this approach does not scale very well since each annotation jar has to be embedded into a corresponding PDE plugin.

I think a robust solution would be to add support for compile-time only dependencies to PDE, just like it would be good to have test-only dependencies. But lets discuss that at PDE in a corresponding issue. :)

> > This way nothing is exposed and we can easily revert the workaround
> 
> Why is a bundle a workaround? What is exposed is the whole decision of the bundle author, and what is consumed a decision of the consumer so we just provide the required bits here.

I meant exposure at runtime. For the runtime it is irrelevant what is needed to compile the bundle. So even if the author of a bundle decided to use those annotation at compile-time the should not be visible at runtime.

I'm not sure what the best solution is for now. Maybe @tjwatson can help?

laeubi:
> I would have fetched those jars directly from Maven-Central in the pom as part of the build process.

This won't also not work well without additional build steps in the Platform SDK.

> I meant exposure at runtime. For the runtime it is irrelevant what is needed to compile the bundle.

Obviously not... PDE runs inside OSGi framework and it is relevant for PDE :-)

HannesWell:
> > I would have fetched those jars directly from Maven-Central in the pom as part of the build process.
> 
> This won't also not work well without additional build steps in the Platform SDK.

That could be the case. As said above that would just be my first attempt/suggestion. Not saying that this is the ideal solution. :)
 
> 
> > I meant exposure at runtime. For the runtime it is irrelevant what is needed to compile the bundle.
> 
> Obviously not... PDE runs inside OSGi framework and it is relevant for PDE :-)

Are you referring to the PDE(-TargetPlatform)-State? That is indeed a stripped down resolver of the OSGi-Framework that is not started. But the JDT classpath that is used to finally compile the sources isn't. So compile-time dependencies bypass the OSGi-framework here.

However I'm not saying this is a totally bad approach, I'm just not sure it is the right one on the long run (or the best for a short-term solution).

laeubi:
> Are you referring to the PDE(-TargetPlatform)-State? That is indeed a stripped down resolver of the OSGi-Framework that is not started.

Junit Container resolves items from the current running eclipse and so could PDE do so we dont need to wrap/embedd anything that already is a bundle.

HannesWell:
> > Are you referring to the PDE(-TargetPlatform)-State? That is indeed a stripped down resolver of the OSGi-Framework that is not started.
> 
> Junit Container resolves items from the current running eclipse and so could PDE do so we dont need to wrap/embedd anything that already is a bundle.

That's actually not a bad suggestion, however for example the org.osgi.service.component.annotations bundle is designed to not resolve at runtime and therefore cannot be used in such way:
- https://github.com/eclipse-equinox/equinox/issues/18#issuecomment-1157433294

And from the latest discussion in the area of replacing the osgi-packages with the original jars, I'm not sure if it is a good goal to add those annotation bundles to the runtime.
But I think @tjwatson can tell better how the best long term solution should look like, especially how we should then obtain the annotation jars. Just having them in the TP would be more convenient from a build perspective.

laeubi:
> however for example the org.osgi.service.component.annotations bundle is designed to not resolve at runtime and therefore cannot be used in such way

We don't need it to resolve, it just needs to be present in the P2 profile!

sravanlakkimsetti:
Is this still required?

laeubi:
If we find no better way, yes...

tjwatson:
> And from the latest discussion in the area of replacing the osgi-packages with the original jars, I'm not sure if it is a good goal to add those annotation bundles to the runtime. But I think @tjwatson can tell better how the best long term solution should look like, especially how we should then obtain the annotation jars. Just having them in the TP would be more convenient from a build perspective.

I've not followed this issue closely, does this change actually have the annotation bundles installed into the runtime?  What seems particularly worrisome is `biz.aQute.bnd:annotation`.  I believe BND tools never installs this as a bundle at runtime, yet we have to?  I also hope it doesn't all of a sudden cause issues for BND tools to install into Eclipose.  For compile only JARs it would be best to figure out a way to not have these types of bundles installed (resolved) in the OSGi framework.

Are bundle projects that use these bundles going to import packages from them to use their annotations?

eclipse.platform.releng.prereqs.sdk/eclipse-sdk-prereqs.target
```diff
@@ -354,7 +354,19 @@
     -->
 
     <location includeDependencyDepth="infinite" includeDependencyScopes="compile" includeSource="true" label="OSGi" missingManifest="error" type="Maven">
-      <dependencies>
+        <dependencies>
+		  <dependency>
+		  <groupId>biz.aQute.bnd</groupId>
+		  <artifactId>annotation</artifactId>
+		  <version>2.4.0</version>
```
bjhargrave, line 361: https://search.maven.org/artifact/biz.aQute.bnd/biz.aQute.bnd.annotation/6.3.1/jar is the latest version of the Bnd annotations. 2.4.0 is quite ancient.

bjhargrave:
> does this change actually have the annotation bundles installed into the runtime?

We don't want or need them installed at runtime. These should be "compileOnly" dependencies.



> I believe BND tools never installs this as a bundle at runtime, yet we have to? I also hope it doesn't all of a sudden cause issues for BND tools to install into Eclipse.

Bnd does not install `biz.aQute.bnd.annotations` at runtime because `biz.aQute.bndlib` includes all the packages in `biz.aQute.bnd.annotations`. So we don't want to install `biz.aQute.bnd.annotations` at runtime duplicating the exports.

merks:
Note that these are already in the release train repo and are pulled in by m2e's "pde" feature:

image

I think nothing prevents that feature from being installed in the IDE. 

EMF's RAP features have this to prevent them from being installed into an IDE:

image

But the m2e PDE feature doesn't appear to have that:

image

For EMF's RAP feature it looks like this:

image

laeubi:
m2e uses BND-lib, what is supposed to be "installed" as we use the BND Analyzer at runtime.

laeubi:
BTW this is related to https://github.com/osgi/osgi/issues/448 as if there would be "standard" annotation we won't have any need to include bnd specific ones :-)

PR #352, opened 2022-06-18
Skip API-analysis automatically on absent api.tools.apiAnalysisNature

HannesWell:
and also simplify this check for 'api-generation' profile.

For the check ANT's 'resourcecontains'-condition is used:
https://ant.apache.org/manual/Tasks/conditions.html#resourcecontains

This avoids most of the `<skipAPIAnalysis>true<skipAPIAnalysis>` properties in poms, because if an API-analysis is not wanted the API-analysis nature and builder can simply be removed from the .project file.
In most cases where `<skipAPIAnalysis>true<skipAPIAnalysis>` is specified the nature+builder are already absent any ways (e.g. for test-plugins) and that property is only specified to avoid problems during build.
Automatically detecting based on the presence of the nature ensures consistently enabled/disabled API-analysis during build and in the UI.
Even more important allows to remove the `skipAPIAnalysis`-property from most poms, actually all poms since the apiAnalysis-nature can simply be removed if API-analysis is not wanted, which makes many more poms trivial and allows to remove them in favor of Tycho-pomless.

@mickaelistria, @akurtakov what do you think?

### DIFF
--- a/eclipse-platform-parent/pom.xml
+++ b/eclipse-platform-parent/pom.xml
@@ -751,15 +751,12 @@
                 <configuration>
                   <exportAntProperties>true</exportAntProperties>
                   <target>
-                    <loadfile srcfile=".project" property="natureIn.project" failonerror="false">
-                      <filterchain>
-                        <linecontains>
-                          <contains value="org.eclipse.pde.api.tools.apiAnalysisNature"/>
-                        </linecontains>
-                      </filterchain>
-                    </loadfile>
                     <condition property="skipAPIDescription" value="false" else="true">
-                        <isset property="natureIn.project" />
+                      <!-- If the property is already set this has no effect because properties are immutable in ANT -->
+                      <and>
+                        <available file="${basedir}/META-INF/MANIFEST.MF"/>
+                        <resourcecontains resource="${basedir}/.project" substring="org.eclipse.pde.api.tools.apiAnalysisNature" />
+                      </and>
                     </condition>
                   </target>
                 </configuration>
@@ -844,10 +841,12 @@
                 <configuration>
                   <exportAntProperties>true</exportAntProperties>
                   <target>
-                    <condition property="skipAPIAnalysis" value="true" else="false">
-                      <not>
+                    <condition property="skipAPIAnalysis" value="false" else="true">
+                      <!-- If the property is already set this has no effect because properties are immutable in ANT -->
+                      <and>
                         <available file="${basedir}/META-INF/MANIFEST.MF"/>
-                      </not>
+                        <resourcecontains resource="${basedir}/.project" substring="org.eclipse.pde.api.tools.apiAnalysisNature" />
+                      </and>
                     </condition>
                     <echo file="${project.build.directory}/${project.artifactId}-apiBaseline.target">
                       <![CDATA[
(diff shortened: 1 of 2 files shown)


eclipse-platform-parent/pom.xml
```diff
@@ -751,15 +751,12 @@
...
-                          <contains value="org.eclipse.pde.api.tools.apiAnalysisNature"/>
-                        </linecontains>
-                      </filterchain>
-                    </loadfile>
                     <condition property="skipAPIDescription" value="false" else="true">
-                        <isset property="natureIn.project" />
+                      <!-- If the property is already set this has no effect because properties are immutable in ANT -->
+                      <and>
+                        <available file="${basedir}/META-INF/MANIFEST.MF"/>
+                        <resourcecontains resource="${basedir}/.project" substring="org.eclipse.pde.api.tools.apiAnalysisNature" />
```
HannesWell, line 758: resourcecontains should return false if the specified file does not exist, but since the API analysis is only possible (now) for plug-ins checking the existance of the MANIFEST.MF should make this check a little faster for features since their .project files does not have to be read.

mickaelistria:
I think it's the right approach. However, we would need to upgrade the version upgrade documentation which IIRC has mentioned `skipAPIAnalysis` to replace with information about nature in .project.
@sravanlakkimsetti Can you please remind us where such documentation is currently maintained?

mickaelistria:
While I like the code improvement, I don't get how this does the job better than current state (with combination of `loadFile/contains` and `isset`. I think what is mentioned in the title here is already expected to work. So either it's not working and the patch fixes it; or it's already working and the patch "just" improve the code. Which case is it?

HannesWell:
> While I like the code improvement, I don't get how this does the job better than current state (with combination of `loadFile/contains` and `isset`. I think what is mentioned in the title here is already expected to work. So either it's not working and the patch fixes it; or it's already working and the patch "just" improve the code. Which case is it?

This patch improves the code for the API-generation, which was already skipped automatically when the apiAnalysis-nature is absent, and fixes the behavior for the API-check, which was only skipped automatically for non-Plugin-projects (i.e. projects without a META-INF/MANIFEST.MF). The API-check is now skipped as well when the apiAnalysis-nature is absent.

mickaelistria:
OK, great. Let's merge this right away so we can remove pom files!

HannesWell:
@laeubi you might also find this interesting.
I'm about to prepare a PR for Equinox and PDE.

mickaelistria:
I admit I now have 2nd thoughts as I'm afraid it can cause a bug: we do want bundles that are new to have API generation enabled (actually we do want it for all bundles, so this should not be a profile), and IIRC those new bundles don't have the nature by default as it may trigger an error.
But I don't think it's a new issue.
I suspect we should always run the api-generation independently of the nature.

HannesWell:
Yes the API-generation is skipped for projects that don't have the nature, but yes that was also the case before.

Running the API generation for all Plug-in projects (i.e. those with a META-INF/MANIFEST.MF) should fix that.
Do you know if the apiGeneration application cares if a nature/builder is present? I know that the APIAnalysis application adds the builder automatically if it is missing.

I wonder what's the right case for new projects that are added as incubators and don't have a stable API yet and therefore don't obey the API-rules. Is it bad that for those the API would be generated as well?

PR #356, opened 2022-06-21
[Build] Use eclipse/dash-licenses' reusable licenses-check workflow

HannesWell:
The workflow checks if all licenses are vetted on each push to the master branch and on each new (push to a) PR.
Furthermore committers can request a review of unvetted licenses by adding the following comment to a PR (open or closed):
```
/request-license-review
```
Attempts to request a review from non-committers to this repository are rejected by the workflow.
If a committer requests a license-review the license vetting summary files created by the dash-tools plug-in are attached to the workflow. So if you are interested in the summary you can comment `/request-license-review` and download the summary, even if you know all licenses are vetted. The workflow handles this case gracefully.

The workflow works quite well for m2e, but for the near future I cannot foresee version updates so in order to test it more we need to use it in more projects.

@akurtakov do you think we should give it a try here?
If you agree I can open a corresponding EF help-desk issue to ask for adding a GITLAB_API token to this repositories secret store.

In the meantime this workflow can be used already. Only license review requests will be rejected until the token is added.


akurtakov:
Overall I agree. But the shared workflow seems to not work with git submodules.

HannesWell:
The mentioned fix was submitted and I updated the calling workflow to recursively check out sub,modules, like it was done before and it looks it works well now.

Furthermore I created another help-desk issue to ask for a gitlab API token for this repository as well:
https://gitlab.eclipse.org/eclipsefdn/helpdesk/-/issues/1464

PR #401, opened 2022-07-15
Update Ant to 1.10.12 for 4.19 Maintenance branch

SarikaSinha:
see #https://github.com/eclipse-platform/eclipse.platform/issues/45


sravanlakkimsetti (changes requested):
Jetty got downgraded here. can you please update the patch with only ant?

PR #403, opened 2022-07-15
Include BouncyCastle from Maven Central

mickaelistria:


mickaelistria:
Opened IP review https://gitlab.eclipse.org/eclipsefdn/emo-team/iplab/-/issues/3008 that requires approval before we can merge.

merks:
A nasty side effect of this change is that while the Mavens are JCE-signed:

https://download.eclipse.org/oomph/archive/reports/download.eclipse.org/eclipse/updates/4.25-I-builds/index/bcpg_1.71.0.html
https://download.eclipse.org/oomph/archive/reports/download.eclipse.org/eclipse/updates/4.25-I-builds/index/bcprov_1.71.0.html

The certificate that's used is not rooted in the Java 17.0.3 cacerts so installing these or updating to these will result in a prompt for for the certificate:

image

FYI @jonahgraham

merks:
The second segment in the chain look the same though:

image

merks:
@mickaelistria  It seems the only way to avoid this problem (prompt) is to sign the jar.  I don't believe the current logic in org.eclipse.equinox.internal.p2.engine.phases.CertificateChecker.checkCertificates(SignedContentFactory) would handle a PGP signature once it detects the jar is jar-signed so we really do need to jar sign it. 

So I think this change should be reverted and if the newer 1.71.0 version is needed, we should make sure it's in Orbit.

Or do you have a different plan?

iloveeclipse:
> So I think this change should be reverted and if the newer 1.71.0 version is needed, we should make sure it's in Orbit.

See https://bugs.eclipse.org/bugs/show_bug.cgi?id=580438

Matthias is going to add it to Orbit.

tomaswolf:
Please note that the Orbit version of BouncyCastle will not contain the license-encumbered algorithms IDEA or AES/OCB. Using the maven artifacts, those would be included. The algorithms are removed from the Orbit version upon request from Eclipse Legal. See [CQ 9421](https://dev.eclipse.org/ipzilla/show_bug.cgi?id=9421).

iloveeclipse:
I've created https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/issues/418 to continue discussion there.

trancexpress:
> Please note that the Orbit version of BouncyCastle will not contain the license-encumbered algorithms IDEA or AES/OCB. Using the maven artifacts, those would be included. The algorithms are removed from the Orbit version upon request from Eclipse Legal. See [CQ 9421](https://dev.eclipse.org/ipzilla/show_bug.cgi?id=9421).

Is that the only such altered bundle? If not, is there a list somewhere?

msohn:
bouncycastle 1.71.0 is now available in https://download.eclipse.org/tools/orbit/downloads/drops/I20220724195112/

PR #412, opened 2022-07-19
Remove Orbit's javax.servlet.jsp

mickaelistria:


merks:
@akurtakov @mickaelistria 

I was trying to figure out why the removed Orbit bundle is still present in the latest 4.25 IBuild:

image

I believe it's because this excludes 2.3 or higher

image

which excludes the direct-from-maven version 2.3.3:

image

I assume we should fix that right?   We could change the upper bound to 2.4, but why not to 3.0?  What do you guys think?

akurtakov:
It's worth trying.  Going for [2.0,3.0) sounds good to me.

merks:
@akurtakov  Thanks for the quick response!

I'm 99.99% sure this will fix the problem because the SimRel analyzer for staging shows this as the only 'strict' dependency:

image

It's great to get rid of this older bundle because it is really, really old and signed with a really old certificate that is typically no longer root in the cacerts...

PR #415, opened 2022-07-21
Update tycho-build to 2.7.4.

akurtakov:
Fixes https://github.com/eclipse-equinox/equinox/issues/83


akurtakov:
Pushed without waiting for verification so we can run I-build sooner.

PR #419, opened 2022-07-24
Generate API description for all Plug-in Projects

HannesWell:
As suggested in https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/pull/352#issuecomment-1159488042, this enables the API-generation for all Eclipse Plug-in projects, regardless of if they have the API-Builder nature.
This also enables API generation for incubating Plug-ins or Plug-ins that are just added and can therefore not have their API checked, but are intended to be checked after the next release.

@mickaelistria if you want this now we can merge this now (given the build succeeds), but I'm on vacation for the next two weeks and can therefore not assist if something has to be fixed. If you want to wait, I'm fine with that either.

### DIFF
--- a/eclipse-platform-parent/pom.xml
+++ b/eclipse-platform-parent/pom.xml
@@ -727,39 +727,11 @@
     <profile>
       <id>api-generation</id>
       <activation>
-        <!-- Does it have to be a profile if we can evaluate condition reliably in antrun? -->
+        <!-- Generate API for all Plug-in projects -->
         <file><exists>META-INF/MANIFEST.MF</exists></file>
-        <property>
-          <name>!longnotexistingproperty</name>
-        </property>
       </activation>
       <build>
         <plugins>
-          <plugin>
-            <groupId>org.apache.maven.plugins</groupId>
-            <artifactId>maven-antrun-plugin</artifactId>
-            <executions>
-              <execution>
-                <id>check-has-apiTools-nature</id>
-                <phase>validate</phase>
-                <goals>
-                  <goal>run</goal>
-                </goals>
-                <configuration>
-                  <exportAntProperties>true</exportAntProperties>
-                  <target>
-                    <condition property="skipAPIDescription" value="false" else="true">
-                      <!-- If the property is already set this has no effect because properties are immutable in ANT -->
-                      <and>
-                        <available file="${basedir}/META-INF/MANIFEST.MF"/>
-                        <resourcecontains resource="${basedir}/.project" substring="org.eclipse.pde.api.tools.apiAnalysisNature" />
-                      </and>
-                    </condition>
-                  </target>
-                </configuration>
-              </execution>
-            </executions>
-          </plugin>
           <plugin>
             <groupId>org.eclipse.tycho.extras</groupId>
             <artifactId>tycho-eclipserun-plugin</artifactId>


HannesWell:
It looks like PDE-API Tools have to be adjusted as well to generate the API-description for Plugins without corresponding nature:

```
[INFO] --- tycho-eclipserun-plugin:3.0.0-SNAPSHOT:eclipse-run (build-.api_description) @ org.eclipse.pde.ds.lib ---
[INFO] Toolchain in tycho-eclipserun-plugin: JDK[/opt/tools/java/openjdk/jdk-11/latest]
[INFO] Expected eclipse log file: /home/jenkins/agent/workspace/latform.releng.aggregator_PR-419/eclipse.pde/ds/org.eclipse.pde.ds.lib/target/apigeneration-workspace/data/.metadata/.log
[INFO] Command line:
	[/opt/tools/java/openjdk/jdk-11/latest/bin/java, -jar, /home/jenkins/agent/workspace/latform.releng.aggregator_PR-419/.m2/repository/p2/osgi/bundle/org.eclipse.equinox.launcher/1.6.400.v20210924-0641/org.eclipse.equinox.launcher-1.6.400.v20210924-0641.jar, -install, /home/jenkins/agent/workspace/latform.releng.aggregator_PR-419/eclipse.pde/ds/org.eclipse.pde.ds.lib/target/apigeneration-workspace, -configuration, /home/jenkins/agent/workspace/latform.releng.aggregator_PR-419/eclipse.pde/ds/org.eclipse.pde.ds.lib/target/apigeneration-workspace/configuration, -data, /home/jenkins/agent/workspace/latform.releng.aggregator_PR-419/eclipse.pde/ds/org.eclipse.pde.ds.lib/target/apigeneration-workspace/data, -application, org.eclipse.pde.api.tools.apiGeneration, -projectName, org.eclipse.pde.ds.lib_1.1.500.v20210209-1250, -project, /home/jenkins/agent/workspace/latform.releng.aggregator_PR-419/eclipse.pde/ds/org.eclipse.pde.ds.lib, -binary, , -target, /home/jenkins/agent/workspace/latform.releng.aggregator_PR-419/eclipse.pde/ds/org.eclipse.pde.ds.lib/target]
The project does not have an API Tools nature so a api_description file will not be generated
```

eclipse-platform-parent/pom.xml
```diff
@@ -956,7 +928,7 @@
       <id>update-branding-plugins-about.mappings</id>
       <activation>
         <file>
-          <exists>${basedir}/about.mappings</exists>
+          <exists>about.mappings</exists>
```
mickaelistria, line 931: This doesn't seem related. Can you please skip this part?

mickaelistria (commented):
Have you evaluated how much longer build takes with this change? API Check is a long operation, first to provision and then to run.

eclipse-platform-parent/pom.xml
HannesWell: That's right, thats just a little simplification. Reverted it to keep this commit narrowed.

HannesWell:
> Have you evaluated how much longer build takes with this change? API Check is a long operation, first to provision and then to run.

Not in detail. But the API-description generation only takes a few second (3-5s in the cases I checked in PDE) and is currently enabled when building all Plug-in Projects that have the api-nature anyways. I could not find any use of the `longnotexistingproperty` to disable it. But if you want I can revet that conditions removal.

Furthermore the PDE-API description generator skips the generation at the moment any way if the api-nature/builder is absent. That has to be changed in order to have a real effect.
But since this then only effects those Plug-in projects without api-nature/build I assume the overall impact to me insignificant.
But it would allow to generate the API for new plug-ins in the initial release without API errors.

Alternativly we could add a setting to API tools to not consider the absence of the plug-in in the base-line as error, so the API-nature/builder can also be added for new projects with that setting being set to warning/error. Maybe this already exists. I have not yet checked that.

mickaelistria:
> Furthermore the PDE-API description generator skips the generation at the moment any way if the api-nature/builder is absent. That has to be changed in order to have a real effect.

The issue is that just generating this PDE-API description generator is already time consuming: it's not just running a mojo, it's using p2 to provision an Eclipse application and then starting it. Just willing to start this API Tools adds 3 to 5 seconds each I believe, so if this is applied to hundreds of bundles, we're getting close to 10 extra minutes.
I think we should really try to keep it lazy until we have a more efficient dedicated mojo for this operation eg #499 .

HannesWell:
> > Furthermore the PDE-API description generator skips the generation at the moment any way if the api-nature/builder is absent. That has to be changed in order to have a real effect.
> 
> The issue is that just generating this PDE-API description generator is already time consuming: it's not just running a mojo, it's using p2 to provision an Eclipse application and then starting it. Just willing to start this API Tools adds 3 to 5 seconds each I believe, so if this is applied to hundreds of bundles, we're getting close to 10 extra minutes.

But it this change only changes anything for those Plug-in projects that have the api-nature/build not enabled, which should be the vast minority (hopefully). So the summed up difference should be much lower.

mickaelistria:
> which should be the vast minority (hopefully)

I think it's for example the case for the ~hundred of test bundles.

akurtakov:
I would say logic should be enhanced to not create it for test bundles as api description for them makes no sense.

mickaelistria:
The current logic does that and prevents from creating the RCP app to run API analysis for test bundles (and others that do not have API).
If we manage to get a dedicated mojo that can check the natures/builders early, before creating the RCP app and running API analysis, then we can migrate without risk of performance penalty.

HannesWell:
> > which should be the vast minority (hopefully)
> 
> I think it's for example the case for the ~hundred of test bundles.

Indeed. I didn't consider them in that regard.

> If we manage to get a dedicated mojo that can check the natures/builders early, before creating the RCP app and running API analysis, then we can migrate without risk of performance penalty.

Yes I will close this and if https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/pull/499 is working properly this can be considered again.

PR #420, opened 2022-07-25
Use more recent BND annotations and include versioning annotations

laeubi:
There was on old version of BND annotations included and the version annotations are missing.


akurtakov:
@laeubi I don't see any issue with merging now. Is it all good with you too?

laeubi:
Yes, its just code-freeze period :-)

akurtakov:
I believe dependency updates are better delivered for M2 than M3 thus merging.

PR #434, opened 2022-08-03
Removes eclipse.team from aggregator

vogella:


vogella:
@laeubi feel free to update this PR or create your own.

vogella:
@laeubi I think we need to add the examples to the platform build 

[ERROR] Cannot resolve project dependencies:
[ERROR]   Software being installed: org.eclipse.sdk.examples.feature.group 4.25.0.qualifier
[ERROR]   Missing requirement: org.eclipse.sdk.examples.feature.group 4.25.0.qualifier requires 'org.eclipse.equinox.p2.iu; org.eclipse.compare.examples 0.0.0' but it could not be found

laeubi:
@vogella the problem is that the examples has compile warning(!) and the build do not allows warnings.

iloveeclipse:
> @vogella the problem is that the examples has compile warning(!) and the build do not allows warnings.

Where is this "compile warning" visible? 
I don't see it in https://ci.eclipse.org/platform/job/eclipse.platform.releng.aggregator/job/PR-434/3/consoleFull

akurtakov:
It is not fine that examples are no longer built as part of the merge as sdk.examples refers to them. These should be fixed.

laeubi:
> Where is this "compile warning" visible?

Just uncomment the examples in the `pom.xml` I also have fixed some warnings already on other places, I'll open a PR, at least I have had the problem that the build complains but in the IDE not all warnings are shown what makes it quite cumbersome to fix these.

akurtakov:
I'm on it.

akurtakov:
Looks good.

PR #437, opened 2022-08-04
Releng JIPP fails builds due to using Java 8

fredg02:
https://gitlab.eclipse.org/eclipsefdn/helpdesk/-/issues/1618

Use CentOS 8 docker image with JDK11


akurtakov:
Merging so we can get faster feedback from real build.

PR #448, opened 2022-08-07
Use Tycho-3.0's 'tycho.buildqualifier.format'-property instead of custom property

HannesWell:
Since Tycho 3.0 the 'tycho-packaging-plugin' provides the property 'tycho.buildqualifier.format' to specify its 'format' config property. 
Instead of configuring a custom property ('qualifier.format') with the same purpose, use Tycho's default property for that. Using more defaults if possible is IMHO a good thing.

When searching my platform-aggregator workspace I found no other references to `qualifier.format` than those adjusted in the platform-parent pom.

@akurtakov any objections?


akurtakov:
I'm fine with that.

Issue #450, opened 2022-08-08
Create infrastructure to build P-build for Java 19.


sravanlakkimsetti:

sravanlakkimsetti:
completed now

PR #463, opened 2022-08-10
Add JDT Plugins for P Build

SarikaSinha:
https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/issues/450


sravanQonway (changes requested):
Need to update pom.xml as well. Pom.xml should contain paths to these plugins

sravanlakkimsetti (changes requested):
Will require the location information to be updated in pom.xml. otherwise maven will not know what to build. Please update pom.xml with the plugin locations in the source.

sravanlakkimsetti (approved):

PR #471, opened 2022-08-12
Merge eclipse.platform.resources into eclipse.platform repository

laeubi:


akurtakov:
Let's get real test then :)

iloveeclipse:
We have a big problem: I don't know how that merge was done, but there is NO GIT HISTORY MORE at all, on any resource related code.

iloveeclipse:
This is a huge issue because it removes any context information, any bisecting of any change impossible.

iloveeclipse:
We lost a *decade* of information.

merks:
Oh dear yes, that's very bad!  It looks squashed:

image

And I think that's more than two decades of information.  We must repair this before moving on!

laeubi:
Can we please investigate this here: https://github.com/eclipse-platform/eclipse.platform/pull/76 as this PR has nothing to do with the merge at all, at least there it says that "merged 4,181 commits into [eclipse-platform:master](https://github.com/eclipse-platform/eclipse.platform)"

PR #472, opened 2022-08-12
[Freeze-Check] simplify URL and remove the dots

HannesWell:
With the dots being printed one that clicks on the printed link opens an invalid URL.


HannesWell:
The code freeze check now works as intended. No need to await the other builds.

PR #487, opened 2022-08-17
[Platform-Setup] temporarily add 'o.e.pde.ds.lib' to Platform-TP

HannesWell:
Because the 'o.e.pde.ds.lib' bundle has been removed from PDE and to be replaced by the official OSGi `o.o.service.component.annotations` bundles from Maven-Central https://github.com/eclipse-pde/eclipse.pde/pull/276 but the change that PDE-DS understands the new bundle-symbolic names https://github.com/eclipse-pde/eclipse.pde/pull/269/ is only available as I-build, this temporarily adds `o.e.pde.ds.lib` to the Platform-TP in order to keep DS-annotation processing working until https://github.com/eclipse-pde/eclipse.pde/pull/269/ is available in a release.

After the next (milestone?) release this could then be removed.

@laeubi, @merks any objections?

### DIFF
--- a/oomph/Platform.setup
+++ b/oomph/Platform.setup
@@ -138,12 +138,16 @@
           name="org.eclipse.pde.feature.group"/>
       <requirement
           name="org.eclipse.equinox.executable.feature.group"/>
+      <requirement
+          name="org.eclipse.pde.ds.lib"/>
       <repositoryList
           name="CBI+Orbit">
         <repository
             url="https://download.eclipse.org/cbi/updates/license"/>
         <repository
             url="https://download.eclipse.org/tools/orbit/downloads/latest-R"/>
+        <repository
+            url="${eclipse.4.24.release.p2}"/>
       </repositoryList>
     </targlet>
   </setupTask>


laeubi:
I don't understand this, does't the platform setup using the IBuidls anyways? Mine does and I can't remember doing anything special....

merks:
I'm slightly confused.  I'm guessing that maybe the PDE that's installed in the IDE needs the older bundle to be in the TP.  If not, what needs this older bundle to be there?  Even in this case, the SDK setup uses the Eclipse IBuild for the installation as well.

image

Is there really a problem when the IDE and the TP both come from the 4.25 I-Build?

HannesWell:
This is odd. My Platform-setup only uses the I-builds in the Modular-Target:
grafik


But not for the Eclipse itself that I use for developing and I use the latest Eclipse-IDE for Eclipse Committers (I felt addressed with that):
grafik

But when comparing the images I see that Ed you are using the `Eclipse SDK` Product from `Eclipse.org Application`.
I didn't knew that exists. Is that new? That's great. I always wanted that product (and even added the I-builds repo to my Installation's extra P2-Director but it failed to resolve). Looks like I should have used the `PlatformSDKConfiguration` instead of selecting the Product and projects manually.
But this explains why I was the only one that faces the problems of not compiling OSGI-service component annotations.

> Is there really a problem when the IDE and the TP both come from the 4.25 I-Build?

When both are from the 4.25 I-Build then everything is fine. My problem just was that my TP uses I-builds but my IDE is from 4.25 M2 at the moment. 
But I will fix that. Thank you guys. Then this is obsolete.

merks:
The SDK product was created 3 years ago when the platform developers listed all the reasons why they couldn't or wouldn't use Oomph.  One very good reason they gave is that they wanted to use their own I-Build to develop with so that they can eat their own dog food.   So with this product, one can pick the latest version and every day can run the p2 task to update to the latest I-Build.   The only downside is that the committers Product has more features installed, but of course the p2 tasks in the setups can ensure that any strictly required tools are also installed...

HannesWell:
Somehow I have completely missed that product. For the reasons you mentioned I always wanted that.
Will use that right this evening.

Issue #489, opened 2022-08-18
org.apache.commons.io got downgraded


sravanlakkimsetti:
<html>
<body>

org.apache.commons.io | 2.8.0.v20210415-0900 | 2.2.0
-- | -- | --

</body>
</html>

got downgraded in M3 candidate build this needs to be checked

cc @akurtakov @mickaelistria

sravanlakkimsetti:
This should be removed. We still have older version in our repository

sravanlakkimsetti:
screenshot of repo
![image](https://user-images.githubusercontent.com/77106773/185303256-01161759-ede1-40cc-b819-ead036f2981d.png)

sravanlakkimsetti:
Still a problem 
![image](https://user-images.githubusercontent.com/77106773/185919494-0fa2b586-b2ec-472a-a1fd-cb7a4e4e510e.png)

akurtakov:
@mickaelistria Can you take a look at this one?

merks:
The following PR's comment has details related specifically to this problem:

https://github.com/eclipse-equinox/p2/pull/136#issuecomment-1219404020

In particular there are no dependencies on the older version that are not satisfied by the newer version.  So why is it in the repo when it's not strictly required by anything?

I suspect strongly that this is a result of the repo is being assembled like this to include all dependencies (which was introduced during the 4.24 release cycle):
```
      <plugin>
        <groupId>org.eclipse.tycho</groupId>
        <artifactId>tycho-p2-repository-plugin</artifactId>
        <configuration>
        	<includeAllDependencies>true</includeAllDependencies>
        </configuration>
      </plugin>
```
I suspect that when Tycho is looking at package imports it resolves them to all bundles that provide a compatible version of the package, and then considers all those bundles to be dependencies, even if only one of them would be sufficient.

If that's the case, it seems the only way to eliminate this IU from the repo is to eliminate it from the target platform, but I don't think you can do that as long as you use put ` commons-fileupload`  in a `includeDependencyDepth="infinite` section.  I'd suggest moving it to a `includeDependencyDepth="none"` section.

mickaelistria:
> I'd suggest moving it to a includeDependencyDepth="none" section.

Although it would be more verbose, as we'd need to list transitive deps, it would actually allow to better control what we have in the .target and the final repo and would more easily allow to bump version of transitive deps as well.
[...]

PR #493, opened 2022-08-22
Do not add transitive Maven deps to .target

mickaelistria:
This allows a finer control of the dependencies, and also avoid
duplicated artifacts in different versions.
https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/issues/489


iloveeclipse:
For the clean diff, could you please revert white space changes and don’t re-arrange blocks?

mickaelistria:
> For the clean diff, could you please revert white space changes and don’t re-arrange blocks?

Sure, done.

PR #564, opened 2022-09-12
Ignore build.properties and fail on baseline payload differences

HannesWell:
As suggested in https://github.com/eclipse-equinox/p2/pull/113#issuecomment-1240392193, this ignores build.properties file when computing git-timestamps but in turn fails the build if the baseline and reactor contain artifacts with same qualifier but different content.

One drawback of the initial suggestion is that when building on Windows all text-files are detected as different due to the default line-endings difference between Windows and Linux.
So eventually the `baselineMode = failCommon` should either be wrapped in a profile (activated only on non-windows?) or could be set via the System-property `tycho.baseline`. But the latter would require to adjust many build files, so I think the former is more handy. The profile could be activated when the `CI` property exists, which is created by GH and Jenkins by default to indicate the CI environment.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -58,7 +58,8 @@ pipeline {
 						.*log,*/target/work/data/.metadata/.*log,\
 						*/tests/target/work/data/.metadata/.*log,\
 						apiAnalyzer-workspace/.metadata/.*log,\
-						eclipse.platform.releng.tychoeclipsebuilder/eclipse.platform.repository/target/repository/*'
+						eclipse.platform.releng.tychoeclipsebuilder/eclipse.platform.repository/target/repository/*,\
+						**/target/artifactcomparison/*'
 				}
 			}
 		}
--- a/eclipse-platform-parent/pom.xml
+++ b/eclipse-platform-parent/pom.xml
@@ -482,6 +482,7 @@
             <timestampProvider>jgit</timestampProvider>
             <jgit.ignore>
               pom.xml
+              build.properties
               .polyglot.build.properties
               .gitignore
             </jgit.ignore>
@@ -541,6 +542,7 @@
           <configuration>
             <baselineMode>warn</baselineMode>
             <baselineReplace>all</baselineReplace>
+            <writeComparatorDelta>true</writeComparatorDelta>
             <baselineRepositories>
               <repository>
                 <url>${comparator.repo}</url>
@@ -692,6 +694,7 @@
             </executions>
             <configuration>
               <defaultP2Metadata>false</defaultP2Metadata>
+              <baselineMode>failCommon</baselineMode>
             </configuration>
           </plugin>
         </plugins>


HannesWell:
@mickaelistria, @akurtakov, @laeubi, altough the originating PR has be submitted now without this, I still think this is useful because this prevents the same problem cases when non-test Plugins are affected and it also detects changes to the jar files that happen without a source code change.
For example when the return type of a method has changed from an Array to List and that method is called in an enhanced for loop. Then the source code does not have to be changed, but the compiled bytecode does. Without this change the compiled .class file will be replaced with the baseline version and will continue to call the method that returns an Array.
We had that in M2E, but IIRC it happened somewhere in the Eclipse Platform TLPs as well.

To complete this there will be necessary to adjust some exclusion patterns in sub-projects, because in some cases jar content changes, e.g. if code is generated and the generator adds a time-stamp (that can't be turned off).

akurtakov:
What is the status of this one?

HannesWell:

It staled because there was no Feedback if it is wanted. But if you think this is a good Idea I can continue this effort.
The main task is to set proper exclusion in the sub-modules where necessary.

akurtakov:
As this would requite a lot of manual work to set these exclusion I fail to see the benefit from it now.

eclipse-platform-parent/pom.xml
```diff
@@ -549,6 +551,7 @@
             <ignoredPatterns>
               <pattern>META-INF/ECLIPSE_.RSA</pattern>
               <pattern>META-INF/ECLIPSE_.SF</pattern>
+              <pattern>org/eclipse/ui/internal/dialogs/cpd/TreeManager$CheckListener.class</pattern>
```
mickaelistria, line 554: How is it related?

Jenkinsfile
```diff
@@ -3,6 +3,7 @@ pipeline {
 		timeout(time: 240, unit: 'MINUTES')
 		buildDiscarder(logRotator(numToKeepStr:'5'))
 		disableConcurrentBuilds(abortPrevious: true)
+		timestamps()
```
mickaelistria, line 6: Couldn't this be a separate PR?

HannesWell: Done in https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/pull/749.

eclipse-platform-parent/pom.xml
HannesWell: In this case the .class file changed without a change in the java file and I just temporarily excluded it in order to gather more of such cases quicker.
I check it back then and the change is actually not harmful, but the intention of this change is also to catch such changes in the jar that are caused by external factors.

HannesWell:
> As this would requite a lot of manual work to set these exclusion I fail to see the benefit from it now.

As described before (https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/pull/564#issuecomment-1279694377) this change allows the detect changes in the jars that are caused due to external factors without any changes in the sources and therefore did not cause a qualifier change.

At the moment those are not detected. In the worst case (as we had it in M2E), this can cause NoSuchMethod errors at runtime.

akurtakov:
So is work going here or it should be closed?

akurtakov:
I assume it's time to close this one.

PR #583, opened 2022-09-21
Revert "Moving to daily ecj snapshot"

sravanlakkimsetti:
Reverts eclipse-platform/eclipse.platform.releng.aggregator#581

Reverting this one as this is causing PR validation to fail


sravanlakkimsetti:
maven is looking for ecj in tycho snapshots instead of eclipse-snapshots. because of this ecj is not getting downloaded. Probably snapshots needs to enabled in the repositories. 

bit busy with java 19 merge to master. Will investigate little later

akurtakov:
Is the same groupId used in both places?

mickaelistria:
I'm late on this train, but that's indeed a great idea to try that! Too bad it doesn't work on 1st try; it makes sense to complete the integration of Java 19 work on master first before re-trying this.

Issue #590, opened 2022-09-26
Eclipse 4.26 Prerequisites: EMF


SDawley:
Tracking issue for EMF.

M1 is scheduled for September 30th.

merks:
The following build is contributed to SimRel for M1.

https://download.eclipse.org/modeling/emf/emf/builds/milestone/S202209250929

SDawley:
Any updates for M2?

merks:
There is no new build for M2.

akurtakov:
Is new EMF build for 4.26 planned?

merks:
I'll decide by the end of the day and either way I'll provide the release repository URL then.

merks:
This is the final build for 2022-12:

https://download.eclipse.org/modeling/emf/emf/builds/release/2.32

PR #592, opened 2022-09-27
4.26 prereqs update EMF

SDawley:
Tracked in #590


merks:
For the past release, this stayed open for the whole release cycle because you need updates again for M2, M3, RC1, and RC2...

sravanlakkimsetti:
We keep the issue open. PR

merks:
Silly me.  Sorry for the noise.  😕

PR #598, opened 2022-09-28
Use slf4j from upstream Maven instead of Orbit

mickaelistria:
And bump to latest release

### DIFF
--- a/eclipse.platform.releng.prereqs.sdk/eclipse-sdk-prereqs.target
+++ b/eclipse.platform.releng.prereqs.sdk/eclipse-sdk-prereqs.target
@@ -29,11 +29,6 @@
       <unit id="org.junit" version="4.13.2.v20211018-1956"/>
       <unit id="org.junit.source" version="4.13.2.v20211018-1956"/>
 
-      <!-- specific (additional) to equinox sdk -->
-      <!-- Upstream maven artifact uses different slf4j.api SymbolicName. Customers needs to be updated -->
-      <unit id="org.slf4j.api" version="1.7.30.v20200204-2150"/>
-      <unit id="org.slf4j.api.source" version="1.7.30.v20200204-2150"/>
-
       <unit id="org.apache.lucene.core" version="8.4.1.v20200122-1459"/>
       <unit id="org.apache.lucene.analyzers-smartcn" version="8.4.1.v20200122-1459"/>
       <unit id="org.apache.lucene.analyzers-common" version="8.4.1.v20200122-1459"/>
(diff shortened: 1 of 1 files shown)


HannesWell:
I assume it is very likely that we can't use SLF4J 2 yet because dependencies (if not of Platform then in the SimRel) are not yet ready for SLF4J 2.
In the meantime we could use [SLF4J 1.7.36](https://mvnrepository.com/artifact/org.slf4j/slf4j-api/1.7.36) from Maven-Central. If everything else is then ready we just have to bump the version here.

mickaelistria:
If Platform can move to slf4j-2, it will. SimRel will follow.

mickaelistria:
@HannesWell it looks like -unlike Orbit- upstream slf4j defines in its MANIFEST that 1 impl is required. OK, why not... I've added the nop one; but now it will cause that the nop will be included in Eclipse IDE. Does that still allow other bindings to work? I'm a bit confused by the API requiring 1 impl, I'm afraid it causes limitations in OSGi.

HannesWell:
> @HannesWell it looks like -unlike Orbit- upstream slf4j defines in its MANIFEST that 1 impl is required. OK, why not... I've added the nop one; but now it will cause that the nop will be included in Eclipse IDE. Does that still allow other bindings to work? I'm a bit confused by the API requiring 1 impl, I'm afraid it causes limitations in OSGi.

It will cause the nop-binding to be included in the Eclipse-Platform p2-repo, that's right. But as long it is only in the repo (and not in a feature) it should be ignored when it comes to building/launching products or assembling repos with deps, if another binding with 'higher priority' is available. For example if somebody specifies another binding in a feature included in a product or if an (transitively included) plugin requires a specific binding (like `m2e.logback` requires logback) and the resolver does a good job in minimizing the number of bundles, then the binding with 'higher priority/demand' should be used and the noop one should be ignored. Because M2E requires logback (as mentioned above) the noop binding probably/hopefully will not even make it into the SimRel repo and only logback and probably/maybe log4j will be present. Log4J is also required by other Eclipse projects like XText or Passage, like M2E requires logback.


that does not necessarily mean that this

Btw. could you please sort out the formatting changes, so that we get a smaller change? With that many changes it is hard to understand what has actually been changed. If the target should be reformatted, I suggest to do that in a dedicated commit.

mickaelistria:
Thanks for the insights. I've fixed the formatting and other unrelated changes. Is it OK to merge in your opinion?

HannesWell (approved):
> Thanks for the insights. I've fixed the formatting and other unrelated changes. Is it OK to merge in your opinion?

Yes it looks good, thanks!
Nevertheless we should keep an eye on the slf4j story in upcoming SimRel milestones/RCs until this has settled, to make sure it is as expected.

The main points are:
- only one `slf4j.api` bundle. Probably the latest from Maven-Central.
- Only one logging implementation per framework (logback, log4j-2, log4-1) and ideally only those from Maven-Central to make sure the approaches for binding/bridging used in the Orbit-bundles and those from Maven-Central is not mixed.

PR #600, opened 2022-09-28
Use javax.annotation from Maven Central instead of Orbit

mickaelistria:


HannesWell:
javax.annotation from Orbit currently exports all javax.annotation packages in version 1.3.5, while the version available at Maven-Central only has:
```
Export-Package: javax.annotation;version="1.3.2",javax.annotation.secu
 rity;version="1.3.2",javax.annotation.sql;version="1.3.2"
```
This step back could be problematic as many Manifests probably import that package with a lower bound of 1.3.5, which isn't satisfied by the artifact from M-C. No glue where Orbit got the more recent version from. Maybe just some Orbit specific changes?

mickaelistria:
> javax.annotation from Orbit currently exports all javax.annotation packages in version 1.3.5, while the version available at Maven

This was a similar situation with ICU4J.
We'll try to adopt upstream artifacts in Platform anyway, fixing relevant dependency definition; and then we'll switch.

merks:
Here are dependencies on the train:

image

Probably it came from here:

https://repo1.maven.org/maven2/jakarta/annotation/jakarta.annotation-api/1.3.5/

That's what the published POMs point at:

image

It looks like 2.1.1 is the latest version.

vogella:
IIRC the 2 version changes from javax.annotation to jakarta.annotation package name, so the last version for javax.annotation should be https://repo1.maven.org/maven2/jakarta/annotation/jakarta.annotation-api/1.3.5/

merks:
@vogella Yes, starting with 2.0 the package name is changed.

image

HannesWell:
> IIRC the 2 version changes from javax.annotation to jakarta.annotation package name, so the last version for javax.annotation should be https://repo1.maven.org/maven2/jakarta/annotation/jakarta.annotation-api/1.3.5/

Thanks for that pointer! Shouldn't we use that version then? This should make the transition to Maven-Central straight forward.
When switching to jakarta many imports and package names have to be adjusted anyway just like the version adjustment.

vogella:
> > IIRC the 2 version changes from javax.annotation to jakarta.annotation package name, so the last version for javax.annotation should be https://repo1.maven.org/maven2/jakarta/annotation/jakarta.annotation-api/1.3.5/
> 
> Thanks for that pointer! Shouldn't we use that version then? This should make the transition to Maven-Central straight forward. When switching to jakarta many imports and package names have to be adjusted anyway just like the version adjustment.

I think we should switch to the 1.3.5 version, moving to the jakarta namespace would require a coordinated effort and most likely a major version increase in the affected plug-ins.

mickaelistria:
Does e4 understand the `jakarta.annotation` stuff?

vogella:
> Does e4 understand the `jakarta.annotation` stuff?

No

merks:
Given nothing uses that namespace in the Java source code, I would assume not.  Keep in mind that these annotations are retained at runtime, e.g., for org.eclipse.debug.ui.launchview.internal.view.LaunchViewImpl.createView(Composite, MPart):

image

So I suppose probably both (either) should be supported at some point for quite some time...

HannesWell:
There is a similar jakarta bundle for javax.inject:
https://repo1.maven.org/maven2/jakarta/inject/jakarta.inject-api/1.0.5/

That could be considered as well. Either in this PR or in another one.

akurtakov:
We will have to ship and support in code both javax.annotation and jakarta.annotation for at least our deprecation period (2 years).

mickaelistria:
OK, so let's stick to progressively replace Orbit's javax.annotation by Maven one as part of this ticket. Further changes to make e4 capable of using jakarta.annotation should be tracked separately.

HannesWell:
> We will have to ship and support in code both javax.annotation and jakarta.annotation for at least our deprecation period (2 years).

The javax-packages are used very widely and I think it will take a long time to switch jakarta completely (mainly because of dependencies). I'm not sure if it is then necessary for e4 to still support javax, but those bundles will be present as dependencies probably for a long time. And in case it is just about checking for two names instead of one, supporting it longer would probably not hurt much.


That being said should this PR also be used to migrate from java.inject to jakarta.inject-api-1.0.5 as mentioned above? Or do you want to do that step by step?

mickaelistria:
> That being said should this PR also be used to migrate from java.inject to jakarta.inject-api-1.0.5 as mentioned above? Or do you want to do that step by step?

I believe we need 1 issue in di to support `jakarta.inject` and this will require to add `jakarta.inject` to the .target. Then, once it's working, we can consider progressively dropping `javax.inject` in Platform and when there are no references left, remove it from .target.

akurtakov:
@mickaelistria What's next here?

mickaelistria:
> @mickaelistria What's next here?

Lots of things:
* This PR needs to be updated to include `jakarta.annotation-api` 1.3.5
* The "Import-Package: javax.annotation" need to have the version restriction trimmed
* The "Require-Bundle: javax.annotation" need to be replaced by "Require-Bundle: jakarta.annotation-api;bundle-version="[1.3.5,2.0.0)"

I'm not sure that Orbit's javax.inject vs Maven Central's jakarta.annotation-api can easily cohabitate. I am not skilled enough with annotation internat to know whether it's just the name that matters (so one or the other or both would work as annotations have the same name), or if there is classloading checks involved, so one bundle linked to javax.annotation wouldn't be compatible with jakarta.annotation-api and the other way round.

HannesWell:
> * The "Require-Bundle: javax.annotation" need to be replaced by "Require-Bundle: jakarta.annotation-api;bundle-version="[1.3.5,2.0.0)"

In this process I suggest to convert `Require-Bundle: javax.annotation` to corresponding Import-Package entries.

akurtakov:
Is it time to give up on this one? IMHO, whenever someone has time to spend on this it would be best to have:
* Parallel available and usable javax.annotation and jakarta.annotation
* Clear warning that javax.annotation is deprecated for removal at runtime
* Port code in the SDK to use jakarta.annotation

HannesWell:
>     * Port code in the SDK to use jakarta.annotation

As long as it only means to replace the `javax.annotation` string by `jakarta.annotation` I think that's not to difficult.

The more difficult part is probably to make the code that processes the annotation work with jakarat.
On the other hand if that just checks for a name or the annotation class it is probably just an extension as well, but one needs to know at which places the annotations are processed.



> > @mickaelistria What's next here?
> 
> Lots of things:
> 
>     * This PR needs to be updated to include `jakarta.annotation-api` 1.3.5
> 
>     * The "Import-Package: javax.annotation" need to have the version restriction trimmed
> 
>     * The "Require-Bundle: javax.annotation" need to be replaced by "Require-Bundle: jakarta.annotation-api;bundle-version="[1.3.5,2.0.0)"


Can you tell why the version restriction should be removed?

mickaelistria:
> Can you tell why the version restriction should be removed?

IIRC, because the upstream javax.annotation bundle doesn't add version to package exports.

HannesWell:
> > Can you tell why the version restriction should be removed?
> 
> IIRC, because the upstream javax.annotation bundle doesn't add version to package exports.

I checked it and they support it, but there are a few Plugins that require javax.inject/annotation bundles... But that can easily be fixed.

I just created https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/issues/1056 to coordinate the steps to eventually migrate javax.inject/annotations to Jakarta.

PR #601, opened 2022-09-28
Revert "Remove bcprov from Maven Central from target platform"

mickaelistria:
bcprov is now IP-cleared from MAven Central!
This reverts commit f503ae14579b9d418829b93abf5752453c380ab3.


mickaelistria:
If we can get all builds fine, I think it's worth merging it for M1, so that the Orbit bundles wouldn't be leaking from Platform.
cc @akurtakov @sravanlakkimsetti

PR #611, opened 2022-10-03
Add two new properties for modules to control pom descriptor

laeubi:
@akurtakov WDYT? I'd like to use that in P2 as an incubator and it seems better to be configured in the parent so it can be enabled by a simple property than to be override the whole execution in the child modules.


akurtakov (approved):
Fine with me.

Issue #610, opened 2022-10-03
The pcpgp and bcprov bundles are signed only with JCE certificates and are not PGP signed


merks:
We're faced with this familiar problem again, one that we saw from the aborted changes during the previous release cycle:

![image](https://user-images.githubusercontent.com/208716/193567989-0c62d635-ed30-4788-a50f-2e3628b9c3a2.png)

For a refresher, here is the past occurrence of this issue:

https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/issues/418

I did enable checking for both a PGP signature and the jar signature and if either is trusted to not show a trust prompt as implemented for this issue:

https://github.com/eclipse-equinox/p2/issues/121

I have a feeling that perhaps nothing was done in Tycho to enable PGP signing a bundle even when it's jar-signed.  As was anticipated, that is needed as a prerequisite to consuming JCE-signed bundles directly from Maven.

Please PGP sign these two bundles or contribute these bundles Orbit and let Orbit sign them.

@jonahgraham  FYI

mickaelistria:
I suspect this is a Tycho configuration issue that skips signatures for jars that are already jarsigned. We may hardcode a rule in Tycho about considering that JCE certificates are to be ignored on that aspect.

PR #612, opened 2022-10-04
Force PGP signature for bcpg and bcprov

mickaelistria:


merks:
It would have been good for this commit to reference the issue that is fixes:

https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/issues/610

PR #614, opened 2022-10-06
Add a profile to generate and attach javadoc artifacts

laeubi:
When deploying artifacts to maven it is common to provide also javadoc artifacts. This adds a profile that could be used in sub-projects to accomplish this.

I was able to use this for equinox to generate javadoc artifacts, but though it might be more appropriate to have this in the master so other projects might easily reuse that as well. This also will allow us to catch javadoc errors earlier than with the current "doc-build".

### DIFF
--- a/eclipse-platform-parent/pom.xml
+++ b/eclipse-platform-parent/pom.xml
@@ -1151,6 +1151,50 @@
         <surefire.moduleProperties>--add-modules=ALL-SYSTEM</surefire.moduleProperties>
       </properties>
     </profile>
+	<profile>
+		<id>javadoc</id>
+		<build>
+			<plugins>
+				<plugin>
+					<groupId>org.apache.maven.plugins</groupId>
+					<!-- Must use older version see https://issues.apache.org/jira/projects/MJAVADOC/issues/MJAVADOC-707 -->
+					<artifactId>maven-javadoc-plugin</artifactId>
+					<version>3.0.1</version>
+					<executions>
+						<execution>
+							<id>attach-javadocs</id>
+							<phase>package</phase>
+							<goals>
+								<goal>jar</goal>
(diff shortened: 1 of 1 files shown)


laeubi:
@akurtakov any concerns or can we merge this?

mickaelistria:
To be honest, I don't see the point of shipping Javadoc as artifacts when we already ship sources (that all IDE can infer Javadoc from) and when we already publish the javadoc.

laeubi:
@mickaelistria javadoc artifacts are mandatory for publishing to maven central see: https://maven.apache.org/repository/guide-central-repository-upload.html

Currently platform only publish "dummy" javadoc see for example:
https://repo1.maven.org/maven2/org/eclipse/platform/org.eclipse.equinox.preferences/3.10.1/org.eclipse.equinox.preferences-3.10.1-javadoc.jar that only contains a readme with:
> Corresponding javadoc can be found in artifact org.eclipse.platform:org.eclipse.platform.doc.isv

what is the second goal here, these `doc.isv` builds are interdependent from the usual verification builds (and horrible to manage), so errors in the javadoc are often not detected before these builds runs and someone takes a look and probably complain about it.

Integrating those in the usual builds (with Jenkins collecting those) would help us getting better results and avoid multiple circles.

Assuming we produce javadoc for all platform bundles, one might even think to only have an aggregating step and get rid of the unstable `doc.isv` stuff all together.

akurtakov:
Ideally if we can replace the javadoc step in platform.doc.isv with the aggregated result it would be perfect.

laeubi:
> Ideally if we can replace the javadoc step in platform.doc.isv with the aggregated result it would be perfect.

start small think big ... so we can merge this?

akurtakov (approved):

PR #621, opened 2022-10-11
Publish Java 19 tests

akurtakov:
As Java 19 is released, Java 18 is EOL already.


akurtakov:
That would not cause any change in maven build. Merge and trigger I-build.

PR #623, opened 2022-10-11
Update runTests2.xml so jobs can be run from a folder

SDawley:
Changing runtests2.xml to use JOB_BASE_NAME instead of JOB_NAME since the latter includes the jenkins folder while JOB_BASE_NAME is just the job name, and should work for both the new and old locations.

Also updated unit test groovy files since I moved collectResults to a central location.


SDawley:
@sravanlakkimsetti After merging this change I need to run https://ci.eclipse.org/releng/view/Miscellaneous/job/Publish_shell_scripts_for_releng_jobs correct?

PR #625, opened 2022-10-14
Update target platform to bouncycastle 1.72

akurtakov:
Makes sure target platform is not using older version than Orbit. See https://github.com/eclipse-equinox/p2/issues/153 for details.


iloveeclipse:
Note, this will break the build if p2 change https://github.com/eclipse-equinox/p2/pull/154 is not merged

akurtakov:
So we merge it first.

akurtakov:
P2 patch merged so let's merge this one too.

PR #636, opened 2022-10-19
Use https links in JenkinsJobs

akurtakov:


akurtakov:
These are executed only by I-builds so verification here will not test it.

PR #639, opened 2022-10-21
Use commons-jxpath from Maven Central

mickaelistria:


merks:
FYI, it looks like these places will all need attention because of the bundle name change:

image

Some launchers too but I don't know if that matter...

mickaelistria:
Let's abandon and do things proper: removing jxpath https://github.com/eclipse-platform/eclipse.platform.ui/issues/423

PR #647, opened 2022-10-26
Update collect results job to use java17 with ant and new job structure

sravanlakkimsetti:


akurtakov:
Should we trigger new build to test it out?

sravanlakkimsetti:
> Should we trigger new build to test it out?

Not required. rerunning collect results job should be sufficient

PR #650, opened 2022-10-28
Add ability to run results collection locally

SDawley:
Added a script to run results collection locally for testing etc. 
Also removed properties that weren't being used anymore so I changed the xml scripts quite a bit.

@sravanlakkimsetti  and @akurtakov I'd appreciate if someone could try pulling this branch and using the script to see how it works. I tested using https://ci.eclipse.org/releng/job/AutomatedTests/job/ep426I-unit-cen64-gtk3-java11/19/ and  https://ci.eclipse.org/releng/job/ep426I-perf-lin64/61/ but would like a sanity check. You'll have to use the local files option to use the new xmls.

I'd also like to choose a single location for the xml files, be that cje-production/scripts/ or eclipse.platform.releng.tychoeclipsebuilder/eclipse/buildScripts/ I don't really know which is better.

I'm also not sure if we still use publish2.xml and would like to remove it if we don't. @sravanlakkimsetti do you happen to know?


sravanlakkimsetti:
> I'd also like to choose a single location for the xml files, be that cje-production/scripts/ or eclipse.platform.releng.tychoeclipsebuilder/eclipse/buildScripts/ I don't really know which is better.

eclipse.platform.releng.tychoeclipsebuilder/eclipse/buildScripts/ was the old location before OKD move. if possible please consolidate the scripts to cje-production/scripts/

> I'm also not sure if we still use publish2.xml and would like to remove it if we don't. @sravanlakkimsetti do you happen to know?

I believe publish.xml is used by unit tests and publish2.xml is for performance tests. But not sure. lets remove publish2.xml and fix the issues that arose from that.

sravanlakkimsetti (approved):

sravanlakkimsetti (approved):

SDawley:
@sravanlakkimsetti Performance results also uses publish.xml

Is the top level production folder also something that can be consolidated into cje-production?

PR #652, opened 2022-11-03
Manage version of maven-toolchains-plugin.

akurtakov:
It's used for running tests in jdt.core and jdt.debug but all at different version.
This is first step towards standardizing.


akurtakov:
Not waiting for the verification build as it will not prove anything here.

PR #653, opened 2022-11-06
Add metatype annotation to target

laeubi:
The [Metatype Service Specification](https://docs.osgi.org/specification/osgi.cmpn/7.0.0/service.metatype.html) defines a set of [annotations](https://docs.osgi.org/specification/osgi.cmpn/7.0.0/service.metatype.html#service.metatype-metatype.annotations) to allow defining possible configuration options of a service as code.

These annotations are only retained at class level and can the used at build time to generate the proper XML definition, we should enable PDE users to use these annotations as well, so this information can then be used for example in Tycho or BND builds.

### DIFF
--- a/eclipse.platform.releng.prereqs.sdk/eclipse-sdk-prereqs.target
+++ b/eclipse.platform.releng.prereqs.sdk/eclipse-sdk-prereqs.target
@@ -397,6 +397,12 @@
 				  <type>jar</type>
 			  </dependency>
 			  <dependency>
+				    <groupId>org.osgi</groupId>
+				    <artifactId>org.osgi.service.metatype.annotations</artifactId>
+				    <version>1.4.1</version>
+				    <type>jar</type>
+			  </dependency>
+			  <dependency>
 				  <groupId>org.osgi</groupId>
 				  <artifactId>org.osgi.service.component</artifactId>
 				  <version>1.5.0</version>


laeubi:
@akurtakov any concerns?

akurtakov:
Not at all. Go for it.

laeubi:
@akurtakov how long does it takes to be used in the sub-repositories e.g. here

https://github.com/eclipse-pde/eclipse.pde/pull/380

will it be with the next Ibuild?

akurtakov:
An hour

HannesWell:
> @akurtakov how long does it takes to be used in the sub-repositories e.g. here
> 
IIRC you have to wait until the Jenkins build of the new commit on the master branch has succeeded.

akurtakov:
Changes to target platform and parent pom are used from maven repository and they are published using https://ci.eclipse.org/releng/job/deploy-eclipse-platform-parent-pom/ and https://ci.eclipse.org/releng/job/deploy-eclipse-sdk-target-pom/ which both run hourly.

HannesWell:
Thanks Aleks, then I had a false assumption about that.
Wouldn't it be beneficial to do both builds in the beginning of the master-branch build of this repo? Then the build would only run on demand and the result would be available much quicker.

akurtakov:
I would love that tbh :). This has served us that way for the last ~10 years and I have to admit I haven't thought about improving this one at all. But 2 less jenkins jobs is too nice of a proposal, do you think you can help with it?

HannesWell:
> I would love that tbh :). This has served us that way for the last ~10 years and I have to admit I haven't thought about improving this one at all. But 2 less jenkins jobs is too nice of a proposal, do you think you can help with it?

Glad you like it. :) Yes I can look into it and propose a patch this evening.

HannesWell:
> I would love that tbh :). This has served us that way for the last ~10 years and I have to admit I haven't thought about improving this one at all. But 2 less jenkins jobs is too nice of a proposal, do you think you can help with it?

Created https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/pull/657.

laeubi:
I still see errors in the SDK though, is anyone able to update its SDK and see this in the target platform state?

HannesWell:
> I still see errors in the SDK though, is anyone able to update its SDK and see this in the target platform state?

I have not tried it yet, but aren't the Oomph targlets only pulling from the I build repo what any Plugin requires. IIRC the sdk-target ist only used for builds, so Something first has to require it and then it becomes available in the workspace target.

merks:
Yes, only requirements are pulled in.  So you can add such a requirement and then resolve the TP again (Help -> Perform Setup Tasks).

laeubi:
I added `biz.aQute.bndlib` (what also is new) and `org.osgi.service.metatype.annotations` as require bundle to the pde.core and then performed the setup task, then eclipse asks for restart (because new software was installed).
But then still the target platform stae do no contains these and PDE complains that the bundles can not be found.

merks:
This appears already in the Oomph'd IDE:

image

When I add the other one:

image

it fails to find it:

image

Which I think it correct because I don't see in the I-Build:

image

laeubi:
This was added by:

https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/pull/654

but is not at the ibuilds because it is not referenced anywhere, but if I can't use it in the SDK I can't reference it and thus it will not be on the ibuilds because it is not referenced anywhere, but if I can't use it in the SDK I can't reference it and thus it will not be on the ibuilds ... :-)

merks:
A chicken and egg problem...  You could of course manually activate the *.target... 

Or use this trick.  Navigate -> Open Setup -> Workspace and create a Targlets task there...  You can copy the following text and paste it to the Workspace object:
```
<?xml version="1.0" encoding="UTF-8"?>
<setup.targlets:TargletTask
    xmi:version="2.0"
    xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
    xmlns:setup.targlets="http://www.eclipse.org/oomph/setup/targlets/1.0"
    xsi:schemaLocation="http://www.eclipse.org/oomph/setup/targlets/1.0 http://git.eclipse.org/c/oomph/org.eclipse.oomph.git/plain/setups/models/SetupTarglets.ecore">
  <targlet name="My Targlet">
    <requirement
        name="biz.aQute.bnd.util"/>
    <requirement
        name="biz.aQute.bndlib"/>
    <repositoryList>
      <repository
          url="https://download.eclipse.org/staging/2022-12"/>
    </repositoryList>
  </targlet>
</setup.targlets:TargletTask>
```
I've tested that this works.  Later you can delete the task, or use the context menu to disable it so you can use it again the next time there is such a problem.

laeubi:
Thanks @merks this finally has worked!

PR #654, opened 2022-11-07
Add BND lib to the target

laeubi:
m2eclipse already uses BND for a while to generate missing manifest for maven dependencies and PDE can (re-)use BND as well for different tasks.

This also moves the bnd annotations to the new BND section and adds org.osgi.service.repository API bundle as it is a dependency of BND lib.

@akurtakov WDYT? 
FYI @HannesWell


akurtakov:
Do you plan to implement some functionality in PDE using bndlib?

laeubi:
> Do you plan to implement some functionality in PDE using bndlib?

Yes specifically [metatype generation](https://github.com/eclipse-pde/eclipse.pde/issues/381) as I'm to lazy to reimplement it myself ;-)
Also DS might be an option (as discussed here 
- https://github.com/eclipse-pde/eclipse.pde/issues/36

I also though about some kind of "optimize the manifest", that uses BND to generate a new one based on the current state, e.g. to automatically convert required bundles into imported packages and alike...

akurtakov:
Cool. Go for it.

PR #657, opened 2022-11-08
Deploy platform-parent pom and platform-releng target in 'master' build

HannesWell:
As discussed in https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/pull/653#issuecomment-1306869654 this PR changes the Jenkins-Pipeline of the `eclipse.platform.releng.aggregator` to deploy the `eclipse-platform-parent/pom.xml` and `eclipse.platform.releng.prereqs.sdk/eclipse-sdk-prereqs.target` in the beginning of the master-branch build of this repo.
This makes the following two Jenkins jobs obsolete and also speeds up the deployment of the mentioned files but at the same time reduces the number of deployments:
- https://ci.eclipse.org/releng/job/deploy-eclipse-platform-parent-pom/
- https://ci.eclipse.org/releng/job/deploy-eclipse-sdk-target-pom/

From looking at those jobs it looks like nothing more has to be done. Maybe only some credentials are missing, but we will see.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -25,6 +25,15 @@ pipeline {
 				sh 'git submodule foreach "git fetch origin master; git checkout FETCH_HEAD"'
 			}
 		}
+		stage('Deploy eclipse-platform-parent pom and eclipse-sdk target') {
+			when {
+				branch 'master'
+			}
+			steps {
+				sh 'mvn clean deploy -f eclipse-platform-parent/pom.xml'
+				sh 'mvn clean deploy -f eclipse.platform.releng.prereqs.sdk/pom.xml'
+			}
+		}
 		stage('Build') {
 			steps {
 				withCredentials([string(credentialsId: 'gpg-passphrase', variable: 'KEYRING_PASSPHRASE')]) {


HannesWell:
@akurtakov looks like that worked even without further addo.
I checked the snapshot repos:
- https://repo.eclipse.org/content/repositories/eclipse-snapshots/org/eclipse/eclipse-platform-parent/4.26.0-SNAPSHOT/
- https://repo.eclipse.org/content/repositories/eclipse-snapshots/org/eclipse/eclipse-sdk-prereqs/4.26.0-SNAPSHOT/

Each one contains the results of the deployments at 18:38 GMT from this replayed verification build (where I disabled the when-condition for testing purpose).

So for me it looks like this works as expected and is ready for submission. Actually the initially mentioned jobs can then be disabled immediately.

akurtakov:
Thanks @HannesWell ! Let's merge and see.

mickaelistria:
> This makes the following two Jenkins jobs obsolete and also speeds up the deployment of the mentioned files but at the same time reduces the number of deployments:

I don't think it makes them obsolete. We often have needs of publishing a change in parent or .target and get very fast feedback in subsequent CI jobs. The I-Build runs not often enough and takes too long to provide this fast feedback loop; moreover in some cases, some changes in .target or parent pom do require further actions on other repos. So the I-Build you'll trigger just after an update will fail.
Overall, it's good thing to publish as part of the I-Build, but disabling a more immediate publishing will cause issue.

akurtakov:
It's not about publishing as part of the I-build but as parter of the verification build of master branch right after commit.

mickaelistria:
Ah OK, my bad. Then I now agree it's OK to remove the 2 other jobs.

sravanlakkimsetti:
- https://ci.eclipse.org/releng/job/deploy-eclipse-platform-parent-pom/
- https://ci.eclipse.org/releng/job/deploy-eclipse-sdk-target-pom/

are used to deploy not master branch versions but the maintenance branches as well. Lets keep them enabled 

As per my understanding with this change we publish within 10 mins of a commit. 
I have another suggestion in the above mentioned jobs we are polling scm every hour. If we switch to github hook the above jobs will launch immediately after a commit.

akurtakov:
@HannesWell can you extend it to work for maintenance branches too?

HannesWell:
> As per my understanding with this change we publish within 10 mins of a commit. 

Exactly. The deployment of those two files happens as very first build step, right after the (recursive submodule) checkout has completed. So the runtime of the latter is the only duration between the submission of a new commit to the master and the deployment. Looking at the latest executions these are between ~4 and ~9min:
https://ci.eclipse.org/platform/job/eclipse.platform.releng.aggregator/job/master/

> I have another suggestion in the above mentioned jobs we are polling scm every hour. If we switch to github hook the above jobs will launch immediately after a commit.

That would be feasible, but personally I think it is not worth the effort and I would sacrifice this few minutes for a simpler build infrastructure.
If it is really that important I can move the `Deploy eclipse-platform-parent pom and eclipse-sdk target` upwards as first stage to save some more seconds.

> @HannesWell can you extend it to work for maintenance branches too?

In order to make this change work with the maintenance branches this change has to be back-ported to each maintanance-branch and the condition has to be adapted for each branch. E.g. for `R4_25_maintenance` 
```
when {
	branch 'master'
}
```
becomes 
```
when {
	branch 'R4_25_maintenance'
}
```
If you think it is necessary I can do that for the latest maintenance branch(i.e. `R4_25_maintenance`) as demonstration, but doing it now for all seems not to be valuable for me. I would only do that on the demand for the corresponding branch.

Nevertheless I suggest to adapt the when-condition in the preparation of future maintenance branches. Is there a TO-DO list that has to be updated?

HannesWell:
> 
> Would be good so new branches get support when branching. And overtime it will be less and less needed for old branches so we can forget about the current job running for them only.

Sounds good. Created https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/pull/659.

We should then disable the existing other two jobs for the current master.
Thinking about it again, for the existing maintenance branches the web-hook suggestion would actually be not too bad. Or one simply back-ports this change for the existing maintenance branches if necessary.

HannesWell:
Since #659 is merged now, how should we proceed with regard to the to Jenkins jobs that are only required for old maintenance branches?
- Keep them but disable them for the master?
- Trigger them only on web-hook events on maintenance branches?
- Just delete them and back-port this change if new changes are applied to the maintenance branches?

I cannot help by any of those options, but didn't want it to become forgotten.

Issue #680, opened 2022-11-21
No download links anymore on build page since I20221120-0600


iloveeclipse:
See
- https://download.eclipse.org/eclipse/downloads/drops4/I20221120-0600/
- https://download.eclipse.org/eclipse/downloads/drops4/I20221120-1800/

Last "good" one is https://download.eclipse.org/eclipse/downloads/drops4/I20221119-1800/

This one (manually created link) worked for me:
https://download.eclipse.org/eclipse/downloads/drops4/I20221120-1800/download.php?dropFile=eclipse-SDK-I20221120-1800-linux-gtk-x86_64.tar.gz

The download links, if manually created, still work, so it is "just" the download page issue, but very ugly one.

![image](https://user-images.githubusercontent.com/964108/202991153-e13f97b7-6e41-41d2-8461-d1e78cf82b5a.png)

@SDawley : please check this ASAP.

SDawley:
Looking

iloveeclipse:
Thanks, looks like #683 fixed the problem, at least I see download links in https://download.eclipse.org/eclipse/downloads/drops4/I20221122-0600/

PR #683, opened 2022-11-21
Fix: Updating publish.xml to use a variable instead of relying on current directory

SDawley:
Should address:
https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/issues/680

https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/blob/master/cje-production/mbscripts/mb300_gatherEclipseParts.sh already passes -DEBuilderDir to helper.xml


akurtakov:
Merging for the 6 AM build.

PR #693, opened 2022-11-28
Fix git branch --set-upstream no longer recognized

akurtakov:
The new parameter name is --set-upstream-to.
Remove useless parts of the script while at it.


akurtakov:
Merge as verification build will not do anything helpful for this one.

PR #706, opened 2022-11-29
Use the tycho-apitools-plugin to generate the api description file

laeubi:
The `tycho-apitools-plugin` provides a mojo now to generate the description file, this should be faster and easier to use/understand than the ANT calls.

Lets see if this causes any issues...

### DIFF
--- a/eclipse-platform-parent/pom.xml
+++ b/eclipse-platform-parent/pom.xml
@@ -745,87 +745,23 @@
       </activation>
       <build>
         <plugins>
-          <plugin>
-            <groupId>org.apache.maven.plugins</groupId>
-            <artifactId>maven-antrun-plugin</artifactId>
-            <executions>
-              <execution>
-                <id>check-has-apiTools-nature</id>
-                <phase>validate</phase>
-                <goals>
-                  <goal>run</goal>
-                </goals>
-                <configuration>
-                  <exportAntProperties>true</exportAntProperties>
-                  <target>
-                    <condition property="skipAPIDescription" value="false" else="true">
-                      <!-- If the property is already set this has no effect because properties are immutable in ANT -->
-                      <and>
-                        <available file="${basedir}/META-INF/MANIFEST.MF"/>
-                        <resourcecontains resource="${basedir}/.project" substring="org.eclipse.pde.api.tools.apiAnalysisNature" />
-                      </and>
-                    </condition>
-                  </target>
-                </configuration>
-              </execution>
-            </executions>
-          </plugin>
-          <plugin>
-            <groupId>org.eclipse.tycho.extras</groupId>
-            <artifactId>tycho-eclipserun-plugin</artifactId>
-            <executions>
-              <execution>
-                <goals>
-                  <goal>eclipse-run</goal>
-                </goals>
-                <phase>process-classes</phase>
-                <id>build-.api_description</id>
-                <configuration>
-                  <skip>${skipAPIDescription}</skip>
-                  <work>${project.build.directory}/apigeneration-workspace</work>
-                  <applicationsArgs>
-	                  <args>-application</args>
-	                  <args>org.eclipse.pde.api.tools.apiGeneration</args>
-	                  <args>-projectName</args>
-	                  <args>${project.artifactId}_${qualifiedVersion}</args>
-	                  <args>-project</args>
-	                  <args>${project.basedir}</args>
-	                  <args>-binary</args>
-	                  <args>${tychoProject.build.outputDirectories}</args>
-	                  <args>-target</args>
-	                  <args>${project.build.directory}</args>
-                  </applicationsArgs>
-                  <dependencies>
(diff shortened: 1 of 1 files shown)


laeubi:
This requires
- https://github.com/eclipse-tycho/tycho/pull/1759

eclipse-platform-parent/pom.xml
```diff
@@ -745,87 +745,23 @@
...
+	            <version>${tycho.version}</version>
+				<executions>
+					<execution>
+						<id>generate</id>
+						<goals>
+							<goal>generate</goal>
+						</goals>
+						<configuration>
+                  			<skip>${skipAPIDescription}</skip>
+                  			<projectName>${project.artifactId}_${qualifiedVersion}</projectName>
```
HannesWell, line 760: Shouldn't it be possible to derive that value in the mojo from the project's metadata?

HannesWell (approved):
Yes fully agree, that using the mojo is simpler to understand/configure and if it is faster, even better!

akurtakov:
Jenkins build failed with :
```
[ERROR] Failed to execute goal org.eclipse.tycho:tycho-apitools-plugin:4.0.0-SNAPSHOT:generate (generate) on project org.eclipse.osgi: Execution generate of goal org.eclipse.tycho:tycho-apitools-plugin:4.0.0-SNAPSHOT:generate failed: A required class was missing while executing org.eclipse.tycho:tycho-apitools-plugin:4.0.0-SNAPSHOT:generate: org/objectweb/asm/ClassVisitor
```
so this seems to need more work.

eclipse-platform-parent/pom.xml
laeubi: Actually it is derived, but as always platform uses a "special" value here (that diverges from the maven version) so to not cause any "confusion" ...

laeubi:
> Jenkins build failed with :
> 
> ```
> [ERROR] Failed to execute goal org.eclipse.tycho:tycho-apitools-plugin:4.0.0-SNAPSHOT:generate (generate) on project org.eclipse.osgi: Execution generate of goal org.eclipse.tycho:tycho-apitools-plugin:4.0.0-SNAPSHOT:generate failed: A required class was missing while executing org.eclipse.tycho:tycho-apitools-plugin:4.0.0-SNAPSHOT:generate: org/objectweb/asm/ClassVisitor
> ```
> 
> so this seems to need more work.

Yes I already submitted a fix to tycho so lets see if that helps, just restarted the build...

akurtakov:
They are https://repo.eclipse.org/content/repositories/eclipse/org/eclipse/eclipse-platform-parent/4.27.0-SNAPSHOT/eclipse-platform-parent-4.27.0-20221201.162228-49.pom .

laeubi:
It seems the build is running now and I see executions of the new api-mojo, so if the build succeeds and now one has concerns this is ready to be merged.

HannesWell:
> It seems the build is running now and I see executions of the new api-mojo, so if the build succeeds and now one has concerns this is ready to be merged.

Do you know if there is a p2-repo produced in the verification build? If yes I think it would be good to replay a run, where the repo is archived and then to check some jars if the contained api-description is fine (compared to the current one from the latest I-build).
If that is not feasible easily at least the next I-build repo should be checked for proper api-descriptions.

But besides that I have no concerns to proceed. :)

laeubi:
> Do you know if there is a p2-repo produced in the verification build? If yes I think it would be good to replay a run, where the repo is archived and then to check some jars if the contained api-description is fine (compared to the current one from the latest I-build).

Don't think so, but the log should "complain" if there are items replaced / different to the baseline, that was why I discovered that I need a different configuration from the defaults because aggregator uses a different version string.

By the way this seem to speed up the build from ~1hr to ~45 minutes ... @akurtakov I'll leave it up to your decision to merge this, the build is green so from my side ready to go and "just try it".

HannesWell (approved):
> > Do you know if there is a p2-repo produced in the verification build? If yes I think it would be good to replay a run, where the repo is archived and then to check some jars if the contained api-description is fine (compared to the current one from the latest I-build).
> 
> Don't think so, but the log should "complain" if there are items replaced / different to the baseline, that was why I discovered that I need a different configuration from the defaults because aggregator uses a different version string.

Indeed the p2-baseline comparator should notice any difference. Im fine with this Change👍🏽

mickaelistria:
Build is currentlyf failing for other reasons. Please wait for a good I-Build before merging it.

mickaelistria:
Last build is successful, let's do it now.

vogella:
If we do not use the api generation Ant tasks anymore in platform, we should IMHO also deprecate them in PDE or even mark them for deletion (in > 2 years we would be able to delete them, haha).

laeubi:
Feel free to start an initiative :-)
But last time I was told that ant is still used alot I think @iloveeclipse mentioned this

PR #711, opened 2022-11-30
Disabling milestone schedule and removing 4.26 job creation

sravanlakkimsetti:


akurtakov:
I think we should keep the possibility to do 4.26 builds (disabled by default!) for the case of respin till release date.

sravanlakkimsetti:
> I think we should keep the possibility to do 4.26 builds (disabled by default!) for the case of respin till release date.

The jobs are created using https://ci.eclipse.org/releng/job/Create%20Jobs/configure. Since the action for removed jobs is set to "ignore" the jobs should not be removed. 

If we have 4.26 listed here, whenever there is a new commit in these files a new job gets created automatically with the enabled schedule.

akurtakov:
Let's do it then.

Issue #712, opened 2022-11-30
 POM and product version changes for 4.27 release


akurtakov:

akurtakov:
This one is done

PR #714, opened 2022-11-30
Bump versions for 4.27 stream

akurtakov:
Tracked in
https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/issues/712


akurtakov:
Merge as this would not succeed until all pom versions are bumped.

PR #746, opened 2022-12-08
Fix checkComposites script

akurtakov:
Only 4.27 and 4.27-I-builds exist and are in development so nothing else should be checked.


akurtakov:
Verification build won't execute this script thus merged directly.

PR #748, opened 2022-12-08
[SDK-TP] Remove javax.servlet.jsp from Eclipse-Orbit

HannesWell:
A more recent javax.servlet.jsp from Maven-Central was added in https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/pull/407 so the old one from Eclipse-Orbit is probably obsolete.

But the build will tell if not.


HannesWell:
Created https://github.com/eclipse-platform/eclipse.platform.common/pull/101 as an attempt to migrate to the new javax.servlet.jsp-api bundle.

.gitmodules
```diff
@@ -22,7 +22,7 @@
   url = https://github.com/eclipse-platform/eclipse.platform.git
 [submodule "eclipse.platform.common"]
   path = eclipse.platform.common
-  url = https://github.com/eclipse-platform/eclipse.platform.common.git
+  url = https://github.com/HannesWell/eclipse.platform.common.git
```
HannesWell, line 25: This has to be removed before merging!

It is just to have some verification build for the changed in the doc bundle.

HannesWell: Reverted it.

Issue #740, opened 2022-12-05
SDK build broken since I20221203-0210


iloveeclipse:
SDK build broken since I20221203-0210, see 
- first fail: https://ci.eclipse.org/releng/job/Builds/job/I-build-4.27/11/console
- last fail: https://ci.eclipse.org/releng/job/Builds/job/I-build-4.27/15/console

The error is happening here:
```
output: /home/jenkins/agent/workspace/Builds/I-build-4.27/eclipse.platform.releng.aggregator/eclipse.platform.releng.aggregator/cje-production/siteDir/eclipse/downloads/drops4/I20221205-0050/buildlogs/reporeports/reports/versionChecksFeatures.html
08:22:38  all feature groups IUs
08:22:38  Number in reference from raw count:79
08:22:38  Number in reference from set: 79
08:22:38  Number in current from raw count:79
08:22:38  Number in current from set: 79
08:22:38  Number in reference missing from current: 0
08:22:38  Number in common in reference and current: 79
08:22:38  Number in current missing from reference: 0
08:22:38  Number in common in reference and current: 79
08:23:00  Unable to init server: Could not connect: Connection refused
08:23:00  P2analyze: Cannot open display: 
08:23:00  P2analyze:
08:23:00  An error has occurred. See the log file
08:23:00  /home/jenkins/agent/workspace/Builds/I-build-4.27/eclipse.platform.releng.aggregator/eclipse.platform.releng.aggregator/cje-production/tmp/reportApplication/p2analyze/configuration/1670224956496.log.
08:23:00  + [[ 13 -ne 0 ]]
08:23:00  + echo 'Failed in Generate Repo reports stage'
08:23:00  Failed in Generate Repo reports stage
08:23:00  + exit 1
```

@sravanlakkimsetti, @akurtakov : according to @mickaelistria 
```
I recently tried to update that application t newer Eclipse versions; I suspect this cause the issue. 
Can we have access to this /home/jenkins/agent/workspace/Builds/I-build-4.27/eclipse.platform.releng.aggregator/eclipse.platform.releng.aggregator/cje-production/tmp/reportApplication/p2analyze/configuration/1670057020766.log file?
```

iloveeclipse:
@sravanlakkimsetti : great, thanks! 
Can you please share **how** did you access the log files? Is any committer able to do this, or are there extra rights needed? In later case, can we modify the job so it can "list" or "publish" the workspace / build artifacts somewhere?

[...]

PR #750, opened 2022-12-09
Use the new location of the cbi p2repo-analyzers product

merks:


merks:
@akurtakov 

I reworked the p2-analyzers to promote builds to the location documented in the job:

https://ci.eclipse.org/cbi/view/p2RepoRelated/job/p2repo-analyzers/

I've also changed the default implementation to use Equinox for jar signing verification rather than launcher the jarsigner executable.  For the use in the simrel, this reduced the job time from ~30 minutes to ~2.5 minutes:

https://ci.eclipse.org/simrel/job/simrel.reporeports.pipeline/

I'm not sure if the verification job will actually verify that this change doesn't cause any problems.  I don't anticipate any.  

Note too that I also added a very crude test to make sure that the report generator actually runs without failures to prevent problems like this one in the future:

https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/issues/740

akurtakov:
Thanks @merks ! It looks quite promising.
The verification build will not execute the repo analyzer so this would not get tested. I have few more releng changes and plan to launch extra I-build to verify them.

akurtakov:
Platform build went from 5 mins to 30 secs! Really nice speedup!

PR #752, opened 2022-12-09
Properly exclude native libs from source tarball

akurtakov:
Should have been done long ago when repos were merged.


akurtakov:
Verification build doesn't generate source tarball thus merge directly.

PR #754, opened 2022-12-09
Reduce smote tests

akurtakov:
Load on releng JIPP gets pretty high after build and tests get delayed thus smoke tests are no longer run on Java 11.
Groovy files no longer used like centos7 and ubuntu 18 and 20 are removed too.


akurtakov:
Verification build will not test these thus pushing directly.

PR #760, opened 2022-12-12
Test for more detailed error output

akurtakov:


akurtakov:
Merge as this at least gives better stacktrace of the failure.

Issue #755, opened 2022-12-09
Stop running I-build tests on Java 11


akurtakov:
There are the following reasons at least:
* Mac tests are no longer running on Java 11
* Reduce load on releng JIPP to delay test results less
* Unstable tests on Java 11 only https://github.com/eclipse-jdt/eclipse.jdt.core/issues/86
* All EPPs are already shipping with Java 17
* Bundles can stay on lower BREE (like Equinox does) but official tests are not run on these older JVMs
* Test dependencies (RedDeer) require Java 17 but it wasn't caught as I-build is not executing pde smartimport tests. See https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/issues/648

akurtakov:
Let's not mix things. I'm in favor of what you ask but let's have your topic separately as this one is far more pressing need.

akurtakov:
First of all this issue is about not running tests on Java 11 for the reasons given. I really ask to not mix the topics here.
Second, everyone that cares about old releases should step up "right now" (tm) and help with these daily, boring things that currently committers can't keep up with. There hasn't been a clean build in months (at least) and actions have to be taken or even more things will degrade.

akurtakov:
This has been completed.

PR #762, opened 2022-12-13
Stop running tests on Java 11 on Linux

akurtakov:
Tracked in
https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/issues/755


akurtakov:
Merge as verification build will not check these scripts, used by I-build only.

Issue #136, opened 2022-03-22
Update to Lucene 9


akurtakov:
This is major upstream release (see https://lucene.apache.org/core/corenews.html#apache-lucenetm-900-available ) making use of new Java features with significant speed improvements and no split packages (smth that hurt us multiple times in the past).
Change will go through Orbit as upstream is not OSGi artifacts.

iloveeclipse:
Is the fail related:
https://download.eclipse.org/eclipse/downloads/drops4/I20221215-1800/testresults/html/org.eclipse.ua.tests.doc_ep427I-unit-cen64-gtk3-java17_linux.gtk.x86_64_17.html

It started with **I20221215-0320** build.

```
expected:<[]> but was:<[java.io.FileNotFoundException: http://127.0.0.1:42945/help/topic/org.eclipse.ui.intro.universal/samples, java.io.FileNotFoundException: http://127.0.0.1:42945/help/topic/org.eclipse.ui.intro.universal/overview, java.io.FileNotFoundException: http://127.0.0.1:42945/help/topic/org.eclipse.ui.intro.universal/qroot, java.io.FileNotFoundException: http://127.0.0.1:42945/help/topic/org.eclipse.ui.intro.universal/whatsnew]>

java.lang.AssertionError: expected:<[]> but was:<[java.io.FileNotFoundException: http://127.0.0.1:42945/help/topic/org.eclipse.ui.intro.universal/samples, java.io.FileNotFoundException: http://127.0.0.1:42945/help/topic/org.eclipse.ui.intro.universal/overview, java.io.FileNotFoundException: http://127.0.0.1:42945/help/topic/org.eclipse.ui.intro.universal/qroot, java.io.FileNotFoundException: http://127.0.0.1:42945/help/topic/org.eclipse.ui.intro.universal/whatsnew]>
at org.junit.Assert.fail(Assert.java:89)
at org.junit.Assert.failNotEquals(Assert.java:835)
at org.junit.Assert.assertEquals(Assert.java:120)
at org.junit.Assert.assertEquals(Assert.java:146)
at org.eclipse.ua.tests.doc.internal.linkchecker.LinkTest.testAllLinks(LinkTest.java:76)
```

akurtakov:
It could be related although I have looked at it today and can't see how lucene is not relevant to what's actually part of the bundle.

iloveeclipse:
[...]

PR #767, opened 2022-12-15
Add Lucene 9.4.2 to target platform

akurtakov:
To allow building against it.
Tracked in
https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/issues/136


akurtakov:
Nothing but addition to target platform thus merging.

PR #768, opened 2022-12-15
Remove lucene 8.4.1 from target platform

akurtakov:
Tracked in
https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/issues/136


akurtakov:
Merge direclty to run I-build as changes are in multi repos and verification can't succeed.

PR #769, opened 2022-12-15
Update net.bytebuddy and org.mockito to the latest version

merks:


merks:
@akurtakov This is another generated update to the available versions.  Any concerns?

akurtakov:
If license check is fine it should be fine.

merks:
Apparently the license checker doesn't like it.

akurtakov:
@merks as you can see from previous comments there is some nice automation to auto open iplab issues. Use it in the future.

akurtakov:
Lucene is approved already for inclusion in Orbit so this is just iplab database issue. Failing build should no longer happen due to https://github.com/eclipse-platform/eclipse.platform.releng/commit/38a6546e18d6f3adcb37706a6e1d3f5787c54663 .

PR #771, opened 2022-12-15
Update tycho-build to 3.0.1

akurtakov:


akurtakov:
Failed license check is due to lucene being approved in iplab but licensecheck doesn't yet know about it.

PR #772, opened 2022-12-15
Add org.glassfish.web.javax.servlet.jsp to TP

mickaelistria:
to progressively replace Orbit's org.apache.jasper.glassfish


akurtakov:
I think goal should be towards using https://mvnrepository.com/artifact/org.apache.tomcat/tomcat-jasper directly rather than anything else that bundles unknown version of it.

PR #774, opened 2022-12-15
Remove Orbit's org.apache.jasper.glassfish

mickaelistria:
Replaced by newer org.glassfish.web.javax.servlet.jsp from Maven Central

Requires:

- [ ] https://github.com/eclipse-platform/eclipse.platform.common/pull/106
- [ ] https://github.com/eclipse-platform/eclipse.platform.releng/pull/175
- [ ] https://github.com/eclipse-platform/eclipse.platform.ua/pull/76
- [ ] https://github.com/eclipse-equinox/equinox/pull/174

### DIFF
--- a/eclipse.platform.releng.prereqs.sdk/eclipse-sdk-prereqs.target
+++ b/eclipse.platform.releng.prereqs.sdk/eclipse-sdk-prereqs.target
@@ -49,11 +49,6 @@
       <repository location="https://download.eclipse.org/tools/orbit/downloads/drops/R20230302014618/repository/"/>
     </location>
     <location includeAllPlatforms="true" includeMode="slicer" type="InstallableUnit">
-
-      <!-- In process of being replaced by org.glassfish.web.javax.servlet.jsp from Central -->
-      <unit id="org.apache.jasper.glassfish" version="2.2.2.v201501141630"/>
-      <unit id="org.apache.jasper.glassfish.source" version="2.2.2.v201501141630"/>
-
       <unit id="org.w3c.css.sac" version="1.3.1.v200903091627"/>
       <unit id="org.w3c.css.sac.source" version="1.3.1.v200903091627"/>
       <unit id="org.w3c.dom.events" version="3.0.0.draft20060413_v201105210656"/>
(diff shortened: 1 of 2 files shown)


mickaelistria:
IMO, we can merge all 4 referenced PRs, wait for next I-Build and then merge this one.
However, it's not 100% sure it's functional and I'm starting 2 weeks of holidays; so I'm not going to push it right now in case it cause issues someone else has to fix while I'm away; but anyone interested in this work is more than welcome to move it forward in the meantime. (@akurtakov maybe?)

akurtakov:
I think you should go https://mvnrepository.com/artifact/org.apache.tomcat/tomcat-jasper (whatever is the latest version prior to jakarta switch). What you chose just bundles some glassfish version, isn't it?

mickaelistria:
tomcat-jasper has some strange Require-Capbility that (some osgi contracts) that cannot be resolved using the official javax.servlet.jsp-api bundle I believe. This glassfish variant doesn't and seems to just work, and is more or less what used to be used via Orbit, so it's safer to use.
I think we should stick to simplest here to reduce our requirement on Orbit; and then when we're done with all JSP stuff, consider moving to jakarta namespace anyway; as the javax.servlet* are now dead.

mickaelistria:
@akurtakov What do you think is the best way to proceed here?

mickaelistria:
It seems like an alternative is to use some felix bundles that add the OSGi contract capability, but mandate using some particular artifacts: eg org.apache.felix.http.servlet-api provides JavaServlet contract but has `Embed-Dependency: tomcat-servlet-api` which is also not so standard.
So overall, tomcat-jasper seems to be the entry point for multiple extra requirements while org.apache.jasper.glassfish seems to just be plug-n-play.

akurtakov:
What are the extra dependencies coming with tomcat-jasper?

mickaelistria:
tomcat-jasper would require the bundle providing the contract capability JavaEL, JavaServlet and JavaJSP. The providers I found for those would be respectively  org.apache.tomcat.tomcat-el-api, org.apache.tomcat.tomcat-servlet-api and org.apache.tomcat-jsp-api ; but I believe we'd then have to remove javax.el, javax.servlet an javax.servlet.jsp which are actually included already in those artifacts.

mickaelistria:
This attempt is overall a failure: the upstream jasper bundle doesn't provide the JDT-based jsp compiler and I don't think we'll invest further in making this work.
I'm reverting the culprit changes.

mickaelistria:
https://github.com/eclipse-platform/eclipse.platform.ua/pull/106#issuecomment-1471435622 describe the failure caused by this attempt to migrate.

PR #776, opened 2022-12-16
Add ECF httpclientjava feature to the target for P2

laeubi:
See eclipse-equinox/p2#205


vogella:
> Is the new provider selected by default?

Not yet, see https://github.com/eclipse-equinox/p2/issues/205

PR #784, opened 2023-01-04
Update Readme with the fact Java 17 is required for the build

akurtakov:
This has been the case for a while now but readme was not adjusted.


akurtakov:
Pure documentation change thus merge.

Issue #794, opened 2023-01-09
Consider disabling signing of builds that are not based on master/maintenance branches


merks:
The m2e project does that like this:

https://github.com/eclipse-m2e/m2e-core/blob/c5866a2f4ac2c731dc2fe2aeb2355af826aef6f7/Jenkinsfile#L42-L55

That idea could be applied here:

https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/blob/66a703cd6ee9d75d6680fe380d0b2b0de13aca4a/Jenkinsfile#L42-L52

This has the added advantage that the GPG details are not available during such builds and hence can't be compromised by a PR...

merks:
@akurtakov 

We could try to do it conditionally like the m2e example.  What do you think of that?

akurtakov:
Sounds good.

merks:
It looks like this failure is caused by this change:
```
17:25:25  [INFO] --- tycho-gpg-plugin:4.0.0-SNAPSHOT:sign-p2-artifacts (default) @ eclipse-junit-tests ---
17:25:25  gpg: signing failed: Bad passphrase
17:25:25  gpg: signing failed: Bad passphrase
```
I'll have to investigate.  I imagine somewhere the PGP signing isn't guarded the same profile as jar signing.

merks:
The build time is reduced to 17 minutes:

![image](https://user-images.githubusercontent.com/208716/211379584-6562ec99-8085-4f3f-b983-9b6f5d822226.png)

versus a more typical 45 minutes:

![image](https://user-images.githubusercontent.com/208716/211379679-88a7237e-6c3b-4295-a268-3b951abcab4e.png)

PR #791, opened 2023-01-09
Update the target platform's Maven dependencies to the latest available

merks:

PR #795, opened 2023-01-09
Disable signing except for master and maintenance branches

merks:
https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/issues/794

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -42,8 +42,16 @@ pipeline {
 			steps {
 				withCredentials([string(credentialsId: 'gpg-passphrase', variable: 'KEYRING_PASSPHRASE')]) {
 					sh '''
+					if [[ ${BRANCH_NAME} == master ]] || [[ ${BRANCH_NAME} =~ ^R[0-9]_[0-9]+_maintenance ]]; then
+						MVN_ARGS="-Peclipse-sign"
+					else
+						MVN_ARGS=
+						export KEYRING="deadbeef"
+						export KEYRING_PASSPHRASE="none"
+					fi
 					mvn clean verify -e -Dmaven.repo.local=$WORKSPACE/.m2/repository \
-						-Pbree-libs -Peclipse-sign \
+						-Pbree-libs \
+						${MVN_ARGS} \
 						-Dmaven.test.skip=true -DskipTests=true -DaggregatorBuild=true \
 						-DapiBaselineTargetDirectory=${WORKSPACE} \
 						-Dgpg.passphrase="${KEYRING_PASSPHRASE}" \


merks:
@akurtakov 

It's like this failure came back:

```
16:06:05  [ERROR] Cannot resolve project dependencies:
16:06:05  [ERROR]   Software being installed: eclipse-junit-tests 4.27.0.qualifier
16:06:05  [ERROR]   Missing requirement: org.eclipse.test.feature.group 3.8.400.v20230109-1044 requires 'org.eclipse.equinox.p2.iu; org.mockito.mockito-core [4.10.0,4.10.0]' but it could not be found
16:06:05  [ERROR]   Cannot satisfy dependency: eclipse-junit-tests 4.27.0.qualifier depends on: org.eclipse.equinox.p2.iu; org.eclipse.test.feature.group 0.0.0
```
Even though I bumped the version:

https://github.com/eclipse-platform/eclipse.platform.releng/pull/176

And the build for this passed afterwards:

https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/pull/791

Is it possible that a new IBuild "used the qualifier up" before the TP changes were committed and now the version needs to change again?

akurtakov:
> Is it possible that a new IBuild "used the qualifier up" before the TP changes were committed and now the version needs to change again?

Absolutely, you have to merge the target platform change and the touch in the same I-build otherwise it would not work.

merks:
Wow, what a nightmare.  One never knows when an IBuild might kick in between the two.  Oh well, such is life...

merks:
@akurtakov 

The build time is reduced to 17 minutes:

image

versus a more typical 45 minutes:

image

If I screwed something up, which I doubt, it's possible that the master build won't do signing, so we need to keep an eye on that.

The most recent master build failed

https://ci.eclipse.org/platform/job/eclipse.platform.releng.aggregator/job/master/887/

for the same reason that this PR's build failed earlier:

https://ci.eclipse.org/platform/job/eclipse.platform.releng.aggregator/job/PR-795/2/

Do you have any concerns about merging this change?

akurtakov:
Considering that this master build is not used for anything (in comparison to m2e one which pushes to snapshots) I don't see any reason to not merge it now.

Jenkinsfile
```diff
@@ -42,8 +42,16 @@ pipeline {
 			steps {
 				withCredentials([string(credentialsId: 'gpg-passphrase', variable: 'KEYRING_PASSPHRASE')]) {
 					sh '''
+					if [[ ${BRANCH_NAME} == master ]] || [[ ${BRANCH_NAME} =~ ^R[0-9]_[0-9]+_maintenance ]]; then
+						MVN_ARGS="-Peclipse-sign"
+					else
+						MVN_ARGS=
+						export KEYRING="deadbeef"
```
akurtakov, line 49: This looks weird. Why is it needed at all?

merks: Which of these 4 lines specifically are of concern? 

The ${MVN_ARGS} is used later so either the option `-Peclipse-sign` is present or it's not present in the mvn command line.  The KEYRING environment variable contains the **private key** and it's better, I think, that the private key not be in the environment where it is visible in any running Java tests, for example, such that someone might have access to it (save it to a file that's archived) and later try to crack the passphrase to make use of it.  Alternatively we could guard the `'initialize PGP'` stage with the same guard as the `'Deploy eclipse-platform-parent pom and eclipse-sdk target'` stage such that it's not placed in the environment in the first place:
```
			when {
				anyOf {
					branch 'master'
					branch 'R*_maintenance'
				}
			}
```
But that's more code duplication.  Or we could "just not care" and remove line 49.  But as I said, I'm not sure specifically which of these lines look "weird"...

akurtakov: I am speaking for the export KEYRING="deadbeef" part.

akurtakov:
Let's merge and be done with it.

PR #798, opened 2023-01-10
Update target platform's Maven OSGi dependencies to the latest available

merks:


merks:
@akurtakov 

Not surprisingly, this fails like this:
```
09:55:46  [ERROR] Cannot resolve project dependencies:
09:55:46  [ERROR]   Software being installed: org.eclipse.rt.osgistarterkit.product 4.27.0.qualifier
09:55:46  [ERROR]   Missing requirement: org.eclipse.equinox.core.feature.feature.group 1.13.1000.v20230105-2227 requires 'org.eclipse.equinox.p2.iu; org.osgi.util.promise [1.2.0.202109301733,1.2.0.202109301733]' but it could not be found
09:55:46  [ERROR]   Cannot satisfy dependency: org.eclipse.rt.osgistarterkit.product 4.27.0.qualifier depends on: org.eclipse.equinox.p2.iu; org.eclipse.equinox.core.feature.feature.group 0.0.0
09:55:46  [ERROR] 
```
A search shows me that the following all include that bundle:

image

So it seems likely that all will need incrementing or touching.

Looking ahead to all the other potential problems from the new versions suggests also PDE is affected:

image

Also this:

image

So it seems all these things spanning three git repositories, will need increment or touching based on these TP changes is this 4th repository...

Is this a correct assumption and outlook of what will be needed?  Of course I can do that and then verify that this thing builds, but then it's possible, like yesterday, that some I-Build kicks in and then one would need to repeat the steps.  Moreover, if I do those changes on-at-a-time based purely on failing verification builds (rather than anticipation), it's even more likely that this will happen, so better I plan ahead and bump all these things, correct?

This seems like an unwieldy process, but I can think of no better way...

akurtakov:
Features where dependency is imported don't hard code versions so they don't need change. Those that include though are problematic as they hardcode version in feature.xml. Over time we should be switching most/all?/ third party deps to import but there have been some issue which I don't even remember anymore . Maybe @mickaelistria can remind us.

PR #799, opened 2023-01-11
Update target platform's junit dependencies to the latest available

merks:


merks:
@akurtakov 

 see I need to force a qualifier update for the org.eclipse.test feature:

image

JDT's feature is also affected, but I get the sense that this feature is not built by the verification build here. How does that work?

akurtakov:
The feature is build as per https://github.com/eclipse-jdt/eclipse.jdt/blob/7196041c1a50d3f0625a0ad1d6447111ba4c3964/pom.xml#L57

PR #800, opened 2023-01-12
Update pmd and checkstyle maven plugins

akurtakov:


akurtakov:
Verification build will not check these thus merge.

Issue #756, opened 2022-12-11
Move I-build Windows tests to Java 17


akurtakov:
Java 17 is used to run tests for Mac (only this version) and on Linux (together with other versions) . As windows resources are limited and Java 17 is the latest LTS JVM and the fact that all EPPs are shipped with Java 17 it makes sense for Windows tests to be run with Java 17 and not Java 11.

akurtakov:
@sravanlakkimsetti Would you please outline what I have to do to achieve that?

akurtakov:
Related to https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/issues/755

sravanlakkimsetti:
Here are the steps:
1. Raise a request with foundation to install java 17(temurin would suffice) on [windows test machine](https://ci.eclipse.org/releng/computer/qa6xd-win11/)
2. Update java links in the test job https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/blob/master/JenkinsJobs/AutomatedTests/I_unit_win32_java11.groovy
3. Since test results publishing scripts parse the name of the job to list java version, you'll need to rename the test job to java 17 and update publishing scripts. The following scripts will need an update
- https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/blob/affbb3a6d1fecebae0f55a5ae9fa9137f9c37d34/cje-production/scripts/publish.xml#L187
- https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/blob/master/eclipse.platform.releng.tychoeclipsebuilder/eclipse/publishingFiles/staticDropFiles/testConfigs.php

akurtakov:
Helpdesk issue opened https://gitlab.eclipse.org/eclipsefdn/helpdesk/-/issues/2392

akurtakov:
@sravanlakkimsetti Can you help with this one https://ci.eclipse.org/releng/job/AutomatedTests/job/ep427I-unit-win32-java17/1/console ?

iloveeclipse:
> I wonder if the directory where Java 11 was installed also had a space?

Nope: see https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/pull/805/files
`-Djvm="C:\\\\openjdk\\\\jdk-11\\\\bin\\\\java.exe"`

akurtakov:
@fredg02 can we get jdk-17 installed in a directory without space?

[...]

PR #805, opened 2023-01-12
Run windows tests with Java 17

akurtakov:
Fixes https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/issues/756


akurtakov:
Can be tested only via I-build so push directly.

PR #808, opened 2023-01-13
Update the java paths using 8dot3 notation

sravanlakkimsetti:
tracked in https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/issues/756

### DIFF
--- a/JenkinsJobs/AutomatedTests/I_unit_win32_java17.groovy
+++ b/JenkinsJobs/AutomatedTests/I_unit_win32_java17.groovy
@@ -115,11 +115,11 @@ For /F "tokens=1* delims==" %%A IN (buildProperties.properties) DO (
 echo on
 set STREAM
 set EBUILDER_HASH
-set JAVA_HOME="C:\\Program Files\\Eclipse Adoptium\\jdk-17.0.5.8-hotspot\\"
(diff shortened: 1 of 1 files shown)


sravanlakkimsetti:
> Out of curiosity, I wonder if a path like `C:\\PROGRA~1\\ECLIPS~1\\` could "change" if something else that starts with ECLIPS and came alphabetically before `Eclipse Adoptium` was installed?

it is based on time. if something else gets installed it will be ECLIPS~2. by the way you actually get these file names using `dir /x`. I used `dir /x` to get the actual names and then updated the script

akurtakov:
So ant runs with Java 11 but uses java 17 to actually execute the tests?

merks:
@sravanlakkimsetti  I see.  It would also work to use `\\ ` to escape the single space (the same way that `\\\\` is used to escape the single `\` right?

sravanlakkimsetti:
> So ant runs with Java 11 but uses java 17 to actually execute the tests?

yes that is correct

sravanlakkimsetti:
> @sravanlakkimsetti I see. It would also work to use `\\ ` to escape the single space (the same way that `\\\\` is used to escape the single `\` right?

In theory yes that should work but need to test on windows platform

PR #815, opened 2023-01-24
Revert "Revert "Replace jxpath by upstream one from Maven Central""

mickaelistria:
This reverts commit cd659acf75ce18b5ee79dfe0a7bdf1fa0f9a9e82.i Now that jxpath isn't directly included anymore (comes transitively) and Tycho issues are resolved, we can finally re-try this.

### DIFF
--- a/eclipse.platform.releng.prereqs.sdk/eclipse-sdk-prereqs.target
+++ b/eclipse.platform.releng.prereqs.sdk/eclipse-sdk-prereqs.target
@@ -65,10 +65,11 @@
       <unit id="org.w3c.dom.svg" version="1.1.0.v201011041433"/>
       <unit id="org.w3c.dom.svg.source" version="1.1.0.v201011041433"/>
 
-      <!-- part of e4 ui tools. See bug 422102 -->
-      <!-- Has Maven deps on non-OSGi jdom, cannot easily use Maven artifact yet -->
-      <unit id="org.apache.commons.jxpath" version="1.3.0.v200911051830"/>
-      <unit id="org.apache.commons.jxpath.source" version="1.3.0.v200911051830"/>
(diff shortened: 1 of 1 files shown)


iloveeclipse:
@mickaelistria : should we re-trigger SDK build to verify both patches?

merks:
Personally I think that would be good to re-trigger.  Life will be simpler from a releng point of view if we don't need to touch features when their 3rd party dependencies change, but we still need to verify that doesn't cause unforeseen problems.

iloveeclipse:
OK, triggered https://ci.eclipse.org/releng/job/Builds/job/I-build-4.27/88/

mickaelistria:
Thanks.

iloveeclipse:
Build succeeds now: https://download.eclipse.org/eclipse/downloads/drops4/I20230124-0530/

Since there is no way to determine tycho snapshot version: was tycho patched to fix something? If yes, would be good to link to the PR in question.

mickaelistria:
> Since there is no way to determine tycho snapshot version: was tycho patched to fix something? 

At the same time, Tycho faced a regression which caused issues while downloading sources: https://github.com/eclipse-tycho/tycho/issues/2009 , this issue -and others- got fixed in Tycho with https://github.com/eclipse-tycho/tycho/pull/2015 .

merks:
FYI, I also verified that [https://download.eclipse.org/eclipse/updates/4.27-I-builds/I20230124-0530/ causes no problems for SimRel contribution and that jxpath is present in the aggregation...

iloveeclipse:
Regression in PDE: https://github.com/eclipse-pde/eclipse.pde/issues/437. I assume just an expectation need to be updated, but I haven't looked into the test yet.

PR #820, opened 2023-01-25
Update to ECJ from M1

akurtakov:
Prep work for https://github.com/eclipse-platform/eclipse.platform.ui/pull/527 due to https://github.com/eclipse-jdt/eclipse.jdt.core/issues/585


iloveeclipse:
@akurtakov : new I Build?

akurtakov:
Already running.

akurtakov:
https://ci.eclipse.org/releng/job/Builds/job/I-build-4.27/93

iloveeclipse:
Great, thanks. Not that I expect we've really messed with compiler, but would be bad if we did but all the test code already uses Java 17 :-)

akurtakov:
We used to have issue with lambda order causing comparator issues but IIRC this should be done now.

PR #821, opened 2023-01-26
Trying to fix Windows test execution (#190)

iloveeclipse:
- removed quotes around set JAVA_HOME, that broke java startup from Ant
- removed duplicated PATH entries (cleanup only)

See https://github.com/eclipse-platform/eclipse.platform.releng/issues/190

### DIFF
--- a/JenkinsJobs/AutomatedTests/I_unit_win32_java17.groovy
+++ b/JenkinsJobs/AutomatedTests/I_unit_win32_java17.groovy
@@ -115,9 +115,9 @@ For /F "tokens=1* delims==" %%A IN (buildProperties.properties) DO (
 echo on
 set STREAM
 set EBUILDER_HASH
-set JAVA_HOME="C:\\PROGRA~1\\ECLIPS~1\\jdk-17.0.5.8-hotspot\\"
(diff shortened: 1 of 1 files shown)


iloveeclipse:
See https://gitlab.eclipse.org/eclipsefdn/helpdesk/-/issues/2567#note_1069514

@akurtakov : if that is the right file that is called on Windows tests, I would like to merge that & trigger build.

Or would it be enough to trigger Windows job only (not sure which source it takes, from head or from last aggregator build)?
I mean this  one: https://ci.eclipse.org/releng/view/Builds/job/AutomatedTests/job/ep427I-unit-win32-java17/

akurtakov:
Once you push a change to groovy files you have to call https://ci.eclipse.org/releng/job/Create%20Jobs/ and it will sync the jobs with what's in master. After that you can call only the win32 job.

iloveeclipse:
OK, triggered the job you've mentioned, that seem to succeed: https://ci.eclipse.org/releng/job/Create%20Jobs/47/console

Triggered Windows job with `I20230125-1800` parameter: https://ci.eclipse.org/releng/view/Builds/job/AutomatedTests/job/ep427I-unit-win32-java17/28/console

Seem to work now!

iloveeclipse:
And it uses `1.10.13` Ant, so yes, there was Ant update.

PR #835, opened 2023-01-31
Don't trigger Java 20 tests

akurtakov:
Java 20 work is going in a branch and these tests can't succeed until this branch is merged in master.


akurtakov:
Merge as this change will not be tested by verification build.

PR #843, opened 2023-02-06
Update target platform's bytebuddy dependencies to the latest available

merks:


merks:
@akurtakov 

The ECA checks is broken right now:

https://gitlab.eclipse.org/eclipsefdn/helpdesk/-/issues/2635

I expect /org.eclipse.test-feature/forceQualifierUpdate.txt needs a bump, unless the include is removed, but the latest time you tried replacing an include with an import, you reverted that change.  But elsewhere I saw that jxpath's include was removed.  So I'm not sure the strategy moving forward.  Please advise.

akurtakov:
I've tried changing all the deps to be imported not included which didn't work and this didn't work out. But for bytebuddy specifically it's transitive dep of mockito and assertj so not listing bytebuddy it in the feature should work. 
Lesson learned from my side - don't try to bite more than can chew at once :)

PR #849, opened 2023-02-09
Migrate the javadoc config from equinox > platform-parent

laeubi:
This migrates some javadoc config that is generic enough from equinox to the platform parent.


akurtakov:
Looks good to me.

laeubi:
@akurtakov do we need to wait for the next I-Build or is master build enough to get this change propagated to the other repositories?

akurtakov:
No need to wait. Master build should publish the parent pom so it should be usable.

PR #850, opened 2023-02-09
Fix for "Releng JIPP not connecting to svc cluster"

iloveeclipse:
Increased memory requested by Linux container to avoid container startup issues / aborted jobs on Linux.

See https://gitlab.eclipse.org/eclipsefdn/helpdesk/-/issues/2478


SDawley:
I ran Create Jobs so the builds have been updated with this change.

Issue #853, opened 2023-02-11
Unanticipated comparator issues in I20230210-1800


akurtakov:
As it can be seen at https://download.eclipse.org/eclipse/downloads/drops4/I20230210-1800/buildlogs/comparatorlogs/buildtimeComparatorUnanticipated.log.txt :
```
Comparator differences from current build
	/home/jenkins/agent/workspace/Builds/I-build-4.27/eclipse.platform.releng.aggregator/eclipse.platform.releng.aggregator/cje-production/siteDir/eclipse/downloads/drops4/I20230210-1800
compared to reference repo at 
	https://download.eclipse.org/eclipse/updates/4.27-I-builds

1.  equinox/bundles/org.eclipse.osgi/pom.xml
   no-classifier: different
      META-INF/MANIFEST.MF: different
         Require-Capability: baseline='osgi.ee; filter:="(| (&(osgi.ee=JavaSE)(version=1.8)) (&(osgi.ee=JavaSE/compact1)(version=1.8)) )"' != reactor='osgi.ee;filter:="(| (&(osgi.ee=JavaSE)(version=1.8)) (&(osgi.ee=JavaSE/compact1)(version=1.8)) )",osgi.ee;filter:="(&(osgi.ee=JavaSE)(version=1.8))"'
    The main artifact has been replaced with the baseline version.
    The following attached artifacts have been replaced with the baseline version: [sources]

2.  eclipse.platform/runtime/bundles/org.eclipse.e4.core.services/.polyglot.META-INF
   no-classifier: different
      META-INF/MANIFEST.MF: different
         Bundle-RequiredExecutionEnvironment: present in baseline version only
         Require-Capability: baseline='osgi.extender; filter:="(&(osgi.extender=osgi.component)(version>=1.2)(!(version>=2.0)))"' != reactor='osgi.extender;filter:="(&(osgi.extender=osgi.component)(version>=1.2)(!(version>=2.0)))",osgi.extender;filter:="(&(osgi.extender=osgi.component)(version>=1.5.0)(!(version>=2.0.0)))",osgi.ee;filter:="(&(osgi.ee=JavaSE)(version=11))"'
    The main artifact has been replaced with the baseline version.
    The following attached artifacts have been replaced with the baseline version: [sources]

3.  eclipse.platform/runtime/bundles/org.eclipse.e4.core.di.extensions.supplier/.polyglot.META-INF
   no-classifier: different
[...]

PR #854, opened 2023-02-11
Stop generating OSGi headers from source

akurtakov:
Temporary disable as we are late in 2022-03 stream and a some things are still work in progress on Tycho and PDE side.
See https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/issues/853 for details.


akurtakov:
Don't wait for verification build as it will not help us with comparator issues. I'm starting I-build with this commit.

PR #855, opened 2023-02-13
Update byte-buddy to 1.13.0

merks:


merks:
@akurtakov 

I'm not sure you want such updates during this week.   Probably the test feature needs a bump, but no point in doing that if we don't want TP updates this week...

akurtakov:
Please do them now. As next is RC1 and after that rc2 so the earlier the better or it has to be delayed for next stream which I don't think would be beneficial.

PR #857, opened 2023-02-13
Update commons-fileupload to 1.5

merks:


merks:
@akurtakov 

I assume we want this one too.  I expect /org.eclipse.sdk.tests (feature)/forceQualifierUpdate.txt will need a bump.

akurtakov:
Yes, please. Once we go in RC phase we will have to slow down.

merks:
@akurtakov 

I'm a bit concerned that the build passed but I expected this include to be a problem:

image

Should I bump that one's qualifier just to be safe?

I see https://gitlab.eclipse.org/eclipsefdn/emo-team/iplab/-/issues/7109 is taking a bit longer than usual...

akurtakov:
If the build succeeds let's trust it and don't touch needlessly. If smth else in this feature has been touched it would happen to be a "touch"

merks:
@akurtakov Okay.  Should I wait for the IP review to finish?  It seems some manual process has kicked in.  Though I believe we are allowed to do the review in parallel.  It's your call...

akurtakov:
Yes, we have to wait for the IP process as we can not release until approved and I'm not looking for a revert in RC phases if this happens to not be approved for some reason.

akurtakov:
Overall, in Platform we play it safe and include things only after approved as the convoluted build is not making revert fun.

merks:
@akurtakov 

Thanks so much, as always, for your ultra-speedy responses!  😃

PR #862, opened 2023-02-15
Bug 570471 - Add utilities for perspectives persistence and migration

eneufeld:
Add new feature to be published.
Also add feature to oomph configuration.

### DIFF
--- a/eclipse.platform.releng.tychoeclipsebuilder/eclipse.platform.repository/category.xml
+++ b/eclipse.platform.releng.tychoeclipsebuilder/eclipse.platform.repository/category.xml
@@ -13,6 +13,7 @@
    <feature id="org.eclipse.equinox.executable" version="0.0.0"/>
    <feature id="org.eclipse.sdk" version="0.0.0"/>
    <feature id="org.eclipse.e4.core.tools.feature.source" version="0.0.0"/>
+   <feature id="org.eclipse.e4.tools.persistence.feature" version="0.0.0"/>
    <feature id="org.eclipse.pde.unittest.junit" version="0.0.0"/>
    <feature id="org.eclipse.pde.unittest.junit.source" version="0.0.0"/>
    <feature id="org.eclipse.tips.feature" version="0.0.0"/>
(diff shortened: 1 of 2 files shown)


eneufeld:
The new feature was added here: https://github.com/eclipse-platform/eclipse.platform.ui.tools/pull/21

eneufeld:
@iloveeclipse This is to late right?

iloveeclipse:
> @iloveeclipse This is to late right?

No, not yet. However I have no stakes in the p2 / deploying area, I can't really say if the change is OK or not.

@merks or @akurtakov: could you please check this PR?

merks:
This looks very good to me.  It's nice to see the setup changes included!  I think it's better to do this sooner rather than later, so I would recommend merging it now.

eclipse.platform.releng.tychoeclipsebuilder/eclipse.platform.repository/sdk.product
```diff
@@ -183,6 +183,8 @@ United States, other countries, or both.
       <feature id="org.eclipse.equinox.p2.user.ui.source"/>
       <feature id="org.eclipse.e4.core.tools.feature" installMode="root"/>
       <feature id="org.eclipse.e4.core.tools.feature.source" installMode="root"/>
+      <feature id="org.eclipse.e4.tools.persistence.feature" installMode="root"/>
```
akurtakov, line 186: I would be fine with having it as separately installable ondemand feature in the p2 repository but not having it in the SDK tarball as done here. If it has to be in the SDK by default it would better have better explanation why this is done.

akurtakov:
I propose to have it in the p2 repo (category.xml part of this PR) but not do it in the tarball (SDK.product part of this PR). That way people that want it will be able to use it but we don't have enough time to see the feature is in good enough state for install on everyone's machine. That part should happen early next cycle IMHO.

eneufeld:
Thank you @akurtakov for the review,
I removed it from the SDK.

akurtakov:
If/when verification succeeds here is my +1 to merge it in so it ends up in M3.

PR #872, opened 2023-02-17
Add a profile to bump versions

laeubi:
If running a build, one can specify `-Pvb` to automatically bump the version(s) of plugins that have changed.

### DIFF
--- a/eclipse-platform-parent/pom.xml
+++ b/eclipse-platform-parent/pom.xml
@@ -619,6 +619,33 @@
      </repositories>
    </profile>
    <profile>
+	  <!-- This provile enables automatic version bumps when running the build -->
+      <id>vb</id>
+      <properties>
+		  <compare-version-with-baselines.skip>false</compare-version-with-baselines.skip>
+      </properties>
+       <build>
+        <plugins>
+          <plugin>
+            <groupId>org.eclipse.tycho</groupId>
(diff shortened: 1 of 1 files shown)


akurtakov:
Let's get it in. I'll definetely try it out when opening next stream.

laeubi:
I just thought it might be more for users that submit PRs, but if its usefull for the release as well that's great. Can you explain why opening next stream requires version bumps maybe an example where one can see this? It can then maybe even fully automated then using a github action...

akurtakov:
First thing that comes to my mind is docs plugins having version in generated content, "branding" plugins like https://github.com/eclipse-platform/eclipse.platform.releng/blob/master/bundles/org.eclipse.rcp/ having same version as the stream and all of these causing features including them to transitively require version bump. At one point @vogella tried bumping pde feature for that reason to have the same but this "major" version bump caused a number of things to break in simrel so it ended up reverted. Although it's probably good idea to revise this strategy.

laeubi:
I think we should observe this the next time and check for these obstacles if we can avoid or at least automate them.

If you are aware of some older PRs I can look at I might be able to simulate this on a fork to see if we can make this more smooth, e.g. with the move of `tools->ui` I also noticed that a lot of version bumps where required (I think it was around 60) while I would have expected only a few.

Issue #870, opened 2023-02-16
The ancient org.apache.commons.beanutils 1.8.0.v201205091237 bundles are seen as unsigned


merks:
This test shows the failures:

https://ci.eclipse.org/oomph/job/repository-analyzer/lastCompletedBuild/testReport/

It's reported here:

https://download.eclipse.org/oomph/archive/reports/download.eclipse.org/eclipse/updates/4.27-I-builds/index.html

Specifically this:

![image](https://user-images.githubusercontent.com/208716/219323412-5b2ac131-473d-463c-a992-47f7350fa76f.png)

This is a regression caused by this change:

https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/pull/773

I'm not sure how removing one Orbit dependency but adding two other Orbit dependencies was an improvement.

@akurtakov

We should either revert that, force PGP sign these things here:

https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/blob/7538dfb1cd1b1e934e8e6343fdef76d478b6757e/eclipse.platform.releng.tychoeclipsebuilder/pom.xml#L45-L66

Or use this newer version if possible:

https://download.eclipse.org/staging/2023-03/buildInfo/archive/download.eclipse.org/staging/2023-03/index/org.apache.commons.commons-beanutils_1.9.4.html

I'm mostly unavailable today and tomorrow...

akurtakov:
I just see this one but can't find beanutils 1.9.4 in latest Orbit https://download.eclipse.org/tools/orbit/downloads/drops/S20230214193619/repository/plugins/ . I think we should go for 1.9.4 from Maven central as 1.8.0 has https://cve.mitre.org/cgi-bin/cvename.cgi?name=CVE-2019-10086 and https://cve.mitre.org/cgi-bin/cvename.cgi?name=CVE-2014-0114 . The choice between Orbit and Maven Central looks easy if beanutils 1.9.4 is not in Orbit.

akurtakov:
@merks Now I'm really puzzled. Open https://download.eclipse.org/eclipse/downloads/drops4/I20230215-1800/buildlogs/reporeports/reports/verified8.txt  look for beanutils and see:
```
org.apache.commons.beanutils.source_1.8.0.v201205091237.jar                                  plugin       jar verified (jarsigner)
org.apache.commons.beanutils_1.8.0.v201205091237.jar                                         plugin       jar verified (jarsigner)
```

[...]

PR #873, opened 2023-02-19
Run the p2analyze report using the latest JustJ Java 17 JRE

merks:
This is a Temurin-derived JRE and is same JRE that is used with all the SimRel products.  Using this URL avoids rate limits that often affect direct Temurin JREs.

https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/issues/870


merks:
@akurtakov 

I can't fully test these changes and I doubt even the verification build tests these changes, though I'm not sure.  I did test as much as I could locally on Windows...

akurtakov:
Verification build would not test it. I'm merging so next I-build in less than 3h tests it.

PR #874, opened 2023-02-19
Update byte-buddy to 1.14.0 and jsoup to 1.15.4

merks:


merks:
I expect a qualifier bump will be needed for the test feature and that a license review is needed.

PR #875, opened 2023-02-19
Deploy the poms also on branches that are preparing a release

laeubi:
This adds a new condition where poms are deployed when the branch start with the prefix 'prepare_R' and skip the build in such case.

@akurtakov @sravanlakkimsetti what do you think? The idea is that with some automation one can create the `prepare_R...` automatically (I'll prepare a PR for this as well), then a build is triggered and the pom is deployed so it could be used by other steps soon. As other repos needs to be adjusted to use the new master, a build is skipped in this case as this can't work anyways...

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -31,6 +31,7 @@ pipeline {
 				anyOf {
 					branch 'master'
 					branch 'R*_maintenance'
+					branch 'prepare_R*'
 				}
 			}
 			steps {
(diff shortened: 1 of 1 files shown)


laeubi:
@akurtakov I'm currently looking at the last increment here f97bfa71c08e3d43ebfd4243a073a6f07d095d86 how do you prepare such commit? Is there a script (maybe `scripts/updateProductVersion.sh`?) / mvn command? Search + Replace?

akurtakov:
https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/blob/master/scripts/updateProductVersion.sh  is the one. It doesn't handle the simrel version.

PR #876, opened 2023-02-20
Add a workflow to prepare the next release cycle

laeubi:
This workflow is triggered by the milestone creation and automatically performs the updates of the master pom and target to be prepared for usage in the submodules and creates a PR. This PR will then trigger the deploy of the changed xmls, so it can be used immediately to adjust the sub-projects.

FYI @akurtakov @sravanlakkimsetti this is the workflow mentioned here:
- https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/pull/875

It then creates a PR automatically once the milestone is created and the description of the milestone includes "Release" in the description like this:
- https://github.com/laeubi/eclipse.platform.releng.aggregator/pull/7

If you agree, I'd like to merge this (together with the Jenkins file) and create a new milestone 4.28 in this repository (we can delete it afterwards before the final milestone creation script runs) so I can prototype and verify an automated update for e.g. the eclipse.platform sub repo.

There are two things to improve:
- `releaseName` is not updated automatically, but it is only used on two places in platform repo, I'm working on that part and probably push an update here soon
- `previous-release.baseline` can not be automatically derived as it includes a timestamp, my idea would be that there will be a link that is more script friendly, e.g `https://download.eclipse.org/eclipse/updates/4.26/baseline/` (instead of currently used https://download.eclipse.org/eclipse/updates/4.26/R-4.26-202211231800/) would this be possible?

### DIFF
--- a/.github/workflows/prepareRelease.yml
+++ b/.github/workflows/prepareRelease.yml
@@ -1,0 +1,43 @@
+name: Prepare Next Release
+on:
+  milestone:
+    types: [created]
+
+jobs:
+  prepare:
+    runs-on: ubuntu-latest
+    if: contains(github.event.milestone.description, 'Release')
+    permissions:
+      pull-requests: write
+      contents: write
+    steps:
+    - id: get-release-name
+      run: |
+        name=$(echo ${{ github.event.milestone.due_on }} | cut -d- -f-2)
+        echo "name=$name" >> $GITHUB_OUTPUT
+    - uses: actions/checkout@v3
+      with:
+        ref: master
+    - name: Set up JDK
+      uses: actions/setup-java@v3
+      with:
+        java-version: '17'
+        distribution: 'temurin'
+        cache: maven
+    - name: Update Main Versions
+      run: mvn -U -ntp -f eclipse-platform-parent org.eclipse.tycho:tycho-versions-plugin:4.0.0-SNAPSHOT:set-version -DnewVersion=${{ github.event.milestone.title }}.0-SNAPSHOT -Dmodules=../eclipse.platform.releng.prereqs.sdk
+    - name: Update Release Versions
+      run: mvn -ntp -f eclipse-platform-parent/pom.xml --non-recursive org.eclipse.tycho:tycho-versions-plugin:4.0.0-SNAPSHOT:set-property -Dproperties=releaseNumberSDK,releaseNumberPlatform,releaseName -DnewReleaseName=${{ steps.get-release-name.outputs.name }} -DnewReleaseNumberSDK=${{ github.event.milestone.title }} -DnewReleaseNumberPlatform=${{ github.event.milestone.title }}
+    - name: Create Pull Request for Release ${{ github.event.milestone.title }}
+      uses: peter-evans/create-pull-request@v4
+      with:
+        commit-message: Prepare Release ${{ github.event.milestone.title }}
+        branch: prepare_R${{ github.event.milestone.title }}
+        title: Prepare Release ${{ github.event.milestone.title }}
+        body: A new Release Milstone was created, please review the changes and merge if appropriate.
+        delete-branch: true
+        milestone: ${{ github.event.milestone.number }}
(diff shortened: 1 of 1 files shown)


akurtakov:
Please go on and merge. It looks very promising. 
Regarding your question about baseline - there is already https://download.eclipse.org/eclipse/updates/latest/ pointing to latest released version so IMHO we should be fine to just hardcode to always compare to latest stable release.

laeubi:
I think latest is bad because if you want to build an older release it points to something bogus, but it seems one could use https://download.eclipse.org/eclipse/updates/4.26/ already? Not sure why the specific one is used.

So would you be fine with first merging this one:
- https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/pull/875

I'm currently working on the release name and then this one can be merged, and I start the deploy test, after that we can decide on the update URL :-)

akurtakov:
@SDawley please try using this work for opening next stream.

laeubi:
The release name is now derived from the due date of the milestone...

laeubi:
@akurtakov @SDawley @sravanlakkimsetti I'd like to create a test milestone in this repository to validate everything works like expected, any concerns?

akurtakov:
Assuming everything can be cleaned after that - fine with me.

sravanlakkimsetti:
Fine with me too. yes it can be cleaned up

laeubi:
@sravanlakkimsetti @akurtakov what usually fills the 

`https://download.eclipse.org/eclipse/updates/<release>/`

repo? Is this done with the first RC1 build? Can we assume that this happens before the new milestones are created?

sravanlakkimsetti:
> @sravanlakkimsetti @akurtakov what usually fills the
> 
> `https://download.eclipse.org/eclipse/updates/<release>/`
> 
> repo? Is this done with the first RC1 build? Can we assume that this happens before the new milestones are created?

This is created 5 days before actual release. Specifically after Simrel RC2 is done. Milestones are created before this. We normally open next release before this repo is created. 
The build promotion job does the work for us.

In 4.27
4.28 stream to be opened March 6
https://download.eclipse.org/eclipse/updates/4.27/ will be created on March 10

laeubi:
I think then there is a gap in the process, because once one opens the stream it will always compare against an old baseline, so one can't effectively work on the new stream. I think it must be something like:

1. `https://download.eclipse.org/eclipse/updates/<release>/` created (maybe with some temporary content for now)
2. Now the milestones are created
3. Then the PRs are created out of this

So what is the intend in "opening" the stream before the baseline is available at all? In the worst, this means version bumps are lost and only discovered a few days later.

sravanlakkimsetti:
There is a 2 week overlap. in the build scripts we use RC2 build as base line. This baseline is updated once the final release is done.

laeubi:
I'm confused :-)

So the build scripts are modified all the time? Why not make RC1 / RC2 available already at `https://download.eclipse.org/eclipse/updates/<release>/` then so one can keep one URL over the whole phases? As it is already a composite repo it should be quite easy to have RC1 + R2 + FINAL folders and one can simply drop the older ones once a "higher" becomes aviable?

sravanlakkimsetti:
> I'm confused :-)
> 
> So the build scripts are modified all the time? Why not make RC1 / RC2 available already at `https://download.eclipse.org/eclipse/updates/<release>/` then so one can keep one URL over the whole phases? As it is already a composite repo it should be quite easy to have RC1 + R2 + FINAL folders and one can simply drop the older ones once a "higher" becomes aviable?

You are right when you say build scripts are modified all the time. 

Why we are doing this:
Till 4.8 we used to have 3 service releases and all these service releases are added to the release composite. In this case we cannot add RC builds of service releases to the composite. So we never added RC to the composite. 

Your Idea is good. Probably we should go with composite containing RC phases and final one

laeubi:
@sravanlakkimsetti I created:
- https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/issues/884

PR #880, opened 2023-02-20
Removes eclipse.platform.ui.tools from aggregator build

vogella:
eclipse.platform.ui.tools was merged into eclipse.platform.ui so we can remove it from the aggregator build.


vogella:
@akurtakov I plan to merge this one, after the Verification build.

Issue #882, opened 2023-02-20
Retire source tarball generation


vogella:
From https://github.com/eclipse-platform/eclipse.platform.ui.tools/issues/31#issuecomment-1437122784

````
After the build repos are not clean so either it would have to be cleaned or even worse repos would have to be cloned again. This raise the question who needs the source tarball anymore? It was added by Red Hat and used for the Fedora/RHEL rpm builds which we no longer use so maybe no one uses them at all.

````

This would save around 10 min in the aggregator build.

https://user-images.githubusercontent.com/139910/220134849-79eed542-de29-4b15-b569-6eb9952e606b.png

cc @laeubi @akurtakov

akurtakov:
I've sent a link to this pr to eclipse-dev so if someone still uses it has a chance to jump in.

sravanlakkimsetti:
I could not create a build using the tar ball 
+100 from my side

iloveeclipse:
We also never used that. Let remove it and simplify build.

HannesWell:
Never used that as well.

akurtakov:
Haven't checked it but if @sravanlakkimsetti couldn't build it out of the box I doubt that anyone can do it without major work. I'll be dropping this one for RC1.

sravanlakkimsetti:
This one caused tests to fail. Tests are dependent one small tarball containing the scripts that is not getting built now.

sravanlakkimsetti:
Will fix shortly and restart build

akurtakov:
I'm eager to see how this relates to tests.

sravanlakkimsetti:
There are two tar balls we are creating in the build source bundles stage. 
1. [eclipse-platform-sources-I20230220-1800.tar.xz](https://download.eclipse.org/eclipse/downloads/drops4/I20230220-1800/download.php?dropFile=eclipse-platform-sources-I20230220-1800.tar.xz)
2. [eclipse.platform.releng.aggregator-hash.zip](https://download.eclipse.org/eclipse/downloads/drops4/I20230220-1800/eclipse.platform.releng.aggregator-f192e41ef63b1b5f4f43d9d55a7b645ef797f052.zip)

te second one contains the test scripts required to test on different configurations. its a zip of https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/tree/master/production/testScripts. These scripts are used for run ant based tests.

akurtakov:
[...]

PR #886, opened 2023-02-21
Stop generating source tarballs

akurtakov:
No one seems to be using them and they didn't work out of the box. Fixes https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/issues/882


akurtakov:
Verification build will not test it thus merge and see I-build in an hour for results.

PR #888, opened 2023-02-21
Retire generation of source tar ball

sravanlakkimsetti:
added generation of eclipse.platform.releng.aggregator-${HASH}.zip file again

fixes https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/issues/882


akurtakov:
Why/how exactly is this zip used?

PR #948, opened 2023-03-09
Include org.eclipse.jdt.ui in P build

jarthana:


jarthana:
Please ignore this, till I remove the two stale commits.

jarthana:
I see this is planned as part of #949 as well. We don't need this.

PR #950, opened 2023-03-09
Remove SWT-Natives build step from I/Y-builds pipelines

HannesWell:
With https://github.com/eclipse-platform/eclipse.platform.swt/pull/514 SWT-natives are build as part of the SWT master-branch build.

@SDawley I'm not sure if it is sufficient to change the scripts here or if they have to be deployed to the Releng Jenkins in order to take effect? Are there more places that have to be adjusted?

### DIFF
--- a/JenkinsJobs/Builds/I_build.groovy
+++ b/JenkinsJobs/Builds/I_build.groovy
@@ -230,11 +230,6 @@ spec:
                 }
             }
         }
-	  stage('Swt build input') {
-	      steps {
-	          build 'SWT-Increment_if_needed'
-	      }
-	    }
 	  stage('Create Base builder'){
           steps {
               container('jnlp') {
--- a/JenkinsJobs/YBuilds/Y_build.groovy
+++ b/JenkinsJobs/YBuilds/Y_build.groovy
@@ -225,11 +225,6 @@ spec:
                 }
             }
         }
-	  stage('Swt build input') {
-	      steps {
-	          build 'SWT-Increment_if_needed'
-	      }
-	    }
 	  stage('Create Base builder'){
           steps {
               container('jnlp') {


HannesWell:
@akurtakov do you want to take a look at this? Or can tell me who is responsible for this in the future?

akurtakov:
If the swt master builds and pushes the natives that should be enough to remove it from aggregator build.  Does it happen already?
About the question of responsibility for aggregator build I personally believe we should be looking for shared ownership by active committers where everyone do various bits and not rely on single person or two being responsible for it. With that said, I'm still available and working on the build but with far more limited capacity to react on issues.

HannesWell:
> If the swt master builds and pushes the natives that should be enough to remove it from aggregator build. Does it happen already?

Yes, https://github.com/eclipse-platform/eclipse.platform.swt/pull/514 is merged and it works well. You can see it from the past commit history:
https://github.com/eclipse-platform/eclipse.platform.swt/commits/master

> About the question of responsibility for aggregator build I personally believe we should be looking for shared ownership by active committers where everyone do various bits and not rely on single person or two being responsible for it. With that said, I'm still available and working on the build but with far more limited capacity to react on issues.

Agree, that would be good. Good to know, thank you.

So I think we can merge this then?
Btw. can you tell what the purpose of Y-builds is exactly and when they are triggered with what input? Is there some documentation about that.

akurtakov:
> > If the swt master builds and pushes the natives that should be enough to remove it from aggregator build. Does it happen already?
> 
> Yes, [eclipse-platform/eclipse.platform.swt#514](https://github.com/eclipse-platform/eclipse.platform.swt/pull/514) is merged and it works well. You can see it from the past commit history: https://github.com/eclipse-platform/eclipse.platform.swt/commits/master
> 
> > About the question of responsibility for aggregator build I personally believe we should be looking for shared ownership by active committers where everyone do various bits and not rely on single person or two being responsible for it. With that said, I'm still available and working on the build but with far more limited capacity to react on issues.
> 
> Agree, that would be good. Good to know, thank you.
> 
> So I think we can merge this then? 

Go for it.

> Btw. can you tell what the purpose of Y-builds is exactly and when they are triggered with what input? Is there some documentation about that.

Y-builds are used by JDT team while working on next Java version support, right now Java 20 . What happens in these builds is that some submodules e.g. jdt.*, pde and if smth else needs changing are switched to use some branch and not master. Unfortunately, I don't know much more so you'll have to ask JDT project for details.

HannesWell:
> Go for it.
🚀 Should the scripts the re-deployed for the current I/Y-builds or should we wait until that happens for the autumn release cycle?

> > Btw. can you tell what the purpose of Y-builds is exactly and when they are triggered with what input? Is there some documentation about that.
> 
> Y-builds are used by JDT team while working on next Java version support, right now Java 20 . What happens in these builds is that some submodules e.g. jdt.*, pde and if smth else needs changing are switched to use some branch and not master. Unfortunately, I don't know much more so you'll have to ask JDT project for details.


Understand. Thanks for the elaboration. For me that is sufficient to know. I was mainly interested if these builds are based on some older version of SWT and therefore didn't have the change of the mentioned PR. But I now assume that's not the case.

PR #951, opened 2023-03-09
Add o.e.pde.spies feature to eclipse/updates repo

HannesWell:
With https://github.com/eclipse-pde/eclipse.pde/pull/479 the PDE-spies have been moved into a dedicated feature, which has to be added to the eclipse/updates p2-repo.

Can anybody verify that with this change the new feature ends up in the next I-builds repo at https://download.eclipse.org/eclipse/updates/4.28-I-builds ?

I don't think we should add it to the o.e.sdk feature.

### DIFF
--- a/eclipse.platform.releng.tychoeclipsebuilder/eclipse.platform.repository/category.xml
+++ b/eclipse.platform.releng.tychoeclipsebuilder/eclipse.platform.repository/category.xml
@@ -40,6 +40,8 @@
    <feature id="org.eclipse.ecf.filetransfer.httpclient5.feature.source" version="0.0.0"/>
    <feature id="org.eclipse.ecf.filetransfer.ssl.feature" version="0.0.0"/>
    <feature id="org.eclipse.ecf.filetransfer.ssl.feature.source" version="0.0.0"/>
+   <feature id="org.eclipse.pde.spies"/>
+   <feature id="org.eclipse.pde.spies.source"/>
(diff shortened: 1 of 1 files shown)


akurtakov:
According to https://download.eclipse.org/eclipse/downloads/drops4/I20230310-1800/buildlogs/reporeports/reports/versionChecks.html the pde spies are not missing from the repo.

merks:
@akurtakov 

Actually, given they are only mentioned in the section titled "IUs in reference, but not in current repo" the report says the opposite...

This generated report also shows them missing:

https://download.eclipse.org/oomph/archive/reports/download.eclipse.org/eclipse/updates/4.28-I-builds/https___download.eclipse.org_eclipse_updates_4.28-I-builds_I20230310-1800.html

And you don't really need a report to see they are missing, you can simply look in the repository:

https://download.eclipse.org/justj/?file=eclipse/updates/4.28-I-builds/I20230310-1800/plugins

HannesWell:
The archived content.xml/artifacts.jar indicate that the pde.spy.* bundles are included with this change and since the PDE build currently fails with baseline problems that mention some spy bundles I highly suspect it is because they are missing in the repos.
Therefore I think we should merge this ASAP so that the spies are at least in the I-builds repos.

Once we have agreed about how to include the spies eventually in https://github.com/eclipse-pde/eclipse.pde/pull/479 I will do the corresponding changes that will achieve the result we agreed upon, even if this means that this change can be reverted.

PR #970, opened 2023-03-15
Make products require Java 17

akurtakov:
As there are bundels with such EE now in them.


mpalat:
Thanks Aleks for the change. 
However a quick check: Is this announced in any mailing list? Maybe a good idea to put it out in the mailing lists as some of the users are taken by surprise. - If it is already announced, please let know  the link..Thanks

mpalat:
Thanks @iloveeclipse @akurtakov

PR #973, opened 2023-03-16
Use the new tycho-api-tools-verify-mojo

laeubi:


eclipse-platform-parent/pom.xml
```diff
@@ -773,7 +773,17 @@
...
         <plugins>
-			<plugin>
+        </plugins>
+      </build>
+    </profile>
+    <profile>
+       <id>api-check</id>
+       <activation>
+         <activeByDefault>false</activeByDefault>
+       </activation>
```
HannesWell, line 783: Why is this necessary? Isn't the default of `activeByDefault` `false` anyway?

HannesWell (changes requested):
Really looking forward to use the api-analysis goal.

Nevertheless I think we should not merge the execution of the `generate` and `analysis` goal into one profile.
AFAIK the `api-check` profile is only active in verification builds but not in the I-build, but we want the generation in the api-description in the I-build.

By the way, I think we should remember some runtimes of verification builds to compare the effect. :)

laeubi:
> Nevertheless I think we should not merge the execution of the generate and analysis goal into one profile.

I looked around a bit but can't really find where the profile is used ... but will restore it then if its used for IBuilds only...

> By the way, I think we should remember some runtimes of verification builds to compare the effect. :)

I'll trigger a build with the "old" code right now, but I think the API checks are not executed as part of this build?

laeubi:
I have now also tried to activate API generation here to see if we can have some verification of this already before the merge, this should be reverted before a merge.

HannesWell:
> > Nevertheless I think we should not merge the execution of the generate and analysis goal into one profile.
> 
> I looked around a bit but can't really find where the profile is used ... but will restore it then if its used for IBuilds only...

Looking at the profile definition it is activated for all modules that have a `META-INF/MANIFEST.MF` file.
```
    <profile>
      <id>api-generation</id>
      <activation>
        <!-- Does it have to be a profile if we can evaluate condition reliably in antrun? -->
        <file><exists>META-INF/MANIFEST.MF</exists></file>
        <property>
          <name>!longnotexistingproperty</name>
        </property>
      </activation>
```
But since the `apitools:generate` goal already checks if that file exists and skips its execution if not:
https://github.com/eclipse-tycho/tycho/blob/757d125589b36f68bd553db9140803ad8c9bcef1/tycho-apitools-plugin/src/main/java/org/eclipse/tycho/apitools/ApiFileGenerationMojo.java#L85-L86

... the mojo already does the some and I think we can just move that mojo to the regular `build/plugins` section.
I would assume that evaluating the profile's condition also has some runtime costs and there is (hopefully) not much different if a profile or the mojo checks for the existance of a file



> > By the way, I think we should remember some runtimes of verification builds to compare the effect. :)
> 
> I'll trigger a build with the "old" code right now, but I think the API checks are not executed as part of this build?

No api-checks are usually only enabled in builds of the different project repos.

eclipse-platform-parent/pom.xml
```diff
@@ -800,125 +800,30 @@
...
+						<configuration>
+						 	<apiToolsRepository>
+						 		<url>${eclipserun-repo}</url>
+						 	</apiToolsRepository>
+						 	<baselines>
+							 	<repository>
+							 		<url>${previous-release.baseline}</url>
+							 	</repository>
+						 	</baselines>
+						 	<skip>${skipAPIAnalysis}</skip>
```
HannesWell, line 822: On the long run, we should migrate all users of this properties to use the mojo's configuration property `tycho.apitools.verify.skip`.

HannesWell (approved):
Looks good. 👍🏽

laeubi:
But one can still compare some numbers already:

## org.eclipse.equinox.common

### Ant based from start of mojo until API result
2023-03-17T06:01:39.221Z
2023-03-17T06:01:58.962Z

19 seconds

### Tycho mojo from start of mojo until API result
2023-03-17T06:11:04.993Z
2023-03-17T06:11:23.167Z

19 seconds

## org.eclipse.equinox.registry

### Ant based from start of mojo until API result
2023-03-17T06:02:33.367Z
2023-03-17T06:02:48.358Z

15 seconds

### Tycho mojo from start of mojo until API result
2023-03-17T06:12:01.651Z
2023-03-17T06:12:13.958Z

12 seconds


So at least it seems not slower than before ;-)

laeubi:
Finally it can perform a full run which takes `51min 33s`

akurtakov:
Should we run I-build to verify it ?

HannesWell:
Great! Lets see it in action.

> Should we run I-build to verify it ?
>
AFAICT I-Builds don't run the api-checks, only verification builds in the sub-repos do.

PR #974, opened 2023-03-17
Test running the aggregator including api checks

laeubi:
# NOT FOR MERGE!

FYI @HannesWell lets see if we can have an aggregator build including api checks to get some times

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -57,7 +57,8 @@ pipeline {
 						-Dmaven.test.skip=true -DskipTests=true -DaggregatorBuild=true \
 						-DapiBaselineTargetDirectory=${WORKSPACE} \
 						-Dgpg.passphrase="${KEYRING_PASSPHRASE}" \
-						-Dproject.build.sourceEncoding=UTF-8 
+						-Dproject.build.sourceEncoding=UTF-8 \
+						-Papi-check
 					'''
 				}
 


laeubi:
Seems running the API tool in the aggregator is not working well :-\

HannesWell:
> Seems running the API tool in the aggregator is not working well :-\

Too bad. But from the looks I cannot see the reason immediately.

laeubi:
> > Seems running the API tool in the aggregator is not working well :-\
> 
> Too bad. But from the looks I cannot see the reason immediately.

I have created 
- https://github.com/eclipse-pde/eclipse.pde/issues/512

the path looks rather strange so probably it is more a "invalid name" than really "not readable", maybe @merks has more insight if this is a common naming for a profile or if some caller is passing wrong data... strange enough it seems that for the local verification build things are working fine, so maybe some leftovers from a previous run as workspaces are reused here?

merks:
I see that org.eclipse.equinox.internal.p2.engine.SimpleProfileRegistry.escape(String) produces names just like what we see, including names with a ';' in them.  On Windows I cannot create a filename that's so long:

> $touch 'TARGET_DEFINITION%58;ile%58;%47;home%47;jenkins%47;agent%47;workspace%47;latform.releng.aggregator_PR-974%47;eclipse.platform.ui%47;bundles%47;org.eclipse.core.databinding.observable%47;target%47;org.eclipse.core.databinding.observable-apiBaseline.target-816379513.profile'touch: cannot touch 'TARGET_DEFINITION%58;ile%58;%47;home%47;jenkins%47;agent%47;workspace%47;latform.releng.aggregator_PR-974%47;eclipse.platform.ui%47;bundles%47;org.eclipse.core.databinding.observable%47;target%47;org.eclipse.core.databinding.observable-apiBaseline.target-816379513.profile': File name too long

But otherwise the file name looks fine.  

Such IDs are created by org.eclipse.pde.internal.core.target.P2TargetUtils.getProfileId(ITargetHandle) which likes uses org.eclipse.pde.internal.core.target.ExternalFileTargetHandle.getMemento() to return a URI leading to this very long name.

But it's always worked this way so I doubt the name is invalid...

laeubi:
> But it's always worked this way so I doubt the name is invalid...

It might be that the name is now too long because the repo name `platform.releng.aggregator` is longer as for example `platform.ui`... what do you think of using the hashcode of the string instead? 
Would it harm anything (e.g loosing cache data)?

Linux also complains that the name is too long, if I remove 17 chars it works...

merks:
I see!  I think part of the problem is that the code trims to 200 characters but then the encoding steps of taking 1 character and turning it into 4 characters, e.g., `%00;`, which can make the string much longer again.  
```
	private static String getProfileId(ITargetHandle handle) throws CoreException {
		StringBuilder buffer = new StringBuilder();
		buffer.append(PROFILE_ID_PREFIX);
		// Memento strings can be very long and exceed max filename lengths, trim down to 200 + prefix + hashcode
		String memento = handle.getMemento();
		if (memento.length() > 200) {
			buffer.append(memento.substring(memento.length() - 200));
			buffer.append(memento.hashCode());
		} else {
			buffer.append(memento);
		}
		return buffer.toString();
	}
```
I wonder if we change the algorithm if we might break existing workspaces (target platforms)?   If I were to change it, I'd use a much shorter names and I'd use something like sha1 of the full memento string for a much more collision-proof hash code.  Maybe I'd even removed all the "bad" characters that encode encodes to replace them with _ and rely on the sha1 to make the result unique...

laeubi:
If I understand this correct, this is to get exclusive lock across process boundaries, so if the name changes it might be that an "old" Eclipse is thinking it can lock and a "new" eclipse as well as the use different files.

So one strategy would be that we create two locations, one "the old way" and one with a SHA512, then if the lock fails on the first with I/O exception try the second and if that also fails we are lost anyways...

merks:
Yes, the part that fails visibly involves locking, but in general the profile ID is used as part of the file location where profile itself is maintained, e.g., org.eclipse.equinox.internal.p2.engine.SimpleProfileRegistry.saveProfile(Profile) does this:
````
	private void saveProfile(Profile profile) {
		File profileDirectory = getProfileFolder(profile.getProfileId());
		profileDirectory.mkdir();
```
So this isn't purely an issue of locking but rather an issue of the profile's ID being too long which affects whether the profile itself is found or not based on that ID. 

I suspect in the worst case, profile just needs to be created from it's definition again if it's not found with a new ID algorithm.  But it's definitely something worth testing manually to ensure that the symptom isn't too horrible...

laeubi:
Lets see if the fix from @merks changes anything here... but maybe we need to wait for a new ibuild for it to be picked up!?

laeubi:
Lets try with the new i-build...

PR #975, opened 2023-03-17
Stop building Ubuntu 18/20 images

akurtakov:
Image building is hitting the limits of build server so aim is to have only latest LTS version per major distro.
See https://gitlab.eclipse.org/eclipsefdn/helpdesk/-/issues/2827


akurtakov:
Verification build will not verify this one.

PR #977, opened 2023-03-17
Move 'tycho-apitools-plugin:generate' execution to build/plugins section

HannesWell:
The `tycho-apitools-plugin:generate` mojo already checks if a project has a manifest file and if not silently skips its execution. Therefore it is not necessary to guard it with a profile activated with the condition 
`<file><exists>META-INF/MANIFEST.MF</exists></file>`.
Mojo code:
https://github.com/eclipse-tycho/tycho/blob/757d125589b36f68bd553db9140803ad8c9bcef1/tycho-apitools-plugin/src/main/java/org/eclipse/tycho/apitools/ApiFileGenerationMojo.java#L85-L86

See also https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/pull/973.

From the discussion in https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/pull/419#issuecomment-1246575930 I wonder if the generate-mojo should skip its execution if the .project files does not have the `org.eclipse.pde.api.tools.apiAnalysisNature` nature.

### DIFF
--- a/eclipse-platform-parent/pom.xml
+++ b/eclipse-platform-parent/pom.xml
@@ -388,6 +388,22 @@
          </execution>
        </executions>
      </plugin>
+     <plugin>
+       <groupId>org.eclipse.tycho</groupId>
+       <artifactId>tycho-apitools-plugin</artifactId>
+       <version>${tycho.version}</version>
+       <executions>
+         <execution>
+           <id>generate</id>
+           <goals>
+             <goal>generate</goal>
+           </goals>
+           <configuration>
+             <projectName>${project.artifactId}_${qualifiedVersion}</projectName>
+           </configuration>
+         </execution>
+       </executions>
+     </plugin>
     </plugins>
     <pluginManagement>
       <plugins>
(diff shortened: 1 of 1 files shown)


laeubi:
>  I wonder if the generate-mojo should skip its execution if the .project files does not have the org.eclipse.pde.api.tools.apiAnalysisNature nature.

I'm currently prepare check for natures in Tycho, but beside that the file will be generated we would not gain much I think it will then simply be empty.

laeubi (approved):
Looks sane to me, but @akurtakov can probably better tell about the history of this profile

akurtakov:
I would have to redirect to @mickaelistria here, he was the one that made this beast work in the maven build.

HannesWell:
> > I wonder if the generate-mojo should skip its execution if the .project files does not have the org.eclipse.pde.api.tools.apiAnalysisNature nature.
> 
> I'm currently prepare check for natures in Tycho, but beside that the file will be generated we would not gain much I think it will then simply be empty.

The motivation in the old profile was performance as mentioned in https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/pull/419#issuecomment-1246575930, we could avoid running the generator and to create an empty file. To be useful of course the nature check has to be fast enough to not eat up the benefits of not running the generator and writing the file.

HannesWell:
Since https://github.com/eclipse-tycho/tycho/pull/2250 the `tycho-apitools-plugin:generate` skips its execution if a project does not have the api-tools nature.
Therefore I think there is no blocker anymore for this one since the mojo behaves the same in this regard.
Furthermore it looks like the `skipAPIDescription` property is referenced nowhere.

HannesWell:
Comparing the runtimes of the latest runs of this PR's build (running 33min) with other PRs and the master (which run between 28 and 59min), I think there is no performance regression by this change.
Merging since there were no objections and I see only advantages.

PR #987, opened 2023-03-21
Unify the platform workflow by defining reusable workflows

laeubi:
Currently there is a quite inconsitent mixture of ci workflows, some are outdated or use different aproaches and configs, some have no workflow defined at all.

This defines (and documents) a set of reusable workflows that can be adjusted / updated at a central place and then reused in the platform repositories.

### DIFF
--- a/.github/matcher/api-tools.json
+++ b/.github/matcher/api-tools.json
@@ -1,0 +1,16 @@
+{
+  "problemMatcher": [
+    {
+      "owner": "api-problems",
+      "pattern": [
+        {
+          "regexp": "^\\[API\\s(\\w+)\\].*line\\s(\\d+):\\s(.+)\\(location:\\s(.+)\\)$",
+          "severity": 1,
+          "line": 2,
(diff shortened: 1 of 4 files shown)


akurtakov:
Thanks @laeubi ! This kind of work is badly needed and is exactly what could make working on eclipse codebase less weird.

laeubi:
> This kind of work is badly needed and is exactly what could make working on eclipse codebase less weird.

I hope so, lets see how this works out!

HannesWell:
> Thanks @laeubi ! This kind of work is badly needed and is exactly what could make working on eclipse codebase less weird.

I can only seconds that. Unify the configuration more and more is good.

That reminds me about the Idea to activate the build-individual-bundels profile by a maven.config in the aggregators submodules so that one can just build them with `mvn clean verify`.
@laeubi do you want to apply that and corresponding doc updates when you are going through the repos anyway? Or have you already completed that task?

PR #990, opened 2023-03-23
Adjust the setup to use a specific maven version

laeubi:
Currently we use the maven version of the runner what could produce some confusion if different runners use different maven versions, therefore we now use a fixed maven version to install for all runners.

### DIFF
--- a/.github/workflows/mavenBuild.yml
+++ b/.github/workflows/mavenBuild.yml
@@ -44,6 +44,10 @@ jobs:
           17
         distribution: 'temurin'
         cache: maven
+    - name: Set up Maven
+      uses: stCarolas/setup-maven@v4.5
+      with:
+        maven-version: 3.8.8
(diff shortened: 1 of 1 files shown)


.github/workflows/mavenBuild.yml
```diff
@@ -44,6 +44,10 @@ jobs:
           17
         distribution: 'temurin'
         cache: maven
+    - name: Set up Maven
+      uses: stCarolas/setup-maven@v4.5
+      with:
+        maven-version: 3.8.8
```
HannesWell, line 50: Is there a reason why we don't use Maven 3.9.1? From my experience this works with the latest Tycho or are there still issues?

laeubi: Mostly because the Jenkins uses Maven 3.8.x and as there is no real benefit in using the 3.9.x line yet.

PR #1013, opened 2023-04-06
Disabling Milestone Schedule and add regular schedule.

MohananRahul:
@sravanlakkimsetti

### DIFF
--- a/README.md
+++ b/README.md
@@ -37,7 +37,7 @@ git submodule update
 mvn clean verify  -DskipTests=true
 
 # find the results in
-# eclipse.platform.releng.tychoeclipsebuilder/sdk/target/products/*
+# eclipse.platform.releng.tychoeclipsebuilder/eclipse.platform.repository/target/products
 ```
 
 Build requirements


akurtakov:
Last release @SDawley created a more complicated cron entry for all the M/RC weeks at once so it didn't had to be edited so often. It would be nice to be done that way again. Sam, would you point to the example how it was done?

laeubi:
This seems to be a groovy script so maybe the calendar that is used by @HannesWell could be used to dynamically select the right patterns here?

HannesWell:
> This seems to be a groovy script so maybe the calendar that is used by @HannesWell could be used to dynamically select the right patterns here?

That would probably be possible, you can find the bash script here:
https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/blob/master/.github/workflows/verifyFreezePeriod.yml

Nevertheless the calender approach has not prooven itself in the past to be to most stable/reliable one.
Therefore I hesitate a bit to advertise its broader use.

PR #1018, opened 2023-04-07
Mention reopening of the master branch after freeze week

HannesWell:
At the moment it seems unclear when the master branches will reopen after a freeze period.
After the RC freeze period there is a dedicated mail that announces the reopening of the master branch after all preparation work has completed. As far as I understand there is no preparation work form M1 to M2 and therefore the master branch is open as as soon as M1 has been officially released/declared.
Therefore it would be good if the mail to announce the M1 release will also mention that the master branch is now open again.
In case there wont be a M1 freeze in the future this section can be removed again (https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/discussions/863).

@MohananRahul, @sravanlakkimsetti can you please check if that is correct?


akurtakov:
Good by me.

PR #1032, opened 2023-04-19
Adjust the path of the results

laeubi:
The mentioned path in the readme does not exits and results seem to be moved under eclipse.platform.repository now.

@akurtakov can you verify this? At least for my local build I did not find the mentioned location but the proposed new one here where some products are located.


akurtakov (approved):
Looks correct.

akurtakov:
The build failure is irrelevant for this documentation patch.

PR #981, opened 2023-03-17
Update the Eclipse SDK TP's slf4j-api to 2.0.7 and remove 1.7 version

HannesWell:
Since version 2.0.7 the slf4j.api bundle exports the user-api package 'org.slf4j' in version 2.x.y and 1.7.36.
This allows to also use libraries, which are programmed against slf4j-1 and import the package 'org.slf4j' with an exclusive upper bound of 2 like '[1.7,2)', with the slf4-api version 2 in OSGi environments.

According to SLF4J the client/user facing API is compatible with slf4j-1:
- https://www.slf4j.org/faq.html#changesInVersion200
- https://www.slf4j.org/faq.html#compatibility

Only the logging-backends/bindings/providers have match the requirements of the specific slf4j version.

The corresponding SLF4J issue and change is
- https://jira.qos.ch/browse/SLF4J-579
- https://github.com/qos-ch/slf4j/pull/331

Fixes https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/issues/588
Fixes https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/issues/682

PR #1047, opened 2023-04-25
Update target platform slf4j-api 2.0.7

merks:


akurtakov:
@HannesWell Is it OK not to get this update?

HannesWell:
> @HannesWell Is it OK not to get this update?

Yes, I just wanted to mention that I already created https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/pull/981 to update to slf4j-2.0.7. That PR is just awaiting a new release of SSHD-Mina and then we can get rid of slf4j-1 completely.
Without the new sshd version we have both the slf4j 1 and 2 api bundle which either does not work well together or is complex to set up (one probably has to configure both and has to provide two logging backends, but I have never tried that).

HannesWell:
A new release of mina-sshd is hopefully done this week:
https://github.com/apache/mina-sshd/pull/343#issuecomment-1512923050

HannesWell:
> No problem! I'll just abandon this. I need to update the TP for EMF's M2 and you can update this at your leisure at the appropriate time.

Thanks. AFAIKT it is really only the sshd release missing and I'm waiting for it longingly, since it seem to be the last missing piece to complete the transition to slf4j-2. I even used a self-build snapshot of sshd in #981 to verify it.

PR #1055, opened 2023-04-29
Use Maven 3.9.1 in mavenBuild workflows

HannesWell:
- Update mavenBuild workflow to use maven 3.9.1

- Ignore Maven test errors and failures in mavenBuild workflow
This allows the Maven build check itself to pass and have a green check mark in the GH UI if there are only test errors/failures. The test errors/failures themself are still reported via the 'Publish Test Results' checks that is shown as failure if there are test failures/errors. This setup simplifies to distinguish between only test errors/failures in the build and more severe failures like compilation failures.

### DIFF
--- a/.github/workflows/mavenBuild.yml
+++ b/.github/workflows/mavenBuild.yml
@@ -47,7 +47,7 @@ jobs:
     - name: Set up Maven
       uses: stCarolas/setup-maven@v4.5
       with:
-        maven-version: 3.8.8
+        maven-version: 3.9.1
     - name: Download the API Tools matcher
       uses: suisei-cn/actions-download-file@v1.3.0
       id: api-tools-matcher


laeubi (dismissed):
> Ignore Maven test errors and failures in mavenBuild workflow

This is especially desired, the check should fail if there is a test failure, so this false behavior should not be reintroduced (again).

HannesWell:
> > Ignore Maven test errors and failures in mavenBuild workflow
> 
> This is especially desired, the check should fail if there is a test failure, so this false behavior should not be reintroduced (again).

But wouldn't the over-all Test result check fail if there is a test failure/error?
As described in my initial comment personally I would find this more convenient to distinguish test-failures/errors from other build failures.

laeubi:
As said it is not recommended to do so and was explicitly complained about, so I don't want to revert this back and forth....

HannesWell:
> As said it is not recommended to do so and was explicitly complained about, so I don't want to revert this back and forth....

OK, then lets keep this as it is. So this PR only updates the build workflows to run with maven 3.9.1.
The current Tycho 4 snapshot works well with that, doesn't it?

laeubi:
Tycho 4 was updated to Maven 3.9.x already.

akurtakov:
This seems to be fine and @laeubi 's change request is done so it's good to go.

Issue #1056, opened 2023-04-29
Migrate javax.inject and javax.annotations to Jakarta


HannesWell:
This is an umbrella issue to coordinate and collect the migration steps of the `javax.inject` and `java.annotations` bundles to Jakarta.

- [x] Migrate to the javax.inject|annotations bundles provided by Jakarta but with unchanged package names
    1. Only use Import-Package with a closed version range to reference `javax.inject|annotations` in Plugin Manifests.
        - Simplifies the exchange of the providing bundle
        - Additionally prevents possible split-package problems with the `javax.annotation` package provided by the `com.google.code.findbugs/jsr305` artifact. It is not used within the Eclipse platform but it guards down-stream consumers that could have it in their TP. E.g. Guava has a transitive dependency on it, so there is some chance it ends up in a TP if one is not careful.
        - PRs
          - https://github.com/eclipse-pde/eclipse.pde/pull/591
          - https://github.com/eclipse-jdt/eclipse.jdt.core/pull/1019
          - https://github.com/eclipse-platform/eclipse.platform/pull/431
          - https://github.com/eclipse-platform/eclipse.platform.debug/pull/133
          - https://github.com/eclipse-platform/eclipse.platform.ua/pull/121
          - https://github.com/eclipse-platform/eclipse.platform.ui/pull/738

    2. Refactor code to not relay on the bundle-symbolicName of the javax-annotations provider bundle so that it can be easaly exchanged
        - https://github.com/eclipse-platform/eclipse.platform.ui/pull/745
        - https://github.com/eclipse-platform/eclipse.platform.common/pull/145
        - https://github.com/eclipse-pde/eclipse.pde/pull/593
      
    4. Replace javax-artifacts in the TP with the ones provided by Jakarta
        - https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/pull/1057 
        - https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/pull/1059

- [x] Support Annotations from `jakarta.inject|annotations` in all processors in the Eclipse platform
    - Add new jakarta.inject|annotations bundles to TP which version?
      - jakarta.inject:jakarta.inject-api:2.0.1
      - jakarta.annotation:jakarta.annotation-api:2.1.1
    - Support Jakarta Annotations in the E4 Injector (will create a dedicated issue):
      - https://github.com/eclipse-platform/eclipse.platform/pull/696
[...]

PR #1057, opened 2023-05-04
Add javax.inject/annotation provided by Jakarta from Maven-Central and include all sources in Eclipse SDK repo

HannesWell:
Use the javax namespace bundles provided by Jakarta.
Remove all javax bundles provided by Orbit.

Remove the unused javax.xml bundle, the modern JREs provide the contained packages by default.

Additionally configure the `tycho-p2-repository-plugin` to include all sources. This allows to remove the javax-bundles from containing features (see https://github.com/eclipse-platform/eclipse.platform.ui/pull/745) and at the same time to keep the main artifacts (which are pulled in as transitive dependencies) plus their sources in the created p2-repo for the Eclipse SDK.

### DIFF
--- a/eclipse.platform.releng.prereqs.sdk/eclipse-sdk-prereqs.target
+++ b/eclipse.platform.releng.prereqs.sdk/eclipse-sdk-prereqs.target
@@ -157,6 +157,18 @@
 				  <type>jar</type>
 			  </dependency>
 			  <dependency>
+				  <groupId>jakarta.annotation</groupId>
+				  <artifactId>jakarta.annotation-api</artifactId>
+				  <version>1.3.5</version>
+				  <type>jar</type>
+			  </dependency>
+			  <dependency>
+				  <groupId>jakarta.inject</groupId>
+				  <artifactId>jakarta.inject-api</artifactId>
+				  <version>1.0.5</version>
+				  <type>jar</type>
+			  </dependency>
+			  <dependency>
 				  <groupId>net.java.dev.jna</groupId>
 				  <artifactId>jna-platform</artifactId>
 				  <version>5.13.0</version>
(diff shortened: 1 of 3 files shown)


.gitmodules
```diff
@@ -43,7 +43,8 @@
   url = https://github.com/eclipse-platform/eclipse.platform.ua.git
 [submodule "eclipse.platform.ui"]
   path = eclipse.platform.ui
-  url = https://github.com/eclipse-platform/eclipse.platform.ui.git
+  url = https://github.com/HannesWell/eclipse.platform.ui.git
+  branch = removeJavaxAnnotationsFromFeature
```
HannesWell, line 47: This has to be removed before this PR is eventually merged.

HannesWell:
For now I changed this PR to only add the jakarta-bundles because of https://github.com/eclipse-platform/eclipse.platform.common/pull/145#issuecomment-1535932554.

The removal of the javax-bundles will happen in a follow up, after a nightly build.

HannesWell:
Btw. @akurtakov, @merks, @laeubi with configuring the `tycho-p2-repository-plugin` with `includeAllSources=true`, we can actually start to remove more third party-bundles from features.

In the last discussion about removing them, then main argument to leave them in was that this way it is ensured that the sources are included into the p2-repo and therefore available without any further magic (like m2e).
With that change the mentioned bundles will get their sources included into the repo even if only pulled in as transitive dependency.
If we remove them we can save us enforec-qualifier updates in the features if a version is updated or adaptions to BSN changes.

merks:
Yes the force qualifier updates changes are a pain in the neck.  The cbi aggregator also aggregates corresponding source bundles for every bundle otherwise aggregated so this would continue to work well for SimRel.

laeubi:
If the bundle is from maven (central) and m2e is available the sources should even be downloaded if not in the update-site.

PR #1059, opened 2023-05-08
Remove all javax bundles provided by Eclipse-Orbit

HannesWell:
Follow up on https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/pull/1057 to remove the now unused javax.xml bundle, the modern JREs provide the contained packages by default.

Also revert explicit inclusion of jakarta.inject/annotation bundles in eclipse.platform.repository, because they will be transitively included from now on.

This probably requires the currently running I-build to succeed first (see https://github.com/eclipse-platform/eclipse.platform.common/pull/145#issuecomment-1535932554).

### DIFF
--- a/eclipse.platform.releng.prereqs.sdk/eclipse-sdk-prereqs.target
+++ b/eclipse.platform.releng.prereqs.sdk/eclipse-sdk-prereqs.target
@@ -7,9 +7,6 @@
         by glassfish. Should we use it? -->
       <unit id="com.jcraft.jsch" version="0.1.55.v20221112-0806"/>
       <unit id="com.jcraft.jsch.source" version="0.1.55.v20221112-0806"/>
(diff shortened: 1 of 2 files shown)


akurtakov:
Would it be simpler to split javax.xml removal in separat commit ? For the sake of having less reasons for aggregator build failing.

HannesWell:
> Would it be simpler to split javax.xml removal in separat commit ? For the sake of having less reasons for aggregator build failing.

In general yes, but after searching `javax.xml` in `MANIFEST.MF`, `*.product`, `feature.xml` and `pom.xml` in the aggregator repo and its sub-modules I now have hopefully covered all cases.

HannesWell:
It took another round, but now eventually the build succeeded.

PR #1066, opened 2023-05-15
Update Apache SSHD OSGi to 2.10.0

HannesWell:
Required for https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/pull/981.

@tomaswolf FYI

### DIFF
--- a/eclipse.platform.releng.prereqs.sdk/eclipse-sdk-prereqs.target
+++ b/eclipse.platform.releng.prereqs.sdk/eclipse-sdk-prereqs.target
@@ -176,7 +176,7 @@
 			  <dependency>
 				  <groupId>org.apache.sshd</groupId>
 				  <artifactId>sshd-osgi</artifactId>
-				  <version>2.9.2</version>
+				  <version>2.10.0</version>
 				  <type>jar</type>
 			  </dependency>
 			  <dependency>


.gitmodules
```diff
@@ -49,7 +49,7 @@
   url = https://github.com/eclipse-equinox/equinox.binaries.git
 [submodule "equinox"]
   path = equinox
-  url = https://github.com/eclipse-equinox/equinox.git
+  url = https://github.com/HannesWell/equinox.git
```
HannesWell, line 52: Just for testing the qualifier bump. Will be reverted, once verified.

HannesWell: Reverted.

tomaswolf:
Shouldn't the license check have failed? It does fail in JGit: https://git.eclipse.org/r/c/jgit/jgit/+/201879 .

mickaelistria:
> Shouldn't the license check have failed? 

that could mean that this new version of sshd defined in the .target is actually not the one used during the build in the features/bundles that require it, ie another version is available in the build and is used.

HannesWell:
> Shouldn't the license check have failed? It does fail in JGit: https://git.eclipse.org/r/c/jgit/jgit/+/201879 .

It didn't fail, but it is strange that it failed for you. The review requested that were created for that Gerrit look quite empty. This could indicate that there is an issue in the license-check build setup?


> > Shouldn't the license check have failed?
> 
> that could mean that this new version of sshd defined in the .target is actually not the one used during the build in the features/bundles that require it, ie another version is available in the build and is used.

I think that's unlikely. Otherwise https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/pull/981 would not have succeeded. 
I also just checked the build logs and every-time `sshd.osgi` or `sshd-osgi` occurs it is always in conjunction with version 2.10 at least for the Jenkins build. 
But what could be possible is that the TP is not fully resolved for the license check anymore? Maybe due to the new delayed TP resolution in Tycho 4?

mickaelistria:
Here is what I get with CLI:
```
$ echo 'maven/mavencentral/org.apache.sshd/sshd-osgi/2.10.0' | java -jar /home/mistria/git/dash-licenses/core/target/org.eclipse.dash.licenses-0.0.1-SNAPSHOT.jar -
[main] INFO Querying Eclipse Foundation for license data for 1 items.
[main] INFO Found 0 items.
[main] INFO Querying ClearlyDefined for license data for 1 items.
[main] INFO Found 1 items.
[main] INFO License information could not be automatically verified for the following content:
[main] INFO 
[main] INFO maven/mavencentral/org.apache.sshd/sshd-osgi/2.10.0
[main] INFO 
[main] INFO This content is either not correctly mapped by the system, or requires review.
```
So I suspect that the license-check of Platform is wrong here and fails at checking the right things.

HannesWell:
I also ran the maven command (`mvn -U org.eclipse.dash:license-tool-plugin:license-check -Ddash.fail=true`) locally for this repo and the generated dash/summary file indeed only lists the projects in the reactor.
I guess the dash-licenes maven plugin fails to read the dependencies in the target platform. I'm relatively sure it identified them in the past.

mickaelistria:
> I'm relatively sure it identified them in the past.

I confirm that it used to properly process 3rd party Maven artifacts from .target. We've used it a lot in other projects and could see proper reporting for those.

merks:
Yes, I've asked @waynebeaton about a similar concern where I was pretty sure the tool was checking the dependencies in the *.target in the past it but does not seem to be doing that in the present, e.g., these two files are empty and I seem to recall this listed all the libraries in the past.

https://ci.eclipse.org/oomph/job/simrel/job/simrel-maven/lastSuccessfulBuild/artifact/maven/target/dash/

laeubi:
> But what could be possible is that the TP is not fully resolved for the license check anymore? Maybe due to the new delayed TP resolution in Tycho 4?

It depends how the dash tool works, if it claims it does not need any dependencies (`requiresDependencyResolution = NONE` what is the default) then no dependencies are computed by Tycho if no other mojos are executed.

One could test this quite easy by passing `-Dtycho.target.eager=true` what forces Tycho to resolve the dependencies even if no mojo requires any dependencies.

Maybe Tycho also can be enhanced to also look for `requiresDependencyCollection` mojo parameter, but currently there is not really a distinction to this weaker assert, from the maven documentation:

> Gets the scope of (transitive) dependencies that should be **collected**. Dependency collection refers to the process of calculating the complete dependency tree in terms of artifact coordinates. **In contrast to dependency resolution, this does not include the download of the files** for the dependency artifacts. It is meant for mojos that only want to analyze the set of transitive dependencies, in particular during early lifecycle phases where full dependency resolution might fail due to projects which haven't been built yet.

waynebeaton:
The Maven plugin uses the Maven Reactor. The mojo is currently configured to use `requiresDependencyResolution = ResolutionScope.TEST`. About three weeks ago, I changed it to `requiresDependencyResolution = ResolutionScope.TEST` for a day before reverting the change due to some reports of breakage.

I don't know how Tycho does resolution, so I have no idea whether or not it looks at `*.target` files. If adding `-Dtycho.target.eager=true` to the invocation of the tool solves the problem, then I can add that to the documentation. If anybody has a better idea, then pull requests are welcome.

HannesWell:
Using `-Dtycho.target.eager=true` indeed ensures that dependencies from the TP are considered as well.
That it is not using without it properly is because in TychoProjectExecutionListener.beforeProjectLifecycleExecution() only the reactor project is resolved.

But maybe it could be auto-detected in `TychoProjectExecutionListener.beforeProjectLifecycleExecution()` that all projects of the session need to be resolved (like in `TychoMavenLifecycleParticipant.afterProjectsRead()/resolveProjects()`), if a project specifies `requiresDependencyResolution != null` in conjunction with `aggregator = true` in its Mojo annotation.

Nevertheless I think projects like M2E where we use a mixed reactor (that contains 'pure' Maven projects that have a Manifest generated by the maven-bundle-plugin or bnd-maven-plugin and are referenced by Manifest-first Eclipse Plugin projects) still make that difficult. At least with `-Dtycho.target.eager=true` the eager resolution fails stating that the 'pure' projects are missing (which makes sense because the metadata have not been generated).
One simple way I can think of that may works, is to be able to configure Tycho to not fail the build if the resolution fails and then just work with what was resolved.
Otherwise to get a clean resolution the manifest of those 'pure' Maven projects would have to be build, even for a license check.

laeubi:
Please open an issue at Tycho with steps how to reproduce the problem (at best by provide an [integration-test](https://github.com/eclipse/tycho/wiki#providing-an-integration-test) to demonstrate the issue), in general `-Dtycho.target.eager=true` should be avoided whenever possible as it limits the feature set as you already noticed a lot and is not a general purpose switch that fits all.

Also if one allows to fail the dependency resolution this means that not all dependencies of a project are considered. just assume you are referencing a bundle indirectly through a package import then it won't be included.

Just in general `requiresDependencyResolution != null` in conjunction with `aggregator = true` seems what most likely is the root-cause here, because as noted above this will mean projects are possibly never executed, so the dependencies are probably incomplete, but executing this for all projects would fail for the same reason as it currently fails for `-Dtycho.target.eager=true` (since no manifest is generated).

Neverless I see some options here:

1. I recently added some code for the new PDE automatic manifest generation feature to compute a preliminary manifest using the source-code based on BND instructions, this might be enhanced for the m2e case so `-Dtycho.target.eager=true` works (but as noted would require a test-case)
2. The license-check mojo can be enhanced to accept a list of dependencies (maybe some kind of bom?) where Tycho list all items in a given format, e.g there is already [tycho-p2-plugin:dependency-tree](https://github.com/eclipse-tycho/tycho/blob/master/RELEASE_NOTES.md#new-tycho-p2-plugindependency-tree-mojo)
3. Similar to the second, one might inject dummy dependencies to `target` projects, but this will then include everything that is in the target (not what is actually used by projects), what maybe is not that bad.

merks:
@waynebeaton 

Thanks, that did the trick for me too and even created issues:

- https://gitlab.eclipse.org/eclipsefdn/emo-team/iplab/-/issues/8461
- https://gitlab.eclipse.org/eclipsefdn/emo-team/iplab/-/issues/8462

laeubi:
I have now added a patch to Tycho that enhances the concept of preliminary metadata to the `bnd-maven-plugin`:

- https://github.com/eclipse-tycho/tycho/pull/2418

that way m2e can use

`mvn org.eclipse.dash:license-tool-plugin:license-check -Ddash.fail=true -Dtycho.target.eager=true -Dtycho.target.pomDependencies=consider`

but there is one restriction: As m2e embeds some jars, one first need to execute the `generate-resources` pahsel as otherwise the required libs are missing and the manifest data is incomplete. On the long run, I think we should not try to build a fat jar but consume this directly from the target platform, that way everything is generated beforehand and the issue is gone and we will produce much more lightweight distribution (and maybe give a less harder time to the license check tool).

laeubi:
See also:

- https://github.com/eclipse/dash-licenses/pull/239

HannesWell:
I tried another approach that does not require eager resolution, allows the simple addition of extra goals/phases for cases like mixed reactors and also allows to only run the license check for specific plugins.

The idea is to not run the license-check mojo only once as aggregator (like it is now) but instead run regularly for every project in the reactor. Then, when used with Tycho, the dependency resolution can happen lazy as in the default case and the mojo can collect the project artifacts as before.
This way one can also exclude projects from being checked by simplify skipping the license-check execution for them.
Furthermore the dependency resolution can be parallelized.

For the case of a mixed reactor like in m2e one can then call this
`mvn clean package -DskipTests org.eclipse.dash:license-tool-plugin:license-check -Ddash.fail=true` and the list of dependencies is as expected.
For the simple case like in the platform.releng.aggr one only has to call 
`mvn org.eclipse.dash:license-tool-plugin:license-check -Ddash.fail=true` like before.
I tested that case locally as well but that fails with the following error
```
Could not resolve dependencies for project org.eclipse.jdt:org.eclipse.jdt.ui.examples.projects:eclipse-plugin:3.2.0-SNAPSHOT: The
following artifacts could not be resolved: org.eclipse.jdt.ui:org.eclipse.jdt.ui.junit.sampleproject:jar:dist:1.0.0-SNAPSHOT (absent): 
Could not find artifact org.eclipse.jdt.ui:org.eclipse.jdt.ui.junit.sampleproject:jar:dist:1.0.0-SNAPSHOT
```
But I assume that's a issue with the aggr build itself.

I just have created a PR for that:
- https://github.com/eclipse/dash-licenses/pull/241

laeubi:
> The idea is to not run the license-check mojo only once as aggregator (like it is now) but instead run regularly for every project in the reactor. 

I think this is actually a very good approach here, especially as with Tycho 5 the plan is to even loosen the whole dependency binding more ...

PR #1074, opened 2023-05-17
Update jackson to 2.15.1 and commons-io to 2.12.0

merks:


merks:
@akurtakov 

I know it's late, but it seems maybe more sensible to update for M3 than to wait for RC1...

akurtakov:
I fully agree. If a dependency update is for 2023-06 it would better be done for M3 if possible so go for it once validation job succeeds.

PR #1091, opened 2023-05-31
Contribute EMF's release repository to the target platform

merks:


merks:
@akurtakov 

I don't know what is causing the license check to fail, but I'm going to ignore it because it's not failing due to any actual license problem.

akurtakov:
Tycho requires Maven 3.9.x now and this action is running 3.8.x. @HannesWell or @laeubi would you please help with this one? I assume license check will fail  from now on.

laeubi:
I'll take a look

PR #1130, opened 2023-06-07
Update batik versions to 1.16.0

MohananRahul:
Update batik version due to Vulnerability issues

### DIFF
--- a/eclipse.platform.releng.prereqs.sdk/eclipse-sdk-prereqs.target
+++ b/eclipse.platform.releng.prereqs.sdk/eclipse-sdk-prereqs.target
@@ -23,14 +23,14 @@
       <unit id="javax.xml" version="1.3.4.v201005080400"/>
       <unit id="org.apache.ant" version="1.10.12.v20211102-1452"/>
       <unit id="org.apache.ant.source" version="1.10.12.v20211102-1452"/>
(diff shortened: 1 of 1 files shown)


sravanlakkimsetti:
Can you please check whether you can build eclipse platform locally with this change? I don't find batik 1.16 libraries in the orbit version used here

akurtakov:
Is this patch dead? aka should it be closed?

akurtakov:
Even if against old branch this helps nothing but convoluting searches staying like that for weeks.

sravanlakkimsetti:
This is a required change. But we will need to add newer Orbit version to fetch latest version of batik. Merging this as is will fail 4.23 build

akurtakov:
So this is a work in progress ? You should be aware that Orbit is not getting much attention (if at all) lately https://git.eclipse.org/c/orbit/orbit-recipes.git/ . Wherever possible things are moving into using Maven artifacts but @jonahgraham or @merks should be the ones to clarify things about Orbit more.

jonahgraham:
> https://git.eclipse.org/c/orbit/orbit-recipes.git/ .

This is the old URL - Orbit is on GitHub - https://github.com/eclipse/orbit

> @jonahgraham or @merks should be the ones to clarify things about Orbit more.

We (Ed and I) are meeting tomorrow to move this forward. Please keep an eye out for communications in the coming days.

PR #1145, opened 2023-06-14
New Script for promoting Equinox.

MohananRahul:
Script for promoting Equinox.

https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/issues/1146

### DIFF
--- a/cje-production/promotion/promoteSitesEquinox.sh
+++ b/cje-production/promotion/promoteSitesEquinox.sh
@@ -330,7 +330,7 @@ export DL_DROP_ID_EQ=${DL_TYPE}-${DL_LABEL_EQ}-${BUILD_TIMESTAMP}
 
 export HIDE_SITE=true
 # Build machine locations (would very seldom change)
-
+export BUILD_ROOT=${BUILD_ROOT:-/home/data/httpd/download.eclipse.org}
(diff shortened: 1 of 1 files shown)


laeubi (changes requested):
Could you please transform this into a github action / Jenkins Job so this can be automated / triggerd by others as well? Also please add a readme.md file that describes how to use this script.

sravanlakkimsetti:
@laeubi there was a failure in promoting equinox. current script promotes eclipse, equinox and repo. During the last run only equinox failed. We want to promote equinox only for that reason we are creating this script temporarily.

laeubi:
When the script is temporary why must it be part of equinox repository?
I think it is good to have the script close to the code but currently it is completely unclear what it does and how it should be used, especially if it seems to be fragile and non repeatable. e.g Jenkins has a nice "replay" feature where one can even modify the `Jenkinsfile` (what can contain shell script commands as well), so probably its worth to rethink the process here, we have a way to much confusing commits that get merged, reverted, bumped, missing bumps and so on that produce a lot of noise in the git-history and mail notifications.

sravanlakkimsetti:
This script is for one time use only. Closing this one

PR #1151, opened 2023-06-15
Fix the minimum maven version declared to require 3.9.0 as minimum

laeubi:


akurtakov (commented):
So many places failed and I haven't thought to adjust that.

PR #1163, opened 2023-06-20
Update Mockito and ByteBuddy versions

akurtakov:
* Mockito to 5.4.0
* ByteBuddy to 1.14.5

### DIFF
--- a/eclipse.platform.releng.prereqs.sdk/eclipse-sdk-prereqs.target
+++ b/eclipse.platform.releng.prereqs.sdk/eclipse-sdk-prereqs.target
@@ -333,13 +333,13 @@
 			  <dependency>
 				  <groupId>net.bytebuddy</groupId>
 				  <artifactId>byte-buddy-agent</artifactId>
-				  <version>1.14.4</version>
+				  <version>1.14.5</version>
 				  <type>jar</type>
 			  </dependency>
 			  <dependency>
 				  <groupId>net.bytebuddy</groupId>
 				  <artifactId>byte-buddy</artifactId>
-				  <version>1.14.4</version>
+				  <version>1.14.5</version>
 				  <type>jar</type>
 			  </dependency>
 			  <dependency>
@@ -351,7 +351,7 @@
 			  <dependency>
 				  <groupId>org.mockito</groupId>
 				  <artifactId>mockito-core</artifactId>
-				  <version>5.3.1</version>
+				  <version>5.4.0</version>
 				  <type>jar</type>
 			  </dependency>
 			  <dependency>


akurtakov:
Licensetool complains due to https://github.com/eclipse/dash-licenses/pull/244 but local run found no issue. 
@laeubi @HannesWell  It seems to be time for moving the shared action to some other project where it can be fixed as dash-license seems to not be active enough and having every build fails for weeks is clearly no go. Can we do this easily?

laeubi:
I have no clue about the "organizational part" of the Eclipse IP process, but technically you can just copy the desired actions here (I'm not completely sure why they are split into two actions but maybe @HannesWell  can tell):

https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/tree/master/.github/workflows

Another idea would be to just call it from the unified workflow instead of an extra workflow, I can try some things in this regard tomorrow if you think it can wait until then we should then be able to perform additional setups as well and only need to delete the old workflow call.

laeubi:
I have put it on my list now for the releng-work, it seems I missed this that it is still a problem due to all the 2023-06 release rushes and for Tycho it has worked well lately.

In general the Eclipse WG has funding some work related to releng tasks so if there is something crucial you can always leave a note here as well  https://gitlab.eclipse.org/eclipse-wg/ide-wg/ide-wg-dev-funded-efforts/ide-wg-dev-funded-program-planning-council-top-issues/-/issues/14 and I'll try to pick this up then (or just leave me a note on a github issue).

laeubi:
> The dash tool ran okay and created issues for me the other day
> 
> https://ci.eclipse.org/oomph/job/simrel/job/simrel-maven/74/console
> 
> Note that all the minor updates in the reports have been scanned in this way by dash.

Because Oomph uses Tycho 2.7.5 (what is Maven 3.8 + ) but Tycho 4 requires Maven 3.9+ ...

merks:
That can’t be the reason

https://github.com/eclipse-oomph/oomph.incubator/blob/f195fa2bc38869e593160546e87be57eab51d216/maven/pom.xml#L12

The log mentions maven 3.9.1 too. 

I think most of the incubator uses Tycho 4.0.0.

akurtakov:
@merks You are probably running the maven plugin as part of your own build but dash-license provides reusable GH action https://github.com/eclipse/dash-licenses/tree/master/.github/actions/maven-license-check-action which is supposed to be usable independend of the build and as separate action that can fail early . https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/blob/master/.github/workflows/licensecheck.yml is how platform uses this ghaction.

merks:
I see!  Indeed I just use it as part of the regular build and parameterized the build to fail (by default) or continue. In either case it opens issues as needed. 

I know @waynebeaton is super busy…

akurtakov:
He can't do everything so has to delegate more to people (in this case making @HannesWell committer)

HannesWell:
> He can't do everything so has to delegate more to people (in this case making @HannesWell committer)

I would be glad to be able to maintain the workflow in dash-licenes as well as be able to proceed with https://github.com/eclipse/dash-licenses/pull/241.


> @laeubi @HannesWell It seems to be time for moving the shared action to some other project where it can be fixed as dash-license seems to not be active enough and having every build fails for weeks is clearly no go. Can we do this easily?

Yes we can, I just created https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/pull/1164. It is basically just a copy of the original with adjusted paths.

> I have put it on my list now for the releng-work, it seems I missed this that it is still a problem due to all the 2023-06 release rushes and for Tycho it has worked well lately.

I hope I didn't steal you work you were eager to do, but I assumed it would be quicker if I just do it 🙈 

> I have no clue about the "organizational part" of the Eclipse IP process, but technically you can just copy the desired actions here (I'm not completely sure why they are split into two actions but maybe @HannesWell can tell):
> 
> https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/tree/master/.github/workflows

It is split into a re-usable/callable workflow and a composite action for the reason that only with a composite action it is possible to include resourses. And for this use-case there is a special [maven-settings file](https://github.com/eclipse/dash-licenses/blob/master/.github/actions/maven-license-check-action/licenseCheckMavenSettings.xml) that adds the dash-licenses-snapshots maven-repository so that the latest snapshot of the license-check mojo can be used without the need to specify the maven-repo in each project that uses the license-check workflow.

laeubi:
> It is split into a re-usable/callable workflow and a composite action for the reason that only with a composite action it is possible to include resourses.

https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/blob/430fdcb69b24c8dc18813d2dd804fc4d39754009/.github/workflows/mavenBuild.yml#L51-L56

her you can see how I copy a resource from from another repository to share it between workflows.

> And for this use-case there is a special [maven-settings file](https://github.com/eclipse/dash-licenses/blob/master/.github/actions/maven-license-check-action/licenseCheckMavenSettings.xml) that adds the dash-licenses-snapshots maven-repository so that the latest snapshot of the license-check mojo can be used without the need to specify the maven-repo in each project that uses the license-check workflow.

I think it would be good to simply deploy the tool to maven central anyways to not rely on snapshots and have a better workflow for tool-users.

Nerveless I think you should be able to set `-Dmaven.repo.remote=<repo1>,<repo2>` just in case you want to try it out :-)

laeubi:
@akurtakov @HannesWell  while working on this I noticed that the subprojects do not use the license check at all, is that desired/expected?

akurtakov:
In general new deps are only added through the target platform which submodules get from here so not that big of an issue IMHO. But wouldn't hurt if run is inherited.

laeubi:
Is the license check really detect if we add something to the target (but it is never used yet?)

laeubi:
> But wouldn't hurt if run is inherited.

At alest one needs the gitlab secret then, or are the automatically setup for each repository @HannesWell

laeubi:
By the way now I fixed m2e check and now it complains about a P2 dependency so I think an unvetted item has somehow included in the Eclipse-Release:
https://github.com/eclipse-m2e/m2e-core/pull/1439#issuecomment-1600532860

HannesWell:
> Nerveless I think you should be able to set `-Dmaven.repo.remote=<repo1>,<repo2>` just in case you want to try it out :-)

Yes that's interesting! I assume that I have checked for such a property when building the workflow, but I'll check that again.

> > But wouldn't hurt if run is inherited.
> 
> At alest one needs the gitlab secret then, or are the automatically setup for each repository @HannesWell

A gitlab token is only necessary if you want to create review-requests. And yes, we would have to ask the infra team to add it to each repo. Since deps are only added in this repo, I don't think it is worth the effort.
And the review request also works if the PR is closed. So even if a completly new dependency is added that is first used by no-body you can still request a review later on that PR as soon as a change that makes use of it is merged in another repo.


> By the way now I fixed m2e check and now it complains about a P2 dependency so I think an unvetted item has somehow included in the Eclipse-Release: [eclipse-m2e/m2e-core#1439 (comment)](https://github.com/eclipse-m2e/m2e-core/pull/1439#issuecomment-1600532860)

Thanks! Looks like we were lucky that the license is approved. The license has to be approved at the release time. That we got the confirmation after the release was risky but I think it does not change the outcome. Of course we should not do that on purpose.

PR #1175, opened 2023-06-27
Merge common

laeubi:
As discussed here: 
- https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/discussions/1021

this is an attempt to merge the "commons" repository here as it can't work standalone anyways and we can get some more validation and less repositories to look at/maintain.

PR #1176, opened 2023-06-27
Fix setup for merge of platform.common into platform.aggregator

merks:
https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/pull/1175


HannesWell:
Thanks Ed for adjusting the setup.
Wouldn't it be better to keep the `Platform Documentation` Working-Set?

merks:
It wouldn’t hurt but I saw a comment about moving the PDE and JDT stuff so I wasn’t so sure.  Would you like me to add it ? Or add it yourself?

HannesWell:
> It wouldn’t hurt but I saw a comment about moving the PDE and JDT stuff so I wasn’t so sure. 
Yes then we don't need it, but it was also mentioned that this might be more work, so it could take a little while and until then I think it would be good to keep the WS.

> Would you like me to add it ? Or add it yourself?

Now that I convinced EGit to jump over the merge, I can do it 😄

PR #1178, opened 2023-06-27
[Setup] Import Doc-Projects and restory Platform Doc Working Set

HannesWell:
As discussed in https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/pull/1176 this restores the Platform Doc Working set.
Additionally it ensures that the doc bundles are also imported.

I currently have problems to test this, because with this change many more than just the intended doc-bundles are imported into the workspace (i.e. all projects that I don't have imported through the sub-module setups directly).

@merks can you tell what I'm doing wrong?

### DIFF
--- a/oomph/Platform.setup
+++ b/oomph/Platform.setup
@@ -802,6 +802,18 @@
             name="org.eclipse.platform.setup.plain.project"/>
         <requirement
             name="org.eclipse.platform.releng.prereqs.sdk.plain.project"/>
+        <requirement
+            name="org.eclipse.jdt.doc.isv.plain.project"/>
+        <requirement
+            name="org.eclipse.jdt.doc.user.plain.project"/>
+        <requirement
+            name="org.eclipse.jdt.tips.user.plain.project"/>
+        <requirement
+            name="org.eclipse.pde.doc.user.plain.project"/>
+        <requirement
+            name="org.eclipse.platform.doc.tips.plain.project"/>
+        <requirement
+            name="org.eclipse.platform.doc.user.plain.project"/>
         <sourceLocator
             rootFolder="${github.clone.platform.releng.aggregator.location}"
             locateNestedProjects="true"/>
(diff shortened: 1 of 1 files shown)


merks (changes requested):
At first this import behavior seems wrong, but the overall targlet resolution is like this:

image

So because the root requirements of the parent project pull in the platform, pde, and jdt features, whatever those transitively require, if they can be located in aggregator's source locator, they  will be included in the resolution.  That explains why those documentation projects are imported, even if one doesn't list them:

image

So I would suggest removing the explicit requirements.  Firstly they're not needed and if these things are moved later, the requirement would be broken and need attention again.

Note that the repository predicate has a relative path  that's better than using a location (because the location parts may look like a pattern) and note too it's a pattern, so /* is not the same at /.* where the latter seems more likely what you want.

So pretty much these changes are redundant and somewhat wrong.  

I have a few minutes now to correct this, but I don't actually know how to update your PR only how to create a new one....

merks:
Not all the requirements were redundant.  I've done the changes via this PR:

https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/pull/1180

HannesWell:
> Not all the requirements were redundant. I've done the changes via this PR:\n\n#1180

Thanks Ed!

> Note that the repository predicate has a relative path that's better than using a location (because the location parts may look like a pattern) and note too it's a pattern, so /* is not the same at /.* where the latter seems more likely what you want.

Thanks, I didn't knew the former. Without that I didn't found a better solution than usinge using Location-Predicate, but yes your solution is better.

> I have a few minutes now to correct this, but I don't actually know how to update your PR only how to create a new one....

You can just push to the originating branch of a PR in a personal fork using that fork's ssh URL, If edit for maintainers is enabled:

https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/working-with-forks/allowing-changes-to-a-pull-request-branch-created-from-a-fork

PR #1182, opened 2023-06-29
Fix typo in jdt.doc.isv build paths

akurtakov:


akurtakov:
Merging and starting new I-build as this should fix some failing tests on all platforms.

PR #1183, opened 2023-06-29
[Setup] Fix incorrect Working-Set task index and assign IDs to WS-tasks

HannesWell:
After https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/pull/1180 the excludedWorkingSet of the Platform Releng Aggregator working set seem to be off by one: It was `@setupTasks.3` but probably should have been `@setupTasks.2`.
This caused errors when I opened the Platform setup.

To make this more robust against future changes I added ids to all working-set task and changed the exclusion property to use that reference as @merks advocated somewhere else in the past.

### DIFF
--- a/oomph/Platform.setup
+++ b/oomph/Platform.setup
@@ -610,7 +610,8 @@
       </targlet>
     </setupTask>
     <setupTask
-        xsi:type="setup.workingsets:WorkingSetTask">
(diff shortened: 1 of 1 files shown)


merks:
Sorry. And thanks.

HannesWell:
> Sorry. And thanks.

No problem. :)

PR #1186, opened 2023-06-29
Get rid of eclipse.platform.debug

akurtakov:
It is merged into eclipse.platform now.


akurtakov:
Local build succeeded thus I'm merging this one as without it the I-build tonight will fail for sure. I'll run I-build now too in case it spots smth obvious fast - hopefully releng JIPP is not stuck.

PR #1189, opened 2023-06-30
Force Java 17 source for javadoc profile

akurtakov:
Currently it fails with "pattern matching in instanceof is not supported in -source 11" or similar in builds.


akurtakov:
Merge directly as this repo doesn't enable javadoc profile and I'm going to verify the change via platform.ui verification build.

PR #1193, opened 2023-07-03
Remove eclipse.platform.ua repo

akurtakov:
It is merged in the platform one now.


akurtakov:
Continuation of https://github.com/eclipse-platform/eclipse.platform/pull/555

akurtakov:
This should be enough to fix the build. I'll try to look into the result in few hours but if someone else checks and the verification succeeded please merge it.

Issue #1094, opened 2023-06-01
Enable logging in Eclipse Platform and Eclipse SDK products


jonahgraham:
In 4.28 a lot of work was done to enable slf4j in #588 and related issues, leading to this entry in the the [N&N](https://www.eclipse.org/eclipse/news/4.28/platform.php#slf4j.api-version-2):

> **Migrate to SLF4J-2 from Maven-Central**
> <p>
>       Besides breaking changes in the other <code>sfl4j</code> packages than <code>org.sfl4j</code>, version 2 has also changed the way to connect the <code>slf4j.api</code> bundle to a logging-backend to use Java's <code>ServiceLoader</code> mechanism.
>       Consequently you have to make sure that the logging-backend used in your application uses is compatible with <code>slf4j.api</code> version 2.
>       Additionally a <em>OSGi Service Loader Mediator</em> is needed to permit Java's <code>ServiceLoader</code> access to the the provider bundle.
>       To fulfill this requirement Eclipse Platform ships <code>org.apache.aries.spifly.dynamic.bundle</code> by default.
>       If you assemble a final application you have to ensure that this bundle is automatically activated during startup.
>       For an Eclipse Product this can be achieved by assigning the Plug-in <code>org.apache.aries.spifly.dynamic.bundle</code> a suitable start-level in the <code>Configuration</code> section of your product.
>       </p>
> 
> 

This issue to to apply those changes to the [Eclipse Platform](https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/blob/master/eclipse.platform.releng.tychoeclipsebuilder/eclipse.platform.repository/platform.product) and [Eclipse SDK](https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/blob/master/eclipse.platform.releng.tychoeclipsebuilder/eclipse.platform.repository/platform.sdk.product) products.

This issue came out of discussions in EPP https://github.com/eclipse-packaging/packages/issues/27 and EPP will adapt the packages in the same way as platform.

I am not sure if it should be covered under the same issue, or a separate issue, but making the default logger in Eclipse log to the Eclipse .log rather than no-op logging may be best?

PR #1195, opened 2023-07-03
Add silent slf4j-simple to Platform/SDK-products and auto-start Aries SPI Fly and slf4j-simple

HannesWell:
This allows the slf4j-api bundle to find the nop binding and thus prevents error messages about missing bindings.

Fixes https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/issues/1094

@jonahgraham I'm a bit unhappy that the slf4j.nop bundle also has to be auto-started, but unfortunately spi-fly only considers bundles that are in STARTING or ACTIVE (i.e. have been started successfully).
Equinox sets bundles with `Bundle-ActivationPolicy: lazy` automatically in the STARTING state. So at least for Equinox a solution would be to add `Bundle-ActivationPolicy: lazy` to the slf4j-providers (i.e. slf4j.nop, slf4j.simple, logback, ...).
Alternativly one could ask in Apache aries to also consider bundles in resolved state, but I guess they have a reason for that behavior.

If we want a solution that redirects all slf4j-logging to an Eclipse ILog, we could at least control the starting behavior.

Btw. can you tell what the `platform.product` and `platform.sdk.product` are used for? As far as I can tell, the sdk.product is the one that is provided as the Eclipse SDK download, but what are the former two used for?

Furthermore I recently made/started a few changes in PDE regarding the product configuration, maybe this is also interesting for you:
- https://github.com/eclipse-pde/eclipse.pde/pull/642
- https://github.com/eclipse-pde/eclipse.pde/pull/643
- https://github.com/eclipse-pde/eclipse.pde/pull/644

### DIFF
--- a/eclipse.platform.releng.prereqs.sdk/eclipse-sdk-prereqs.target
+++ b/eclipse.platform.releng.prereqs.sdk/eclipse-sdk-prereqs.target
@@ -206,7 +206,7 @@
 			  <!-- slf4j-api requires 1 impl, providing a dummy one... -->
 			  <dependency>
 				  <groupId>org.slf4j</groupId>
-				  <artifactId>slf4j-nop</artifactId>
+				  <artifactId>slf4j-simple</artifactId>
 				  <version>2.0.7</version>
 				  <type>jar</type>
 			  </dependency>
--- a/eclipse.platform.releng.tychoeclipsebuilder/eclipse.platform.repository/platform.product
+++ b/eclipse.platform.releng.tychoeclipsebuilder/eclipse.platform.repository/platform.product
@@ -1,7 +1,7 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <?pde version="3.5"?>
 
-<product name="Eclipse Platform" uid="org.eclipse.platform.ide" id="org.eclipse.platform.ide" application="org.eclipse.ui.ide.workbench" version="4.29.0.qualifier" useFeatures="true" includeLaunchers="true" autoIncludeRequirements="true">
+<product name="Eclipse Platform" uid="org.eclipse.platform.ide" id="org.eclipse.platform.ide" application="org.eclipse.ui.ide.workbench" version="4.29.0.qualifier" type="features" includeLaunchers="true" autoIncludeRequirements="true">
 
    <configIni use="default">
    </configIni>
@@ -9,7 +9,7 @@
    <launcherArgs>
       <programArgs>--launcher.defaultAction openFile --launcher.appendVmargs
       </programArgs>
-      <vmArgs>-Dosgi.requiredJavaVersion=17 -Dosgi.dataAreaRequiresExplicitInit=true -Dorg.eclipse.swt.graphics.Resource.reportNonDisposed=true -Xms40m -Xmx512m --add-modules=ALL-SYSTEM
+      <vmArgs>-Dosgi.requiredJavaVersion=17 -Dosgi.dataAreaRequiresExplicitInit=true -Dorg.eclipse.swt.graphics.Resource.reportNonDisposed=true -Xms40m -Xmx512m --add-modules=ALL-SYSTEM -Dorg.slf4j.simpleLogger.defaultLogLevel=off
       </vmArgs>
       <vmArgsMac>-Xdock:icon=../Resources/Eclipse.icns -XstartOnFirstThread -Dorg.eclipse.swt.internal.carbon.smallFonts
       </vmArgsMac>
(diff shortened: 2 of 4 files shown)


jonahgraham:
> Btw. can you tell what the platform.product and platform.sdk.product are used for? As far as I can tell, the sdk.product is the one that is provided as the Eclipse SDK download, but what are the former two used for?

There are tons of product files in the repo:

```
$ find * -name *.product
eclipse.platform.releng.tychoeclipsebuilder/equinox-sdk/equinox-sdk.product
eclipse.platform.releng.tychoeclipsebuilder/eclipse-junit-tests/eclipse-junit-tests.product
eclipse.platform.releng.tychoeclipsebuilder/equinox.starterkit.product
eclipse.platform.releng.tychoeclipsebuilder/equinox.starterkit.product/EclipseRTOSGiStarterKit.product
eclipse.platform.releng.tychoeclipsebuilder/eclipse.platform.repository/rcp.product
eclipse.platform.releng.tychoeclipsebuilder/eclipse.platform.repository/rcp.sdk.product
eclipse.platform.releng.tychoeclipsebuilder/eclipse.platform.repository/platform.product
eclipse.platform.releng.tychoeclipsebuilder/eclipse.platform.repository/platform.sdk.product
eclipse.platform.releng.tychoeclipsebuilder/eclipse.platform.repository/sdk.product
eclipse.platform.releng.tychoeclipsebuilder/rcp.config/config.product
```

I don't know what they are all used for, nor if they are still in use or perhaps being blindly updated every version. But I do know that Eclipse project publishes both SDK and Platform products (the latter being the one without SDK parts, but includes p2 IIRC) available on https://download.eclipse.org/eclipse/downloads/

As for the rest of your observations, I don't have a strong knowledge/opinion.

merks:
Yes, and those two things are used on the generated product catalogs too.

HannesWell:
> There are tons of product files in the repo:

Thanks. But I think we can leave that clean-up to another PR. Until know I can only tell that 
- org.eclipse.sdk.ide
- org.eclipse.platform.ide

are definitively used.

HannesWell:
I have now updated this PR to additionally replace `slf4j-nop` by `slf4j-simple` in the TP, which also includes slf4j-simple in the platform/sdk products, to 
Fix https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/issues/1226

By default all logging via slf4j-simple is disabled (by defining the system property `-Dorg.slf4j.simpleLogger.defaultLogLevel=off`)
But users configure sfl4j-simple as desired, by defining other system-properties in the `eclipse.ini` file, as described here:
https://www.slf4j.org/api/org/slf4j/simple/SimpleLogger.html

For example to generally log errors to the console and debug all jgit logs, one can set
```
-Dorg.slf4j.simpleLogger.defaultLogLevel=off
-Dorg.slf4j.simpleLogger.log.org.eclipse.jgit=debug
```

What do you think about that?
Also @sravanlakkimsetti.

HannesWell:
@sravanlakkimsetti, @jonahgraham, @merks any remarks on this one?
I think it would be good to land this soon so that EPP has a template and can make similar (but not the same, depending on whether logback is available) adjustments for M3.

jonahgraham (approved):
The change LGTM - I think it would be great to get this in for M3 (deadline 18 Aug) so we can test it out in EPP M3 and still have time to fix it.

jonahgraham:
> but not the same, depending on whether logback is available

Can you explain please?

HannesWell:
> The change LGTM - I think it would be great to get this in for M3 (deadline 18 Aug) so we can test it out in EPP M3 and still have time to fix it.

Thanks. I plan to submit this tomorrow, unless somebody objects.


> > but not the same, depending on whether logback is available
> 
> Can you explain please?

Sure. The m2e.logback feature supplies logback as logging backend to an installation. Its main purpose is actually to only be used in conjunction with the Maven-plugins running in the IDE as part of the m2e/Maven workspace-build, but many others seems to like to be use it too. In the future we might make this a bit more flexible from the m2e side.
So if the m2e.logback feature is installed, instead of slf4j-simple the logback-classic bundle needs to be auto-started in order to use it as the (only) logging backend.
Currently I'm contemplating to command that auto-starting using P2-instructions embedded into the m2e.logback feature, so that the products don't need to do that. But even then aries needs to be auto-started by the product.
So the changes in EPP would probably be:
- Auto start aries, probably with level-1
- If not m2e.logback is present and slf4j-simple is used auto-start that as well.
Given that I can implement the p2-instructions as planned in m2e.

HannesWell:
> > but not the same, depending on whether logback is available
> 
> Can you explain please?

I updated my answer in the previous comment. If you apply similar changes in EPP, feel free to ping me. I'll also let you know about what I achieved for m2e.

PR #1196, opened 2023-07-03
Add .project file for o.e.platform.releng.prereqs.sdk

HannesWell:
Instead of generating it in the Oomph-setup.
@merks what do you think?

### DIFF
--- a/eclipse.platform.releng.prereqs.sdk/.gitignore
+++ b/eclipse.platform.releng.prereqs.sdk/.gitignore
@@ -1,5 +1,4 @@
 /target
 /.settings
 /.classpath
-/.project
 /workspace
--- a/eclipse.platform.releng.prereqs.sdk/.project
+++ b/eclipse.platform.releng.prereqs.sdk/.project
@@ -1,0 +1,11 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<projectDescription>
+	<name>org.eclipse.platform.releng.prereqs.sdk</name>
+	<comment></comment>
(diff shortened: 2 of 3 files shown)


merks:
That would make more sense!

mickaelistria:
FWIW, I personally have the whole source tree imported in the workspace and don't see any benefit in having this folder open as a project over just a folder. What does that bring?

merks:
Do you see any harm?  Oomph sets it up this way to ensure that the product itself is analyzed during targlet resolution (including its dependencies), and indirectly as a result, so that the target platform also contains this product.  But I'm perfectly with the setup doing this, though it seems strange that such an important thing is so "second class" as to be buried deeply in some folder...

mickaelistria:
> Do you see any harm?

Not at all, I don't mind whether a .project gets added or not, I was genuinely wondering if there is some extra value I'm missing in my workflow without having a project for this folder.

iloveeclipse (approved):

iloveeclipse:
> I was genuinely wondering if there is some extra value I'm missing in my workflow without having a project for this folder

There are different workflows.
I don't import all of code in one huge project, so having .project in dedicated project folders matches exact my workflow.

HannesWell:
> FWIW, I personally have the whole source tree imported in the workspace and don't see any benefit in having this folder open as a project over just a folder. What does that bring?

Besides what has been said additionally it IMHO makes the target file more visible and therefore easier to find if one wants to access it.

Issue #1203, opened 2023-07-05
Build SDK using latest build of ecj


iloveeclipse:
This is a follow up on https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/issues/1184 and also an item that we've tried to address before but failed (can't find the ticket).

*AS IS*:
- JDT **compiler** ([ecj](https://github.com/eclipse-jdt/eclipse.jdt.core/tree/master/org.eclipse.jdt.core.compiler.batch)) is an essential part of the SDK (even those who don't use Java directly use it inside the help application).
- We continuously change compiler with every release, adding both features / bug fixes.
- We have lot of compiler regression tests.
- **But** we update compiler used to build SDK **only after** producing milestone builds => usually 4 times during release cycle (3 months). The update is a manual process.
- We see that real life use of compiler (building SDK itself) may [discover corner cases](https://github.com/eclipse-jdt/eclipse.jdt.core/issues/1201) not covered by regression tests.

*TO BE*:

- We should change SDK build to build [ecj](https://github.com/eclipse-jdt/eclipse.jdt.core/tree/master/org.eclipse.jdt.core.compiler.batch) **first**, before starting "main" SDK build. [The org.eclipse.jdt.core.compiler.batch project](https://github.com/eclipse-jdt/eclipse.jdt.core/tree/master/org.eclipse.jdt.core.compiler.batch) is self contained Java library and has no dependencies to SDK.
- After the ecj is built we should run SDK build using exact that freshly compiled ecj build.
- There will be no need to manually updating ecj used to build SDK after each SDK milestone.
- We will catch compiler bugs earlier.
- We might see some "unstable" builds because the produced bytecode might change, but that shouldn't happen very often. At the end that would be less work because it would be not a "big bang" after each milestone like we saw in https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/issues/1184.

This would allow us to "eat our own dog foot" with the compiler. Note: this is what Oracle does with building JDK itself every time.

*Technical notes*:
[...]

PR #1207, opened 2023-07-09
Use a local compile of ECJ in build

laeubi:
See https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/issues/1203


laeubi:
Will prefer solution from:
- https://github.com/eclipse-jdt/eclipse.jdt.core/pull/1221

PR #1209, opened 2023-07-09
Enforce qualifier update to pick-up JDT extension-point doc fixes

HannesWell:
So that the changes from https://github.com/eclipse-jdt/eclipse.jdt.debug/pull/278 are picked up to fix https://github.com/eclipse-platform/eclipse.platform/issues/561.

### DIFF
--- a/eclipse.platform.common/bundles/org.eclipse.jdt.doc.isv/forceQualifierUpdate.txt
+++ b/eclipse.platform.common/bundles/org.eclipse.jdt.doc.isv/forceQualifierUpdate.txt
@@ -1,1 +1,2 @@
 https://github.com/eclipse-platform/.github/issues/130
+https://github.com/eclipse-platform/eclipse.platform/issues/561


laeubi:
Why don't bump the version instead?

HannesWell:
> Why don't bump the version instead?

It already has been bump for the 4.29 stream.

laeubi:
> > Why don't bump the version instead?
> 
> It already has been bump for the 4.29 stream.

It is not forbidden to bump it more than once ;-)

HannesWell:
> It is not forbidden to bump it more than once ;-)

Yes, but personally I find it a bit odd, nevertheless there is no strong reason not to do it. But what's the disadvantage of using the forceQualifierUpdate.txt. I would say, that use cases like this are its purpose.

HannesWell:
Checking the build results (I replayed the build to include the jdt.doc.isv bundle) looks very promising. The leading space was gone in the html page. 🎉

PR #1215, opened 2023-07-13
Update setup to webtools/CI/3.31.0

merks:


iloveeclipse:
May be use different url, like here: https://github.com/eclipse-platform/eclipse.platform.ui/pull/937

merks:
That's more stable but less "latest":
```
<repository name="Eclipse Web Tools Update Site - latest" type="org.eclipse.equinox.internal.p2.metadata.repository.CompositeMetadataRepository" version="1.0.0">
<properties size="2">
<property name="p2.timestamp" value="1686540036000"/>
<property name="p2.compressed" value="true"/>
</properties>
<children size="2">
<child location="https://download.eclipse.org/webtools/downloads/drops/R3.30.0/R-3.30.0-20230603084739/repository/"/>
<child location="https://download.eclipse.org/webtools/downloads/drops/R3.30.0/R-3.30.0-20230603084739/repositoryunittests/"/>
</children>
</repository>
```

jukzi:
So do i understand correctly that ui's version now diverges from this. Feels wrong

akurtakov:
It's even worse - I-build uses 3.27.0 https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/blob/master/cje-production/buildproperties.txt#L58 :). I'll update that now to 3.30 too.

merks:
Based on the "eat your own dog food" principle , I try to install the latest everything:

image

I'm not sure which divergence you consider wrong?

akurtakov:
https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/pull/1217 should update the I-build to that too.

Issue #1216, opened 2023-07-13
Merge eclipse.platform.releng into this one


akurtakov:
In addition to one less repo to maintain it will make version bumps detection for help and test features being properly detected and calculated and thus reduce the need for manual touching.

akurtakov:
@sravanlakkimsetti Do you have anythings against ?

akurtakov:
My view is to keep the eclipse.platform.releng folder as we have it already but it being proper folder rather than submodule.

sravanlakkimsetti:
Please go ahead with the merge

akurtakov:
@laeubi Can you spend some time to help with this one? You're our professional in merging now :).

laeubi:
> @laeubi Can you spend some time to help with this one? You're our professional in merging now :).

So you just want that [eclipse.platform.releng](https://github.com/eclipse-platform/eclipse.platform.releng/tree/c861ab478210504182362ddced2d77155f09c518) is physically located here as we have done for [eclipse.platform.common](https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/tree/master/eclipse.platform.common) already?

akurtakov:
> > @laeubi Can you spend some time to help with this one? You're our professional in merging now :).
> 
> So you just want that [eclipse.platform.releng](https://github.com/eclipse-platform/eclipse.platform.releng/tree/c861ab478210504182362ddced2d77155f09c518) is physically located here as we have done for [eclipse.platform.common](https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/tree/master/eclipse.platform.common) already?

100%

laeubi:
See
- https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/pull/1218

akurtakov:
I consider this one done. Thanks @laeubi !

PR #1218, opened 2023-07-14
Merge releng

laeubi:
See https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/issues/1216


akurtakov:
Things went wrong here:
* eclipse.platform.releng/bundles are moved into eclipse.platform.releng directly
* eclipse.platform.releng folder has no pom.xml thus breaking build

laeubi:
> Things went wrong here:
> 
>     * eclipse.platform.releng/bundles are moved into eclipse.platform.releng directly
>     * eclipse.platform.releng folder has no pom.xml thus breaking build

Should be fixed now!

akurtakov:
Thanks! Starting an I-build to verify it now.

PR #1237, opened 2023-07-30
[Setup] Add eclipse-sdk-prereqs.target to platform.releng.aggr Targlet and exclude releng.aggregator sub-modules

HannesWell:
This PR does two things:
1. Add eclipse-sdk-prereqs.target to platform.releng.aggr Targlet
And explicitly import `eclipse.platform.releng.prereqs.sdk` to ensure the `eclipse-sdk-prereqs` target is available in the workspace when the targlet-task is executed.
2. Exclude releng.aggregator sub-modules from releng.aggr Targlet
This prevents the imports of large parts of the SDK even if their setups are not included in the workspace setup.

For example I have eclipse-platform base, ui, releng.aggregator and equinox and p2 in one workspace but not PDE and JDT (I have them in a separate one). Without the exclusion I then get many projects from PDE and JDT imported, when I also apply the releng.aggreator setup to this workspace.

### DIFF
--- a/oomph/Platform.setup
+++ b/oomph/Platform.setup
@@ -664,54 +664,24 @@
       <targlet
           name="Platform Releng Aggregator">
         <requirement
-            name="org.eclipse.platform.releng.tychoeclipsebuilder.plain.project"/>
-        <requirement
-            name="org.eclipse.platform.setup.plain.project"/>
-        <requirement
-            name="org.eclipse.platform.releng.prereqs.sdk.plain.project"/>
-        <requirement
-            name="org.eclipse.jdt.doc.isv"/>
-        <requirement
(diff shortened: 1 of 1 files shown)


merks (approved):
That seems reasonable.  This way *.target changes are visible before the next I-Build.  This doesn't end up importing projects with errors I assume.

HannesWell:
> That seems reasonable. This way *.target changes are visible before the next I-Build. 

Thanks, yes that was the main goal. :)

> This doesn't end up importing projects with errors I assume.

At least for my testing, with a fresh workspace that only imported the `Releng Aggregator` setup this worked well.
The target looks like this:
grafik

I also tested this in my existing platform WS that also contains other setups mentioned above.

PR #1239, opened 2023-08-01
Create codeQLworkflow.yml

jukzi:
https://github.com/eclipse-platform/.github/discussions/141

### DIFF
--- a/.github/workflows/codeQLworkflow.yml
+++ b/.github/workflows/codeQLworkflow.yml
@@ -1,0 +1,89 @@
+name: "CodeQLworkflow"
+on:
+  workflow_call:
+    inputs:
+      maven-goals:
+        description: maven goals to use, defaults to 'clean verify'
+        required: false
+        default: 'clean verify'
+        type: string
+      submodules:
+        description: |
+          Whether to checkout submodules: `true` to checkout submodules or `recursive` to recursively checkout submodules.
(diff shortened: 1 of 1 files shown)


HannesWell:
If the codeQL workflow would be added to the existing `mavenBuild` workflow (maybe with an option to disable/enable it), we would probably save a lot of duplication?

jukzi:
mavenBuild is a matrix of JVM versions, while codeQL is a matrix of programming languages.
mavenBuild is normally scheduled for each PR, codeQL is normally scheduled once a week or commit on master - the results are not shown on the PR but in the security tab.

laeubi:
I have ssen CodeQL runs also on PR verification, nerveless it should be an independent workflow that could then be added to the projets CI workflow like the license check.

jukzi:
by default it uses autobuild, which does not work for our repos. It needs to be build somehow. I guess it analyses the .class files. It worked with maven. If you have another working solution please do.

jukzi:
https://docs.github.com/en/code-security/code-scanning/automatically-scanning-your-code-for-vulnerabilities-and-errors/configuring-the-codeql-workflow-for-compiled-languages#autobuild-for-java--and-kotlin

laeubi:
Your inital description contains:

> CodeQL https://codeql.github.com/ can hint to **problems in sourcecode**

So I somehow assumed it analyzes the source... but if it requires a mavenbuild, maybe one can reuse the results of the regular maven build like it is done with the Junit results?

PR #1252, opened 2023-08-14
Move Aries SPIFly bundle after simpleconfigurator to start it afterwards

HannesWell:
Fixes https://github.com/eclipse-equinox/equinox/issues/302 (hopefully).


laeubi (changes requested):
I don't think this is sufficient. Relying on the implementation to write it in the order of the product (and Equinox reading and starting/install it in that order) is waiting for problems to happen. Beside this, it would not be possible to update spifly as far as I know.

Why not move spyfly to level 2 as its done with ds, event and alike?

HannesWell:
Changing the order in the configuration section didn't Change the generated config.ini.

> Why not move spyfly to level 2 as its done with ds, event and alike?

Yes that's the only releiable thing left. We have to hope that then all bundels that need the Mediator are activated in a later level.

To migitate that we should switch to a Framework-Extension in the next release, ideally one that does not need asm. :)

Since we already missed the deadline we need Project-lead or PMC approval.
@iloveeclipse, @akurtakov are you fine with this?

Please note that this does not affect any bundle.

iloveeclipse:
Is this a problem with some new feature?
Could you please provide a link to original change so one can understand what we are talking about here.

HannesWell:
> Is this a problem with some new feature?

Not directly a feature but with a new configuration necessary for slf4j-2 and IIRC now also others that use ServiceLoader.

> Could you please provide a link to original change so one can understand what we are talking about here.

Of course, it was https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/pull/1195

iloveeclipse:
Does the #1195 actually work?
I see following printed to the console if starting SDK latest with egit that uses slf4j:
```
Aug 15, 2023 1:13:57 PM org.apache.aries.spifly.BaseActivator log
INFO: Registered provider org.slf4j.simple.SimpleServiceProvider of service org.slf4j.spi.SLF4JServiceProvider in bundle slf4j.simple
SLF4J: No SLF4J providers were found.
SLF4J: Defaulting to no-operation (NOP) logger implementation
SLF4J: See https://www.slf4j.org/codes.html#noProviders for further details.
```

iloveeclipse (approved):
If that fixes the error, let's merge.

laeubi (approved):
Looks good.

HannesWell:
> Does the #1195 actually work?
> I see following printed to the console if starting SDK latest with egit that uses slf4j:
> ```
> Aug 15, 2023 1:13:57 PM org.apache.aries.spifly.BaseActivator log
> INFO: Registered provider org.slf4j.simple.SimpleServiceProvider of service org.slf4j.spi.SLF4JServiceProvider in bundle slf4j.simple
> SLF4J: No SLF4J providers were found.
> SLF4J: Defaulting to no-operation (NOP) logger implementation
> SLF4J: See https://www.slf4j.org/codes.html#noProviders for further details.
> ```

The `Registered Provider print-out` tells that at least aries was successfully started and slf4j.simple was registered.
But you shouldn't see `SLF4J: No SLF4J providers were found`.
Currently I'm on a train for a short vacation till thursday. If that remains I can look into that when back.

> (note: I've set `-Dorg.slf4j.simpleLogger.defaultLogLevel=on` in eclipse.ini for testing, but see no difference to the "off")

`On` is not a valid log-level. I have described a possible configuration and linked the slf4j-doc in https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/pull/1195#issuecomment-1654624461, which
also uses jgit as an example and that worked in my Testing.

iloveeclipse:
> I have described a possible configuration and linked the slf4j-doc in https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/pull/1195#issuecomment-1654624461, which
also uses jgit as an example and that worked in my Testing.

It would be good to add this to N&N section.

jcompagner:
> The `INFO: Registered provider org.slf4j.simple.SimpleServiceProvider of service org.slf4j.spi.SLF4JServiceProvider in bundle slf4j.simple` print-out tells that at least aries was successfully started and slf4j.simple was registered.
> But you shouldn't see `SLF4J: No SLF4J providers were found`.
> Currently I'm on a train for a short vacation till thursday. If that remains I can look into that when back.
> 

Is by accident also not slf4j 1.7 installed (so both?) I think you will see this double output

jcompagner:
.
> 
> Why not move spyfly to level 2 as its done with ds, event and alike?

We use in our product even lvl 3 already for years for this, but I guess it greatly depends on where slf4j is used in more core stuff.

What I do wonder why Aries needs asm.. not for stuf where we really use it for right?

I have a feeling that this service loader thing should really just be done in the core implementation of osgi.. (so not even needing Aries)

HannesWell:
> .
> > 
> > Why not move spyfly to level 2 as its done with ds, event and alike?
> 
> We use in our product even lvl 3 already for years for this, but I guess it greatly depends on where slf4j is used in more core stuff.
> 
The Eclipse SDK neither uses SLF4J loggers nor service loaders but the earlier it is started the better, just to be sure.


> What I do wonder why Aries needs asm.. not for stuf where we really use it for right?
> 

Aries transforms the class-files of classes that use the ServiceLoader in Order to inject special classloaders that consider the OSGi environment.

> I have a feeling that this service loader thing should really just be done in the core implementation of osgi.. (so not even needing Aries)
> 
Me too. You might be interested in https://github.com/eclipse-equinox/equinox/pull/295

HannesWell:
> > I have described a possible configuration and linked the slf4j-doc in https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/pull/1195#issuecomment-1654624461, which
> also uses jgit as an example and that worked in my Testing.
> 
> It would be good to add this to N&N section.

SLF4J-simple is not used in all packages, many use logback through m2e.logback, therefore I'm not sure that's necessary respectivly could be confusing.

iloveeclipse:
But how the users of SLF4J-simple should know about the way to configure SDK or more general: that they **can** now configure SLF4J-simple in the SDK? Or do I miss something and it was always possible but then I don't understand the change anymore?

HannesWell:
> But how the users of SLF4J-simple should know about the way to configure SDK or more general: that they **can** now configure SLF4J-simple in the SDK? Or do I miss something and it was always possible but then I don't understand the change anymore?

I don't know for sure how it was with SLF4J from Orbit was, but from a quick look at https://download.eclipse.org/eclipse/updates/4.25/R-4.25-202208311800/ there was no binding at all. Since slf4j from Maven there was only the noop Binding/Provider present. So no it was'nt possible.

For the SDK it would probably be good yes, but I hestitate a bit because in some EPP packages the logback logging backend supplied by m2e.logback is used, which supplied a logback configuration.
Altough it is actually a implementation-detail many rely on it. And if we add doc for the SDK we should tell the difference. But I don't want to make the M2E.logback config permanent.
So that should be made clear.

Issue #1197, opened 2023-07-04
Eclipse 4.29 Prerequisites: EMF


MohananRahul:
Tracking issue for EMF. M1 is scheduled for July 7th.

jukzi:
After pulling this my workspace does not work anymore - unable to get com.sun.jna:
![image](https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/assets/51790620/b88b8cc1-6d31-4a86-8f5d-2e9e45342fbd)

merks:
@HannesWell 

Since we added resolution of the actual *.target, there have been repeated such failures.  I had the same thing over the weekend and spent 20 minutes trying to clear various caches to make the problem go away.  

It's exactly like was commented elsewhere.  One can open the *.target file and it resolves but not when the targlet tries to resolve it.  I too see the same problem again now:

![image](https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/assets/208716/0a95156d-cdf3-4c00-87f1-c592bd91f871)

Maybe it's related to the fact that the *.target resource is visible twice?

![image](https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/assets/208716/3f9d7697-324d-4026-8c6f-375f30007da6)

@jukzi 

If you are desperate to get this to work, delete this:

![image](https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/assets/208716/f71cddf9-65c1-49ae-a7d1-6ca93ed3bdbc)

merks:
After I opened "both"target files, then resolution worked afterwards:

![image](https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/assets/208716/45bed998-635b-4c90-b8ff-813dbc29c641)

So this issue does appear to be related to the fact that the target is visible via to different IFiles.

merks:
I added something to the *.target and could reproduce the problem yet again.  But this time I modified the aggregator project such that it filters out the children of the eclipse.platform.releng.prereqs.sdk folder such that the *.target is visible only once in the workspace:

![image](https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/assets/208716/ffd934dd-6b5c-43e4-8a3e-c71268576eca)

Then the resolution works properly.

So the underlying problem does appear to be the two IFile occurrences of the *.target.

@akurtakov 

[...]

PR #1255, opened 2023-08-16
Filter out the workspace content of folders that are nested projects

merks:
This is particularly relevant to avoid eclipse-sdk-prereqs.target appearing twice in the workspace.

https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/issues/1197

### DIFF
--- a/.project
+++ b/.project
@@ -8,4 +8,15 @@
 	</buildSpec>
 	<natures>
 	</natures>
+	<filteredResources>
+		<filter>
+			<id>1692173149981</id>
+			<name></name>
+			<type>30</type>
+			<matcher>
+				<id>org.eclipse.ui.ide.multiFilter</id>
+				<arguments>1.0-projectRelativePath-matches-true-true-(eclipse.platform.releng.prereqs.sdk|eclipse.platform.releng.tychoeclipsebuilder|oomph)/.*</arguments>
+			</matcher>
+		</filter>
+	</filteredResources>
 </projectDescription>


HannesWell:
Thanks Ed!

Never knew that this was possible. 😅

Can you tell If the root cause is in PDE or Oomph? I would be willing to fix that.

merks:
Sorry, no, I'm not sure the root cause.   

We do look the target up by name, org.eclipse.oomph.util.pde.TargetPlatformUtil.getTargetDefinition(String), and when there are two with the same name, it's anybody's guess which of the two will win. 

In org.eclipse.oomph.targlets.internal.core.TargletContainer.ComposedTargetContent.resolve(SubMonitor, boolean) we get the target definition (by name) and call org.eclipse.pde.core.target.ITargetDefinition.resolve(IProgressMonitor) on it, so the resolution failure appears to be happening purely in PDE and there's not much else we can do via the APIs to "force it to resolve properly".

mickaelistria:
FWIW, this change is breaking my workflow (not too bad, I can workaround it, I don't mind much, but it's food for thought): I do *not* open all projects in my workspace, I have the root code tree and *some* modules I am developing or testing open as project, other remain folders. The eclipse-sdk-prereqs module is for example never open as a project (because it's not useful to me) and I'm happy that I don't have to do it.
With this change, the folder becomes fully hidden in my workspace.
The workaround is to remove the resource filter from .project.

PR #1257, opened 2023-08-20
Remove org.hamcrest.core 1.3.0

merks:
It's no longer needed nor used.


merks:
@iloveeclipse 

FYI, I'm going to remove the older version, because I think it's not used and not needed **and** removing it might get rid of the spurious org.hamcrest.core.source 1.3.0 from the p2 repository.

iloveeclipse:
Thanks.

Issue #1199, opened 2023-07-04
Eclipse 4.29 Prerequisites: Orbit


MohananRahul:
Tracking issue for Orbit. M1 is scheduled for July 7th.

jonahgraham:
No new orbit for M1

merks:
Ideally we would use bundle org.hamcrest 2.2 , which is available now but there is some use of org.hamcrest.core that needs to be cleaned up:

![image](https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/assets/208716/ebc9022f-6560-4871-81fe-d38343886f95)

HannesWell:
> Ideally we would use bundle org.hamcrest 2.2:

Yes please, that would be great.
We have to do some bnd-instruction magic in our TP to redirect org.hamcrest.core to org.hamcrest.
IIRC also org.junit requires it too, but it is probably from Orbit.
I can check it when back home.

merks:
FYI, see for related issues:

https://github.com/eclipse-orbit/orbit-simrel/issues/4#issuecomment-1650385614

merks:
@HannesWell 

Note that this makes me wonder why a consumer of a bundle must duplicate the package imports of that bundle.  Or is this a Tycho problem?  I haven't tested if PDE gets confused too, but I do recall "the package in directly required" errors in the past that always tempted me to reexport my required bundles...

HannesWell:
> Note that this makes me wonder why a consumer of a bundle must duplicate the package imports of that bundle. Or is this a Tycho problem? I haven't tested if PDE gets confused too, but I do recall "the package in directly required" errors in the past that always tempted me to reexport my required bundles...

I think the root of this problem is that problem is that the class org.junit.Assert has methods that have for example `org.hamcrest.Matcher` as argument `public static <T> void assertThat(T actual, org.hamcrest.Matcher<? super T> matcher);`
But these methods are deprecated in favor of the `MatcherAssert` class of the hamcrest bundle.

But actually when the `Assert` class is loaded the `org.junit` bundle class loader should be used and should find hamcrest.Matcher so that no NoClassDefFoundError should be thrown. So as long as one does not use Hamcrest my first guess would be that a test-bundle does not need to import it. And those that use it, it is actually ok to import the corresponding package. This is actually normal if only Import-Package is used: A bundle must list all packages it accesses, regardless of if methods from classes of other bundles return it.
[...]

PR #1261, opened 2023-08-22
Update to Orbit's rc1 milestone repository

merks:
https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/issues/1199


akurtakov:
Isn't it better to switch to the stable url now?

merks:
How do you define stable? This is the latest, and the contents will not change. Any other existing url points at what is still a moving target which is best avoided I think.

merks:
For rc2 I will spin an orbit release and then there will be a final url with immutable content that will be permanently retained.  The url in this pr is retained until m1 of the 2023-12 release cycle.

PR #1264, opened 2023-08-23
Update mockito-core to 5.5.0

merks:


merks:
@akurtakov 

I'm not sure if "we" want further dependency updates at this point?   Please close this if the answer is no.

I expect this will fail and a qualifier touch is needed, but no point in doing that if the answer is no...

merks:
Note that the license check is kind of bogus too, this was opened and closed:

https://gitlab.eclipse.org/eclipsefdn/emo-team/iplab/-/issues/10160

But then the next build opened this duplicate:

https://gitlab.eclipse.org/eclipsefdn/emo-team/iplab/-/issues/10162

Which is closed now too...

PR #1275, opened 2023-08-28
Update snakeyaml to 2.2 and bytebuddy to 1.14.7

merks:


merks:
@akurtakov 

In think these are only updates that are used in test and infrastructure, so probably still okay.  But if you don't want these, please close the PR.  

 These are the versions that will be available in the Orbit release and are currently already available here:

https://download.eclipse.org/tools/orbit/simrel/orbit-aggregation/nightly/latest

akurtakov:
Only tests are relying on these so please merge when verification build is done.

PR #1279, opened 2023-08-29
A mapping rule is needed to map com.sun.jna to net.java.dev.jna:jna

merks:


merks:
Then ensures that the mapping is correct, e.g., 

image

PR #1325, opened 2023-09-07
Move P build from 4.28 based to 4.29

jarthana:
I am not very sure about all the changes here. All I have done is to update all the build related values to +1 - for e.g., 4.28 to 4.29 and 4.29 to 4.30 and so on. Perhaps we can wait till we have the 4.29 release build and update again, but just wanted to create the PR so we can track it.


jarthana:
@sravanlakkimsetti @MohananRahul Can you please check if the changes are fine. At this point, both Y and P builds are broken.

jarthana:
I have rewritten the original changes with just the pom changes that change the versions from 4.30.0-SNAPSHOT to 4.29.0-SNAPSHOT. I think it was wrong to reference 4.30 for the patch build.

akurtakov:
@jarthana Do you need this one merged?

MohananRahul:
PBuild is available from 4_29 maintenance, so this PR is not needed. @jarthana please confirm.

jarthana:
Agree, we already have the required changes in. This is redundant now. Closing this now.

PR #1348, opened 2023-09-13
backport: Update Batik to 1.16

gireeshpunathil:
Fixes:
* CVE-2022-42890
* CVE-2022-41704

/cc @akurtakov

### DIFF
--- a/eclipse.platform.releng.prereqs.sdk/eclipse-sdk-prereqs.target
+++ b/eclipse.platform.releng.prereqs.sdk/eclipse-sdk-prereqs.target
@@ -25,14 +25,6 @@
       <unit id="javax.servlet.jsp" version="2.2.0.v201112011158"/>
       <unit id="javax.servlet.jsp.source" version="2.2.0.v201112011158"/>
       <unit id="javax.xml" version="1.3.4.v201005080400"/>
-      <unit id="org.apache.batik.css" version="1.11.0.v20190515-0436"/>
-      <unit id="org.apache.batik.util" version="1.11.0.v20190515-0436"/>
-      <unit id="org.apache.batik.i18n" version="1.11.0.v20190515-0436"/>
-      <unit id="org.apache.batik.constants" version="1.11.0.v20190515-0436"/>
-      <unit id="org.apache.batik.css.source" version="1.11.0.v20190515-0436"/>
-      <unit id="org.apache.batik.util.source" version="1.11.0.v20190515-0436"/>
-      <unit id="org.apache.batik.i18n.source" version="1.11.0.v20190515-0436"/>
-      <unit id="org.apache.batik.constants.source" version="1.11.0.v20190515-0436"/>
       <unit id="org.apache.xmlgraphics" version="2.3.0.v20190515-0436"/>
       <unit id="org.apache.xmlgraphics.source" version="2.3.0.v20190515-0436"/>
       <unit id="org.apache.commons.logging" version="1.2.0.v20180409-1502"/>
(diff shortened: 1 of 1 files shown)


akurtakov:
I can't help with backports as I can't spend any time on verifying these backports are actually fine.

gireeshpunathil:
> I can't help with backports as I can't spend any time on verifying these backports are actually fine.

fair point, understood.

/cc @MohananRahul - pls review and take it forward!

gireeshpunathil:
I have converted this to a draft for now - looks like the other IUs in `pom.xml` need to be updated as well, when we point the repository location to the latest orbit that has the batik fix.

question is - will it have side effects? for example a unit `com.ibm.icu` will now move from `64.2.0.v20190507-1337` to match the available version `72.1.0.v20221115-2007` in the latest orbit. what if there are backward incompatible changes in the bundle? the major version here is seen as bumped by 8!

merks:
ICU4j has been bumping major versions for each release.  As far as I know, none of the major version bumps over the past years have caused a problem for any downstream consumers in SimRel....

gireeshpunathil:
> ICU4j has been bumping major versions for each release. As far as I know, none of the major version bumps over the past years have caused a problem for any downstream consumers in SimRel....

thanks Ed. But it was just an example, there are other IUs that will need to be updated as well, so I am just wondering about the objective of these dependencies when their versions can be `arbitrarily` moved!

I read somewhere that orbit is less maintained and recommended way is to use these IUs from maven directly.

pro: we can stick to any version we want
con: bloated syntax in the build config

what would you advise?

merks:
Orbit is being restructured:

https://github.com/orgs/eclipse-orbit/discussions/49

Yes, if a dependency is available at Maven central as an OSGi bundle, the restructured Orbit will prefer to use that.  Orbit provides the following repository with the latest such bundle versions, already PGP signed:

https://download.eclipse.org/tools/orbit/simrel/maven-osgi/nightly/latest

Indeed projects can do that themselves directly in their *.target as the Platform itself does.  (Such projects need to set up PGP signing for such bundles.)  In addition, Orbit generates reports for SimRel projects that do this, e.g., this report for the Platform:

https://github.com/eclipse-orbit/orbit-simrel/blob/main/report/maven-osgi/platform/REPORT.md

Orbit actually analyzes the *.target files of various SimRel project to compose a union. 

Projects can choose to use Orbit for direct-from-maven dependencies or can specify such things themselves.   The syntax is not so much more verbose than for an IU dependency. 

Given that the Platform has decided to use  maven locations with maven dependencies, it definitely makes the most sense to backport it **without** changing that design.

gireeshpunathil:
thanks @merks for the detailed explanation.

for this back port, I used a workaround of splitting the IUs to two different repo locations so that it is least disturbing. Will keep the enhancements for a future PR.

@MohananRahul - I have un-drafted it and is ready for review.

MohananRahul (approved):

laeubi:
> I can't help with backports as I can't spend any time on verifying these backports are actually fine.

I really wonder if it makes sense to backport third party updates... if I want the latest and greatest why not using the latest release then?

gireeshpunathil:
> I really wonder if it makes sense to backport third party updates... if I want the latest and greatest why not using the latest release then?

we always advocate for the latest and yet, there are consumers who cannot move with the same velocity - for numerous reasons.

laeubi:
> we always advocate for the latest and yet, there are consumers who cannot move with the same velocity - for numerous reasons.

But if they can't move forward, why can they do so for third party dependencies? That seems a bit strange and even if one wants a never third party one can simply install it anyways so it seems a bit half-way here to say some things *can* be upgrade but other not, how would you decide if one consumer says it can't use newer batik while another wants the latest version?

So for me the most consistent way would be to simply keep third party dependencies as-is and only backport functional/bugfix changes.

akurtakov:
FWIW the project doesn't do any builds from maintenance branches so all backports are quite useless for people/organizations not ready to do their own builds.

gireeshpunathil:
> But if they can't move forward, why can they do so for third party dependencies? That seems a bit strange and even if one wants a never third party one can simply install it anyways so it seems a bit half-way here to say some things can be upgrade but other not, how would you decide if one consumer says it can't use newer batik while another wants the latest version?

1. if you look at the back-port patterns, you will see it is largely security fixes. these consumers have organisational mandates to stay latest in terms of security and those are non-negotiable. So they do what it takes to `upgrade` parts of their installations to meet those mandates.
2. do we get request to upgrade any other part of SDK? no
3. how would I decide if one consumer wants X and another Y version of a component? we build patches on top of these back-ports and publish those. folks can download and install those patches into the SDK. Those who are not interested, don't do anything. The unit of patching is feature, and we use the standard feature patch semantics for this.
4. further, we maintain builds from these branches, so the consumers have a pool of SDKs to baseline and then enhance.

gireeshpunathil:
> Thanks for the explanation, this sounds like a huge effort an maintenance of all those different builds/sites...

yes. we (Eclipse SDK team at IBM) are trying to absorb the gap if any in terms of promoting commercial adoption of LTS eclipse versions by balancing between the community's process (contributing to moving forward and staying latest philosophy) and our consumer's constraints (largely migration costs real effort, and don't migrate unless something breaks philosophy). We don't see an easy way to bridge these two worlds, so!

PR #1352, opened 2023-09-14
Simply gpg signing

merks:
Use the bc signer because it's much faster.
Use <skipIfJarsignedAndAnchored> to avoid a <forceSignature> list; this will PGP sign any jar that's not considered to be signed and anchored, so it handles the jce-signed jars as well as any jars with old no-longer-supported weak jar signatures.

### DIFF
--- a/eclipse.platform.releng.tychoeclipsebuilder/pom.xml
+++ b/eclipse.platform.releng.tychoeclipsebuilder/pom.xml
@@ -43,26 +43,11 @@
 					<artifactId>tycho-gpg-plugin</artifactId>
 					<version>${tycho.version}</version>
 					<configuration>
+						<signer>bc</signer>
 						<keyname>b6d3ab9bcc641282</keyname>
-						<skipIfJarsigned>true</skipIfJarsigned>
-						<forceSignature>
-							<bundle>bcpg</bundle>
-							<bundle>bcprov</bundle>
-							<bundle>com.sun.el.javax.el.source</bundle>
-							<bundle>com.sun.el.javax.el</bundle>
-							<bundle>javax.el-api.source</bundle>
-							<bundle>javax.el-api</bundle>
-							<bundle>org.apache.commons.codec.source</bundle>
-							<bundle>org.apache.commons.codec</bundle>
-							<bundle>org.apache.commons.jxpath.source</bundle>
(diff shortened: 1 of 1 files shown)


merks:
@akurtakov 

FYI, this change simplifies maintenance and improves performance.   I'm not sure the verification build does GPG signing and if so, if there is a p2 repository available for analysis after the build.  Failing that, I'll analyze the next I-Build...

mickaelistria:
Indeed, the Jenkinsfile says
```
if [[ ${BRANCH_NAME} == master ]] || [[ ${BRANCH_NAME} =~ ^R[0-9]_[0-9]+_maintenance ]]; then
						MVN_ARGS="-Peclipse-sign"
```
so signing is disabled on master.
If you wish to verify it without waiting for I-Build, you could tweak this file to enable the `eclipse-sign` profile and trigger a new build; and if successful, revert the change in jenkinsfile and merge.

merks:
@mickaelistria 

Thanks for the confirmation!.  That's what I assumed.  

What I put in the pom.xml here is **exactly** what SimRel uses for GPG signing the SimRel repository, so I'm 99.9% sure that's it's 100% correct.  😁

akurtakov:
Both of you ( @merks and @mickaelistria ) know more about the topic than me so I'm plain observer here.

HannesWell:
> If you wish to verify it without waiting for I-Build, you could tweak this file to enable the eclipse-sign profile and trigger a new build; and if successful, revert the change in jenkinsfile and merge.

You can also press the `Replay` Button in Jenkins in the first run and modify the Pipeline only for the next run 'locally'.

merks:
FYI, I ran the tests and everything looks good.  This bundle, for example, is both jar signed (JCE) and PGP signed as was the case before:

https://download.eclipse.org/oomph/archive/reports-extra/I20230915-0520/download.eclipse.org/eclipse/updates/4.30-I-builds/I20230915-0520/index/bcpg_1.76.0.html

laeubi:
Its great to see when things converge: :+1:  :1st_place_medal: 
- https://github.com/eclipse-tycho/tycho/pull/1720
- https://github.com/eclipse-tycho/tycho/pull/1726
- https://github.com/eclipse-tycho/tycho/pull/1731

PR #1356, opened 2023-09-14
Fix javadoc generation after search.core being split

akurtakov:
Continuation of
https://github.com/eclipse-platform/eclipse.platform.ui/pull/1105


iloveeclipse:
Looks like this is the fix for https://github.com/eclipse-platform/eclipse.platform/issues/682 ?

akurtakov:
Unfortunately, I'm not sure about that. We will have to wait and see.

PR #1360, opened 2023-09-16
Use o.e.platform.ide when building the platform-source zip

HannesWell:
... to replace the removed org.eclipse.platform.sdk

Additionally remove the unused
org.eclipse.platform.releng.tychoeclipsebuilder/rcp/pom.xml. From the history it looks like it was introduced as pom fro the also removed rcp.product.

Follow-up on
https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/pull/1235


merks:
@HannesWell 

I don't think this build tests what you have changed.  E.g., look at this one:

https://ci.eclipse.org/platform/job/eclipse.platform.releng.aggregator/view/change-requests/job/PR-239/

So it's probably more expedient to merge this and start and I-Build which definitely will test these changes...

HannesWell:
> I don't think this build tests what you have changed. E.g., look at this one:

Yes definitely not the change in the platformrepo.xml. But the for the removal of the rcp/pom.xml I would like confirmation that it is not referenced from some remote place. At least the direct parent does not include a `rcp` module:

https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/blob/a436490466cd3e651fa2c25bfc7910cdddd6b39e/eclipse.platform.releng.tychoeclipsebuilder/pom.xml#L30-L36

HannesWell:
Build succeeded. I'm about to kick off a new I-build.

PR #1370, opened 2023-09-18
Add jakarta.annotation / jakarta.inject 2.0 API

laeubi:
Required for
- https://github.com/eclipse-platform/eclipse.platform.ui/issues/1129


HannesWell:
There is already a location that contains the 'old' jakarta annotations, why not just add it there?

laeubi:
The location seem to be a collection of unrelated items, also they use `includeDependencyScopes="compile"` while for API i prefer to not include any dependencies. But if one wants can of course move these here as well, I just don't checked any side effects.

HannesWell:
> But if one wants can of course move these here as well, I just don't checked any side effects.

Yes that would be good.
Having the dependencies fragmented makes it just harder to manage them.

PR #1373, opened 2023-09-19
add build-helper-maven-plugin to pluginManagement section

laeubi:
There are many warnings in the platform builds in the following form:

'build.plugins.plugin.version' for
org.codehaus.mojo:build-helper-maven-plugin is missing

this is injected automatically by tycho-build-extension but as the extension can't know the desired version it has to be configured in the configurator-pom (or parent like here),


laeubi:
There is only one warning left because JDT currently overrides the buildhelper with an outdated version:

https://github.com/eclipse-jdt/eclipse.jdt.core/pull/1392

Issue #1381, opened 2023-09-21
Switch Java 20 Linux build to Java 21


iloveeclipse:
[Java 21 is released](https://www.oracle.com/java/technologies/downloads/#java21), we should switch [ep430I-unit-cen64-gtk3-java20](https://ci.eclipse.org/releng/view/Builds/job/AutomatedTests/job/ep430I-unit-cen64-gtk3-java20/) build running on Java 20 now to Java 21 (create new one and run that instead of [ep430I-unit-cen64-gtk3-java20](https://ci.eclipse.org/releng/view/Builds/job/AutomatedTests/job/ep430I-unit-cen64-gtk3-java20/)).

I have no idea if Java 21 is available in our infra, I guess a ticket is needed.

@MohananRahul : could you take care of this task?

merks:
FYI, 

I still could not find it here:

https://adoptium.net/temurin/releases/

Of course I'd like to create JustJ JREs for it, but the ones I did find are all still labeled "beta"...

iloveeclipse:
@MohananRahul : thanks for creating the job definition.

Some more questions: 
- Could you please create it on jenkins?
- Could you please check with the infra team if the Java 21 is already available on eclipse.org servers?

iloveeclipse:
@akurtakov : it is safe to run that job at any time? (I Build is running right now).

akurtakov:
Relatively so. Running jobs will not be affected but if the job calls another one (like the tests) and they are modified breakage may occur.

iloveeclipse:
OK, I will run it after the build finishes.

iloveeclipse:
> OK, I will run it after the build finishes.

Done, see https://ci.eclipse.org/releng/job/Create%20Jobs/85/

Now we have https://ci.eclipse.org/releng/job/AutomatedTests/job/ep430I-unit-cen64-gtk3-java21/.

Next: get it running, or will it automatically run on next I Build?

akurtakov:
[...]

PR #1383, opened 2023-09-21
Update publish.xml to use Java 21 build

iloveeclipse:
See https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/issues/1381


akurtakov:
Looks good. Verification build would not test this one thus merging.

PR #1387, opened 2023-09-22
Update dependencies to Orbit m1.

merks:


merks:
@akurtakov 

The verification build runs a lot faster now, but it seems all the tests are skipped.   Is that intentional?

merks:
@iloveeclipse 

Would it be a good idea to kick off another I-Build?  Or is it better to let all the tests finish for the most-recent I-Build?

PR #1399, opened 2023-09-25
[GH] Explicitly specify mvn-toolchain-id for setup JDKs to match BREEs

HannesWell:
The toolchains.xml generated by the setup-java GitHub action has '8' as 'version' value for the Java-1.8 JDK and therefore the tycho-compiler-plugin does not match it when has 'useJDK=BREE' configured for a Plugin with BREE JavaSE-1.8.

Therefore explicitly set the 'id' of each jdk-toolchain to match the corresponding BREE identifier so that the tycho-compiler-plugin can find it via this secondary way.

### DIFF
--- a/.github/workflows/mavenBuild.yml
+++ b/.github/workflows/mavenBuild.yml
@@ -56,6 +56,10 @@ jobs:
           8
           11
           17
+        mvn-toolchain-id: |
+          JavaSE-1.8
+          JavaSE-11
+          JavaSE-17
         distribution: 'temurin'
         cache: maven
     - name: Set up Maven


HannesWell:
Tested with https://github.com/eclipse-pde/eclipse.pde/pull/765.

laeubi:
@HannesWell I think it would be valid that Tycho accepts `8`  as a version for Java 1.8 as well, just in case you like to provide a PR :-)

mickaelistria:
This reminds me of https://github.com/eclipse-tycho/tycho/issues/2670#issuecomment-1655497422 which has caused trouble when setting up GitHub action build for JDT earlier this year. The trick was to override the toolchain id to force Tycho to detect it: https://github.com/eclipse-jdt/eclipse.jdt.core/blob/master/.github/workflows/ci.yml#L25

PR #1406, opened 2023-09-28
Use org.glassfish:jakarta.el instead of org.glassfish:javax.el

merks:
https://github.com/eclipse-platform/.github/issues/155


merks:
@akurtakov 

I'm quite sure this is a drop-in replacement; it even has the same bundle symbolic name. 

I assume merging must wait until M1 is fully declared, right?

akurtakov:
Yes, it's too late for M1

akurtakov:
M1 announced. Merging.

merks:
Thank you for following up. 👍

PR #1416, opened 2023-10-04
Touch to fix comparator errors

akurtakov:
Caused by new launcher binaries
eclipse-equinox/equinox.binaries@e186a1b


akurtakov:
This is what https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/pull/1414 should have been.

Issue #1384, opened 2023-09-21
Update Jetty to version 12


akurtakov:

akurtakov:
https://github.com/eclipse-pde/eclipse.pde/issues/779 was identified as issue in Jetty itself https://github.com/eclipse/jetty.project/issues/10665

PR #1418, opened 2023-10-04
Fix platform javadoc build with Jetty 12

akurtakov:
Tracked in https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/issues/1384


akurtakov:
Merge as build will not verify the javadoc. I will launch extra I-build for this one.

PR #1422, opened 2023-10-06
Publish platform features to Maven central

### LINKED ISSUE #1421: Publish Eclipse features also to Maven central
AFAICS we currently do not publish our features to Maven central. https://central.sonatype.com/search?q=org.eclipse.e4.rcp

A target platform can include features from Maven, see https://github.com/eclipse-m2e/m2e-core/pull/636 therefore it would be nice if we publish also our features to Maven so that a target platform can avoid using p2 update sites.

@merks IIRC you are handling the release to Maven? I hope that publishing the feature to Maven is a simple task, e.g., by just adding the artifacts to the publishing script.

vogella:
Feature can be used in a target platform to retrieve the plug-ins listed in them, see eclipse-m2e/m2e-core#636

Publishing these features would allow a user to use these features to define their set of available platform plug-ins based on pure Maven targets.

Fixes #1421

### DIFF
--- a/eclipse.platform.releng/publish-to-maven-central/CBIaggregator.sh
+++ b/eclipse.platform.releng/publish-to-maven-central/CBIaggregator.sh
@@ -1,6 +1,6 @@
 #!/bin/bash
 #*******************************************************************************
-# Copyright (c) 2016, 2018 GK Software SE and others.
+# Copyright (c) 2016, 2023 GK Software SE and others.
 #
 # This program and the accompanying materials
 # are made available under the terms of the Eclipse Public License 2.0
@@ -116,13 +116,6 @@ fi
 #==== remove the p2 repository (not logged): ====
 
 /bin/rm -rf p2.index p2.packed content.jar artifacts.jar
-
-#==== remove features: ====
-
-echo "== Features: ==" | tee >> .logs/removed.txt
-
-/usr/bin/find * -type d -name \*feature.group -print -exec /bin/rm -rf {} \; -prune >> .logs/removed.txt
-/usr/bin/find * -type d -name \*feature.jar -print -exec /bin/rm -rf {} \; -prune >> .logs/removed.txt
 
 #==== remove eclipse test plug-ins: ====
 


merks:
Note that no verification build will actually test that this works.  Also note that should you choose to change this to suit your personal wishes and goals, then you will be assuming responsibility for maintaining and supporting it in the future, including and especially when this fails to publish to maven central after the release.

vogella:
> to suit your personal wishes and goals

I would not say that I have personal wishes to see features published at Maven central. But RCP clients are moving into the direction of using Maven based target platforms and listing all plug-ins into a target is time consuming and error-prone especially as the  the required set of plug-ins is changing. 

The support of Feature dependencies in the target files was added by @laeubi and @ptziegler. 

Of course it is fine to revert this change if publishing to Maven fails with feature jar files.

merks:
It would seem better to support BOMs, i.e., something Maven actually understands.   But that's my personal my opinion.  

Please note, if it's not clear already, that actual publishing to Maven happens after the release and so any failure to follow all the nexus publishing rules will happen at the most inopportune moment possible.  

Please ensure that you are able to and prepared to take full responsibility for changes to the current process, especially as it will have an impact in December.

vogella:
@merks do we have a good estimation when the Maven publish is planned to happen so that I can plan to reserve time to look into potential issues? I have a big client who is very interested in using the features from Maven central.

merks:
> @merks do we have a good estimation when the Maven publish is planned to happen so that I can plan to reserve time to look into potential issues? I have a big client who is very interested in using the features from Maven central.

Please read this carefully:

https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/issues/1421#issuecomment-1820528327

Your big client, or you, will need to make a significant investment in satisfying these interests.

akurtakov:
Do you plan working on this one? If not maybe we should close it.

akurtakov:
Lars, now is the right time to work on this one for June release unless you gave up on this one.

vogella:
I gave up on this one

Issue #1419, opened 2023-10-05
SourceTests.testBug206679_247198 fails


iloveeclipse:
Side effect from https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/issues/1384.
After merging of https://github.com/eclipse-equinox/equinox/pull/317 `SourceTests.testBug206679_247198`started to fail in I Builds.

I can see it locally too. This seem to be different from https://github.com/eclipse-pde/eclipse.pde/issues/779 because PDE complains about unsatisfied imports from `org.eclipse.jetty.*` packages in the `org.eclipse.help` feature, that should be there.

```
/data/git/eclipse.pde.ui/build/org.eclipse.pde.build/scripts/genericTargets.xml:112: Processing inclusion from feature org.eclipse.help: Bundle org.eclipse.equinox.http.jetty_3.9.0.v20231004-0455 failed to resolve.:
	Unsatisfied import package org.eclipse.jetty.http_[12.0.1,13.0.0).
	Unsatisfied import package org.eclipse.jetty.server_[12.0.1,13.0.0).
	Unsatisfied import package org.eclipse.jetty.server.handler_[12.0.1,13.0.0).
	Unsatisfied import package org.eclipse.jetty.session_[12.0.1,13.0.0).
	Unsatisfied import package org.eclipse.jetty.ee8.servlet_[12.0.1,13.0.0).
	Unsatisfied import package org.eclipse.jetty.ee8.nested_[12.0.1,13.0.0).
	Unsatisfied import package org.eclipse.jetty.util_[12.0.1,13.0.0).
	Unsatisfied import package org.eclipse.jetty.util.component_[12.0.1,13.0.0).
	Unsatisfied import package org.eclipse.jetty.util.ssl_[12.0.1,13.0.0).
	Unsatisfied import package org.eclipse.jetty.util.thread_[12.0.1,13.0.0).
```

Looking on the feature I see outdated bundles included:
![image](https://github.com/eclipse-pde/eclipse.pde/assets/964108/16abccdb-c1d5-4b3b-8ba2-293a14a98cb7)

I assume the only thing is missing is a "bump" in  `org.eclipse.help` feature so it "includes" new `org.eclipse.equinox.http.jetty` version.

I will push a PR.

akurtakov:
IMHO this is https://github.com/eclipse-pde/eclipse.pde/issues/779 and not touching features.

iloveeclipse:
[...]

PR #1423, opened 2023-10-06
2nd attempt to touch org.eclipse.help

iloveeclipse:
Needed to "include" latest org.eclipse.equinox.http.jetty, second try...

Fixes https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/issues/1419


akurtakov:
These touches will not help. Jetty 12.0.2 is due tomorrow and the problem should be fixed by such an update as already explained in the issue.

iloveeclipse:
> the problem should be fixed by such an update as already explained in the issue.

I don't think it is same issue, but we can wait if jetty update helps.

akurtakov:
Is this one still needed?

iloveeclipse:
> Is this one still needed?

Nope.

PR #1427, opened 2023-10-06
Move common build arguments to maven.config and remove unnecessary

HannesWell:
The property 'project.build.sourceEncoding' is already defined in the eclipse-platform-parent/pom.xml used as parent for all modules, thus it is not necessary to explicitly add it in the Jenkinsfile.

Moving the definition of the 'aggregatorBuild' property to the maven.config simplifies local builds because then the property does not have to be declared.

### DIFF
--- a/.mvn/maven.config
+++ b/.mvn/maven.config
@@ -1,0 +1,1 @@
+-DaggregatorBuild=true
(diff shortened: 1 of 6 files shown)


HannesWell:
@akurtakov what do you think?

akurtakov:
Looks good to me. It's not once or twice that issues haven't been called due to aggregatorBuild not being used locally and thus extra resolving from I-builds repo.

PR #1428, opened 2023-10-08
Update commons-io to 2.14.0.

merks:


merks:
As I suspected, because the orbit maven-osgi build links to these same iplab issues, someone bypassed the license check when updating bnd to 7.0.0, and because we live in a glass house, we know who that was.  🙈

HannesWell:
> As I suspected, because the orbit maven-osgi build links to these same iplab issues, someone bypassed the license check when updating bnd to 7.0.0, and because we live in a glass house, we know who that was.  🙈 

It's strange that the license check fails now since it passend in https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/pull/1426:
https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/actions/runs/6436917495/job/17481207068


It also passed in M2E after only a review of bnd.lib.utils was necessary:
https://github.com/eclipse-m2e/m2e-core/pull/1570

merks:
I've had this experience as well because of course I'm constantly doing this with all the automatic updates in orbit.  I'll run a build, and it will open a bunch of issues (though sometimes the build fails completely because of "too many requests", but that's another issue).  Then a subsequent build **with no changes** opens more issues.  So it's not 100% reliable that a build will find and open all the issues needed...

HannesWell:
I have no clue. Maybe @waynebeaton has an idea why this happens?

PR #1429, opened 2023-10-09
Update WTP site for 3.32.0

merks:


vogella:
What is WTP used for? Maybe we could use wild web developer (cc @mickaelistria); AFAIK WWD is in a better state.

akurtakov:
Actually WTP is not used but the helper releng tools are used org.eclipse.wtp.releng.tools.feature . Unfortunately, I don't remember what exactly are these used for.

zulus:
We use this also in PDT, these helper allow us to modify p2 metadata. Mirrors url (in theory supported by tycho now) and stats url

laeubi:
> We use this also in PDT, these helper allow us to modify p2 metadata. Mirrors url (in theory supported by tycho now) and stats url

If anything is needed from Tycho and one likes to contribute that feel free to open a PR, if it is important for platform as well I can even consider adding support for it.

akurtakov:
Quite often the problem is not functionality is not in Tycho but that old way of doing things stays as no one has the time to redo it in the new way. 
P.S. I'm not saying this is or isn't the case here but I had to do exactly this many many times.

zulus:
I've checked

1. Hardcoded WTP repository (from 2010) that remove "Uncategorized" category, in theory no longer necessary: `eclipse.platform.releng.tychoeclipsebuilder/eclipse/buildScripts/process-artifacts.xml`
2. `org.eclipse.wtp.releng.tools.addRepoProperties` as in PDT, it's setup MIRRORURL, p2ArtifactRepositoryName, p2MetadataRepositoryName, see `cje-production/promotion/promoteSites.sh`

As far as I remember, tycho long time ago, always ignored properties from category.xml, I don't know how it looks now

laeubi:
@zulus thanks for investigation, I'm not sure I completely understand (1) so you like to remove the category from the generated updatesite so it is not shown to the user? Why do we need to add uncategorized items to the updatesite at first place? Maybe with modern Tycho we can do better here at the source.

For the second: Since Tycho 4 it is much easier / flexible to add support I think the best would be to provide an [integration-test](https://github.com/eclipse/tycho/wiki#providing-an-integration-test) to demonstrate such property usage in `category.xml`, if it does not work we can quite easy add support for it!

zulus:
> @zulus thanks for investigation, I'm not sure I completely understand (1) so you like to remove the category from the generated updatesite so it is not shown to the user? Why do we need to add uncategorized items to the updatesite at first place? Maybe with modern Tycho we can do better here at the source.

This was workaround for old bug, already fixed, so could be dropped.

> For the second: Since Tycho 4 it is much easier / flexible to add support I think the best would be to provide an [integration-test](https://github.com/eclipse/tycho/wiki#providing-an-integration-test) to demonstrate such property usage in `category.xml`, if it does not work we can quite easy add support for it!

In theory Tycho allow setup such properties directly in `tycho-p2-repository-plugin` via `extraArtifactRepositoryProperties` and `repositoryName` configuration. The problem is that you have to know target repository URL during build. It's not a problem in smaller projects like PDT, but platform have "promotion" procedure and need to setup these url's later, on already generated repository.

vogella:
> This was workaround for old bug, already fixed, so could be dropped.

Thanks @zulus can you send a PR to remove this? You already looked into it, so I assume it is easy for you to get rid of this.

PR #1436, opened 2023-10-12
Add jakarta.inject-api / jakarta.annotation-api to isv build

laeubi:


merks:
@laeubi 

FYI, ignore the license problem.  It's unrelated to your change.

laeubi:
CI is fine so lets go for it...

akurtakov:
This is still not fixed as the the same symbolic name jakarta.annotation-api is used for both versions which confuses the handcrafted script.

laeubi:
> This is still not fixed as the the same symbolic name jakarta.annotation-api is used for both versions which confuses the handcrafted script.

Do you mean Tycho extra requirements or can you reference the script in question?

laeubi:
@akurtakov another question: Can we somehow see this error in this build here as well?

mickaelistria:
@laeubi do you mean logs from https://download.eclipse.org/eclipse/downloads/drops4/I20231012-1800/logs.php#javadoc ?

akurtakov:
I haven't looked enough into how can we make the jenkins build fail if javadoc build fails if this is the question.

laeubi:
I have now created:
- https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/pull/1439

laeubi:
@mickaelistria I mean what ever fails and needs fixing here :-)

PR #1437, opened 2023-10-12
Use Tycho 4.0.4-SNAPSHOT

laeubi:
Tycho has some enhancements especially at API tools and transport handling (compression), that I would like to release soon, to perform some more confidence check I'd like update platform using the current SNAPSHOT until end of October to see if we have any issues that need to be fixed before the release.


akurtakov (approved):
Code looks good. If verification build succeeds please merge.

PR #1439, opened 2023-10-13
Use version ranges to include versions jakarta.inject.jakarta.inject-api

laeubi:
@akurtakov @mickaelistria WDYT?

### DIFF
--- a/eclipse.platform.common/bundles/org.eclipse.platform.doc.isv/pom.xml
+++ b/eclipse.platform.common/bundles/org.eclipse.platform.doc.isv/pom.xml
@@ -89,7 +89,7 @@
                   <requirement>
                     <type>eclipse-plugin</type>
                     <id>jakarta.inject.jakarta.inject-api</id>
-                    <versionRange>0.0.0</versionRange>
+                    <versionRange>[1,2)</versionRange>
                   </requirement>
                   <requirement>
                     <type>eclipse-plugin</type>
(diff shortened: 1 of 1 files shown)


laeubi:
Alright I'll try to see if I find a solution here...

akurtakov:
The way javadoc is produced is so error prone that it really costs multiple workdays per release to handhold it. It would be nice to get rid of the ant part and make it pure maven. Can this be covered by the IDE WG releng work ticker, jonahgraham ?

akurtakov:
So the issue seems to come from maven-dependency-plugin first as running `mvn clean dependency:copy-dependencies` old versions are not copied in target/dependencies. This is due to Maven resolving to *single* version.

jonahgraham:
> Can this be covered by the IDE WG releng work ticket, @jonahgraham ?

I think so - @laeubi should talk to Paul Buck at EF if in doubt. My hope is that if it fits within the existing budget on the SoW then Paul says it is fine. Please feel free to cc in any email to Paul if I can help.

akurtakov:
I'm closing this one in favor of https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/pull/1449 which hardcodes versions. 
@laeubi If you can work on this one to make it *pure* maven build without all these hacks please take it in another issue.

laeubi:
@akurtakov @jonahgraham I have now started with an initiative to get rid of more ant-magic here, of course it is only a first step and likely need enhancements and more investigations:

- https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/pull/1489
- https://github.com/eclipse-tycho/tycho/pull/2969

PR #1449, opened 2023-10-17
Fix javadoc build in platform.doc.isv

akurtakov:
Hardcodes the old annotation/inject jars (with versions) in order to make it build. Ugly but can't find another way to fix it for M2 - as they are the older versions not much churn should be there. Also removes some unused build machinery to not confuse more.


akurtakov:
Verification build will not actually test it thus I'm pushing and will start I-build.

PR #1459, opened 2023-10-22
Move more product configuration from p2.inf to product-file

HannesWell:
And remove the unnecessary specification of the property `org.eclipse.equinox.p2.type.product=true`, which Tycho/P2 already set by
default.
Additionally unify max-memory launcher-arguments of Platform/SDK products

P2 and Tycho 4.0.4 already support names for Product Update repositories with
- https://github.com/eclipse-equinox/p2/pull/353
- https://github.com/eclipse-tycho/tycho/pull/2929

but I think it is good to not submit this before PDE has Editor support for that as well as described in https://github.com/eclipse-equinox/p2/issues/345.

A similar simplification can then also be applied to all EPP packages.

### DIFF
--- a/eclipse.platform.releng.tychoeclipsebuilder/eclipse.platform.repository/platform.product
+++ b/eclipse.platform.releng.tychoeclipsebuilder/eclipse.platform.repository/platform.product
@@ -194,6 +194,13 @@ United States, other countries, or both.
       <property name="org.eclipse.update.reconcile" value="false" />
       <property name="eclipse.buildId" value="${unqualifiedVersion}.${buildQualifier}" />
       <property name="osgi.bundles.defaultStartLevel" value="4" />
+      <property name="osgi.instance.area.default" value="@user.home/workspace" />
+      <property name="osgi.instance.area.default" value="@user.home/Documents/workspace" os="macosx" />
    </configurations>
+
+   <repositories>
+      <repository location="https://download.eclipse.org/eclipse/updates/4.30" name="The Eclipse Project Updates" enabled="true" />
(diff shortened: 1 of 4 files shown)


akurtakov:
This looks very promising !

eclipse.platform.releng.tychoeclipsebuilder/eclipse.platform.repository/platform.product
```diff
@@ -194,6 +194,13 @@ United States, other countries, or both.
       <property name="org.eclipse.update.reconcile" value="false" />
       <property name="eclipse.buildId" value="${unqualifiedVersion}.${buildQualifier}" />
       <property name="osgi.bundles.defaultStartLevel" value="4" />
+      <property name="osgi.instance.area.default" value="@user.home/workspace" />
```
mickaelistria, line 197: IIRC, EPP uses `eclipse-workspace` here. Should we also change it for SDK products?

HannesWell: Do you mean that the default suggestion for the workspace should be co-located to the installation?
Yes that's something I would also support. I always annoys me if I download an SDK to test something out that I first have to define the test workspace to be somewhere else than in such prominent location.

I'll create a follow-up PR on this so that we can discuss that then.

merks: No, I don't think he means that.  Me means it's specified like this:
```
-Dosgi.instance.area.default=@user.home/eclipse-workspace
```
I think co-locating it with the installation you have to be careful it's not a read-only installation; maybe you'd co-located it relative to the installation surrogate but might even end up with a read-only location...

Issue #1361, opened 2023-09-16
Are there bundles included by features that are unused?


merks:
I did an analysis of the Platform repository, excluding all features, to see if there are IUs that are not strictly required by anything, i.e., ones with no children in the view:

![image](https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/assets/208716/377488a1-e98d-4e10-93fe-24d6834b2850)

I suspect these can be removed...

There are also these, but I think these are not present for use at runtime but rather are needed at development time, so that's okay:

![image](https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/assets/208716/5a0adc73-3ae2-4343-b2d8-04dc02b28b00)

HannesWell:
Sounds good to me.

> There are also these, but I think these are not present for use at runtime but rather are needed at development time, so that's okay:
> 
> ![image](https://user-images.githubusercontent.com/208716/268436888-5a0adc73-3ae2-4343-b2d8-04dc02b28b00.png)


Yes they are used by PDE:
https://github.com/eclipse-pde/eclipse.pde/blob/b01e2c4bffc42b5aecfe6479fffd6dd9c2495f8e/ui/org.eclipse.pde.core/src/org/eclipse/pde/internal/core/annotations/OSGiAnnotationsClasspathContributor.java#L44-L46

akurtakov:
Is there something planned here?

merks:
I think these can be removed and the appear to be present purely because features include them.

![image](https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/assets/208716/5a28462d-8228-4005-81a5-26abb1968c10)

![image](https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/assets/208716/328e592b-976b-4963-b131-8d85d7253804)

![image](https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/assets/208716/65881ee5-8504-4f8b-a359-b1ad65944114)

![image](https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/assets/208716/1777863a-fa58-4900-876e-f1380ba8c450)

So we could plan and try to remove them, though one is always concerned that something might stop working for unexpected reasons...

Looking at a one of these, it does appear that the exported packages that are _potentially_ used are provided by alternatives, e.g., mortbay:

[...]

PR #1470, opened 2023-10-24
Add jakarta.servlet-api, javax.servlet.jsp-api, javax.el-api to isv.

merks:
The javadoc build seems to need them.

https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/issues/1361

### DIFF
--- a/eclipse.platform.common/bundles/org.eclipse.platform.doc.isv/platformOptions.txt
+++ b/eclipse.platform.common/bundles/org.eclipse.platform.doc.isv/platformOptions.txt
@@ -137,7 +137,10 @@
 ;${dependency.dir}/icu4j_*.jar
 ;${dependency.dir}/com.jcraft.jsch_*.jar
 ;${dependency.dir}/jakarta.annotation-api_*.jar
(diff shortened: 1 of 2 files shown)


merks:
@iloveeclipse 

I've already merged this.  Shall I kick of yet another I-Build?

iloveeclipse:
> I've already merged this. Shall I kick of yet another I-Build?

Yes, please.

merks:
In progress:

https://ci.eclipse.org/releng/job/Builds/job/I-build-4.30/83/

iloveeclipse:
Thanks, that solved javadoc compilation errors.

merks:
Thank you for the confirmation!  This is also a good reminder that garbage cleaning the target platform is not a trivial task...

akurtakov:
I'm a bit late here but there must be smth else in the target platform that provides these apis so code compiles which could have been added to the scripts/config instead. Not that it matters as the way they are used here is just at generation time so whatever works but I had to share that for if/when someone faces similar problem in the future.

Issue #1463, opened 2023-10-24
Default Eclipse JVM should not have -XmX parameter


grandinj:
I have noticed that recent installers, when using the default supplied VM to run Eclipse, has an -XmX parameter.

This leads to an extremely frustrating experience with large projects. I suddenly found my debugger doing all sorts of very weird things, which led to a loooooooong goose chase before I found the cause.

Either
(a) the platform needs some indication that it is running into memory issues
(b) just leave that parameter out, in which case the VM will grow to use available memory on the machine.

(I vote for option b)

Thanks.

iloveeclipse:
I believe option (b) makes sense. On "usual" working environments today we have way more RAM as before and JVM would "automatically" reserve more than default 2GB (SDK sets `-Xms256m -Xmx2048m` in the sdk.product). 

The only problem I see "smaller" configurations like Jenkins platform builds that usually suffer from too strict resources limits and where JVM would set smaller heap. 
But there we could set VM heap limits in Jenkins files.

merks:
FYI, this issue was discussed here as well:

https://github.com/eclipse-packaging/packages/issues/37

Most folks are using one of the EPP packages, so what the Platform does has much narrower impact than what EPP does...

jonahgraham:
As I mentioned in https://github.com/eclipse-packaging/packages/issues/37#issuecomment-1656967319 the EPP normally aims to follow where the Platform/SDK leads. This has been the case because we used to inherit the Platform settings, that was changed [recently(ish)](https://github.com/eclipse-packaging/packages/commit/a6b90b060b8c3684596e691254bf2f74a559b8aa#diff-16e6384c25129c8b79f16290a9820ce1f15fbbce94fddcd6f14d39996078ceb5) which gives us the option to drop the xmx from EPP.

In this case perhaps EPP should lead and platform/SDK can follow along later?

iloveeclipse:
> In this case perhaps EPP should lead and platform/SDK can follow along later?

Or we can just try to do this in SDK and see if something breaks on Jenkins/official build & fix Jenkins files etc in case we still need Xmx option. I will push a PR in a moment.

iloveeclipse:
> Or we can just try to do this in SDK and see if something breaks on Jenkins/official build & fix Jenkins files etc in case we still need Xmx option.

Any objections? => https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/pull/1471

[...]

PR #1471, opened 2023-10-24
Don't specify JVM memory limits in eclipse.ini

iloveeclipse:
Fixes https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/issues/1463


HannesWell (approved):
Sounds reasonable for me, lets try this.

PR #1484, opened 2023-10-28
Use workspace co-located to the installation by default for SDK-products

HannesWell:
As suggested in https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/pull/1459#discussion_r1368244436, this is a suggestion to use as default for the workspace a directory colocated to the installation instead of `${user.home}/workspace`.

What do you others think?

We use that configuration for our products and it works well on Windows and Linux but since we don't provide Mac build I cannot tell how that setting works there.

### DIFF
--- a/eclipse.platform.releng.tychoeclipsebuilder/eclipse.platform.repository/platform.product
+++ b/eclipse.platform.releng.tychoeclipsebuilder/eclipse.platform.repository/platform.product
@@ -194,8 +194,7 @@ United States, other countries, or both.
       <property name="org.eclipse.update.reconcile" value="false" />
       <property name="eclipse.buildId" value="${unqualifiedVersion}.${buildQualifier}" />
       <property name="osgi.bundles.defaultStartLevel" value="4" />
-      <property name="osgi.instance.area.default" value="@user.home/workspace" />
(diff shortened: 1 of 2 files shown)


laeubi:
> What do you others think?

I think that the Workspace does not belongs to the installation at all, I have wiped a lot of eclipse installs over time but never wanted to loose my workspace.

iloveeclipse:
${user.home}/workspace is writable by default. Location next to install location can be read only if install was done by the admin user. e.g. All rpm based installations are done with root rights on root owned directories. No one regular user will be able to create workspace there.

So please let the default be as it was, it was chosen for a good reason.

HannesWell:
> Do you know what the impact this will have? I.e., will the workspace prompt dialog just be empty rather than coming up with a default? 

No, the prompt will not be empty, when launching the Product the first time after e.g. after unzipping it, the dialog suggest a folder named `workspace` in the installation folder (in this case extracted the zip in my Downloads folder):

grafik



> Is that actually a good thing? How does that correspond to the title of the PR because for that taken literally, I have this concern:
> 
> [#1459 (comment)](https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/pull/1459#discussion_r1375243100)

That's of course the question, but looking at your and the other answers it looks like for most the answer is no.
So I think this proposal does not make it -> Closing.

PR #1485, opened 2023-10-29
Add screenshots of failed tests to Platform UI test artifacts

basilevs:
Platform UI has unstable tests that are hard to debug and platform specific. They save screenshots but build system does not archive anything but Junit reports in XML format. As screenshots are essential for failure analysis and test stabilization this change adds them to test artifacts.

See https://github.com/eclipse-platform/eclipse.platform.ui/pull/1253#issuecomment-1783909424 for an example of test failure investigation.


mickaelistria:
Thank you!

PR #1489, opened 2023-10-31
Use the new tycho-document-bundle-plugin:schema-to-html to generate doc

laeubi:
The new tycho-document-bundle-plugin:schema-to-html can replace the ant task "pde.convertSchemaToHTML" even if we still need a quite verbose configuration this already has two benefits:

- configuration is located at maven so much more visible no need to look around
- the mojo has access to the full build and dependency information and can be enhanced in the future to remove the need for manual configuration


laeubi:
I think one quite obvious enhancement to the process is that we simply add all reactor projects to the search path I can't think of a situation where I do **not** want to find a reference and give me an error instead (@akurtakov error are currently only printed to console but never failed the build, wdt should we make it an option to fail or simply fail on any error or leave it as is?).

Second is, currently in the build.xml there are some copy tasks left:
```
<!--copy extra files linked to by generated extension point doc-->
<copy file="${eclipse.platform.ua}/org.eclipse.ui.cheatsheets/schema/cheatSheetContentFileSpec.html" todir="${dest}"/>
<copy file="${eclipse.platform.ua}/org.eclipse.ui.cheatsheets/schema/cheatSheetContentFileSpec.exsd" todir="${dest}"/>
<copy file="${eclipse.platform.ua}/org.eclipse.ui.cheatsheets/schema/compositeContentFile.html" todir="${dest}"/>
<copy file="${eclipse.platform.ua}/org.eclipse.ui.cheatsheets/schema/compositeContentFile.exsd" todir="${dest}"/>
<!--copy extra HTML file linked to by generated extension point doc-->
<copy file="${eclipse.platform.ua}/org.eclipse.ui.intro/schema/introContentFileSpec.html" todir="${dest}"/>
```

some copy the exsd and html others only html ... so I wonder what is the intend here, e.g. do they originate from "referenced schemas" actually and the mojo should simply always copy referenced ones? Also include the exsd? Or is this something referenced in comments of the schema and we can't automate this? Maybe @merks as the only exsd schema expert I'm aware of :-)

laeubi:
> [ERROR] Failed to execute goal org.eclipse.tycho.extras:tycho-document-bundle-plugin:4.0.4-SNAPSHOT:build-help-index (index) on project org.eclipse.platform.doc.isv: Index destination path not specified in the extension. -> [Help 1]
07:28:22  org.apache.maven.lifecycle.LifecycleExecutionException: Failed to execute goal org.eclipse.tycho.extras:tycho-document-bundle-plugin:4.0.4-SNAPSHOT:build-help-index (index) on project org.eclipse.platform.doc.isv: Index destination path not specified in the extension.

@akurtakov need to check why removing genhtml breaks help index but I though we stopped generating help index at all?

laeubi:
About the copy it seems these htmls are already check in to version control and **not** generated .. anyone knows why?

akurtakov:
Indexes are not part of the published bundles. Maybe I didn't manage to disable generating them fully.

laeubi:
Alright I removed that now, next issue is: 

`Caused by: org.apache.maven.plugin.MojoExecutionException: Failed to create assembly: Error adding file to archive: /home/jenkins/agent/workspace/atform.releng.aggregator_PR-1489/eclipse.platform.releng.tychoeclipsebuilder/eclipse-junit-tests/../../eclipse.platform.common/bundles/org.eclipse.jdt.doc.isv/target/jdtconvert.txt`

I think I can create a dummy file but probably simply remove the inclusion of that file?

laeubi:
Build is green now, if no one objects I would like to merge this to see if ibuild is happy as well...

merks:
The build is still running at the moment.

laeubi:
> The build is still running at the moment.

Yes I squashed and rebased it so lets wait a bit..

akurtakov:
Build succeeded. I'm pushing and starting new I-build to verify everything is fine.

PR #1492, opened 2023-10-31
Update to jetty 12.0.3

merks:


akurtakov:
I've just updated to latest I-build and help center is not starting at all for which I think this PR is the reason.

merks:
In a runtime launch with these versions:

image

The help comes up:

image

I'll try updating my install...

merks:
The real installation with jetty 12.0.3 installed also works fine:

image

Am I testing the right thing?

merks:
I downloaded this:

https://download.eclipse.org/eclipse/downloads/drops4/I20231031-0530/eclipse-SDK-I20231031-0530-win32-x86_64.zip

And that works too:

image

akurtakov:
I'll download again. No idea what goes wrong with my install, nor I have the time to investigate now.

PR #1495, opened 2023-11-01
Remove no longer required search path

laeubi:
With the recent Tycho changes an explicit search path should no longer be required.


akurtakov:
wouldn't same generated docs for all content be put in all doc bundles that way?

laeubi:
> wouldn't same generated docs for all content be put in all doc bundles that way?

The searchpath is only used for finding referenced schemas, the actual generated content is controlled by the `<manifest>`

akurtakov:
Ah, perfect! Sorry for the misunderstanding.

PR #1496, opened 2023-11-01
Do not check pde/jdt schema logs in build tests

akurtakov:
In case of issues build is supposed to fail now.


akurtakov:
This is continuation of https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/pull/1489

PR #1500, opened 2023-11-02
Move Javadoc config to the pluginManagement

laeubi:
Currently the javadoc plugin is configured in a profile for one explicit execution. If we want to use it on other plcaes it is better configured in the pluginManagement


akurtakov (approved):
Looks good to me.

PR #1501, opened 2023-11-02
Use maven javadoc for pde

laeubi:
Currently we use a hand crafted options file to build the javadoc files
what is cumbersome and error prone.

This aims to replace the ant calls by the maven-javadoc-plugin so that
dependencies declared in the pom are automatically passed to javadoc
build. Beside that this makes the doc used for referenced links is added
as a dependency and then unpacked to reduce the need to specify absolute
path to other projects in the reactor.


akurtakov:
I'm merging this one and starting I-build now so we can verify the docs are in good shape.

PR #1507, opened 2023-11-03
Set a bunch of project encodings to UTF-8

fedejeanne:
It was done using a "Quick Fix" and the value set (UTF-8) is the same value that any other project in the workspace uses. The objective of this commit is to declutter the "Problems" view.


kthoms (approved):

kthoms:
Thanks, @fedejeanne

fedejeanne:
> Thanks, @fedejeanne

My pleasure :)

PR #1510, opened 2023-11-04
Add deprecation notice for javax support by the E4 injector

HannesWell:

I also added a corresponding N&N entry: https://github.com/eclipse-platform/www.eclipse.org-eclipse/pull/79

### DIFF
--- a/eclipse.platform.common/bundles/org.eclipse.platform.doc.isv/porting/removals.html
+++ b/eclipse.platform.common/bundles/org.eclipse.platform.doc.isv/porting/removals.html
@@ -61,10 +61,8 @@ For further details or to provide feedback on this change, see <a href="https://
 </p>
 
 
-
-
-
 <!-- ############################################## -->
+
 <h2>API removals after June 2018</h2>
 
 <h3 id="fullscreen">1. org.eclipse.ui.cocoa.fullscreenWindow</h3>
@@ -86,8 +84,8 @@ For further details or to provide feedback on this change, see
 </p>
 
 
-
 <!-- ############################################## -->
+
 <h2>API removals after June 2020</h2>
 
 <h3 id="mpartdescriptor">1. Remove Dirtable flag from MPartDescriptor</h3>
@@ -114,6 +112,8 @@ For further details or to provide feedback on this change, see <a href="https://
 </p>
 
 
+<!-- ############################################## -->
+
 <h2>API removals after March 2021</h2>
 
 <h3 id="commandsutil">1. org.eclipse.core.commands.util package</h3>
@@ -133,8 +133,8 @@ For further details or to provide feedback on this change, see <a href="https://
 </p>
 
 
-
 <!-- ############################################## -->
+
 <h2>API removals after June 2021</h2>
 
 <h3 id="commands">1. Delete deprecated contents of org.eclipse.ui.commands package</h3>
@@ -173,7 +173,9 @@ The following deprecated contents of org.eclipse.ui.commands are not used anymor
 For further details or to provide feedback on this change, see <a href="https://bugs.eclipse.org/bugs/show_bug.cgi?id=431177" target="_blank">bug 431177</a>.
 </p>
 
+
 <!-- ############################################## -->
+
 <h2>API removals after May 2022</h2>
 
 <h3 id="sharedImages">1. Hover images from ISharedImages</h3>
@@ -197,6 +199,9 @@ Multiple images not used since Eclipse 3.0 in the platform will be deleted.
 For further details or to provide feedback on this change, see <a href="https://bugs.eclipse.org/bugs/show_bug.cgi?id=559593" target="_blank">bug 559593</a>.
 </p>
 
+
+<!-- ############################################## -->
+
 <h2>API removals after June 2022</h2>
 
 <h3 id="icu4j">1. ICU4J bundle from SDK</h3>
@@ -206,7 +211,9 @@ Support and/or usage of ICU4J is being gradually removed from internals. Whereve
 For further details or to provide feedback on this change, see <a href="https://bugs.eclipse.org/bugs/show_bug.cgi?id=562582" target="_blank">bug 562582</a>.
 </p>
 
+
 <!-- ############################################## -->
+
 <h2>API removals after September 2022</h2>
 
 <h3 id="observablefactories">1. Observable factory classes</h3>
--- a/eclipse.platform.common/bundles/org.eclipse.platform.doc.isv/porting/removals.html
+++ b/eclipse.platform.common/bundles/org.eclipse.platform.doc.isv/porting/removals.html
@@ -254,6 +254,24 @@ For further details or to provide feedback on this change, see <a href="https://
 <h1 id="removed">Removed API in previous releases</h1>
 
 
+<!-- ############################################## -->
+
+<h2>API removals after December 2025</h2>
+
+<h3 id="javax-annotations">1. Support for javax-annotation by the E4 Injector</h3>
+
+The support of the Eclipse E4 Platform Dependency Injector for annotations from the <code>javax.inject</code> and <code>javax.annotation</code> package is deprecated
+and scheduled to be disabled by default after the December 2025 release.
+The eventual removal of support for these <code>javax</code> annotations will happen at any time after the December 2025 release without any further prior notice,
+whenever maintaining it is no longer feasible or hinders other development.
+<p>
+Processing of <code>javax</code> annotations can be explicitly disabled for an application by specifying the VM property <code>-Declipse.e4.inject.javax.disabled=true</code>
+and explicitly enabled by specifying <code>-Declipse.e4.inject.javax.disabled=false</code>.
+For further details or to provide feedback on this change, see <a href="https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/issues/1056">eclipse.platform.releng.aggregator#1056</a>.
+</p>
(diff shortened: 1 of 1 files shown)


eclipse.platform.common/bundles/org.eclipse.platform.doc.isv/porting/removals.html
```diff
@@ -247,6 +254,17 @@ <h3 id="IModelProviderEvent">4. Remove IModelProviderEvent's fields TARGET_CHANG
...
+<!-- ############################################## -->
+
+<h2>API removals after December 2025</h2>
+
+<h3 id="javax-annotations">1. Support for javax-annotation by the E4 Injector</h3>
+
+The support of the Eclipse E4 Platform Dependency Injector for annotations from the <code>javax.inject</code> and <code>javax.annotation</code> package is deprecated and scheduled for removal after the December 2025 release.
+<p>
+For further details or to provide feedback on this change, see <a href="https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/issues/1056">eclipse.platform.releng.aggregator#1056</a>.
+</p>
```
akurtakov, line 266: IMHO, we will have to come up with a bit stricter plan. E.g. remove after December 2025 AND e.g. disable after June 2025 but still allow to enable it so people face it but have some (not long so they don't forget it) period being able to explicilty enable .

HannesWell: OK, but then the facilities to disable/enable should be built in and for now the default is simply to enable javax-processing. Then after June 2025 the switch is filled to disable support. This will also allow users to test their application with javax-processing disabled.

akurtakov: I agree with you and if you have the time for it now - even better. If not this "disabling" solution can be done in some of the coming releases.

laeubi: I just wanted to note that this will immediately break any older Eclipse Plugins so I'm not sure if removing support for that really brings any benefit, I can see that we might not ship it by default but the code used here is actually almost never touched so I don't by any maintenance arguments here.

So what would be the real benefit of removing javax support?

HannesWell: Yes. Created https://github.com/eclipse-platform/eclipse.platform/pull/820.
Will update this PR once that is submitted.

HannesWell: > I just wanted to note that this will immediately break any older Eclipse Plugins so I'm not sure if removing support for that really brings any benefit, I can see that we might not ship `javax.*` bundles by default but the code used here is actually almost never touched so I don't by any maintenance arguments here.

For example with https://github.com/eclipse-platform/eclipse.platform/pull/791 a whole bunch of tests was added and what I see they are not in the best shape.
Furthermore every code, even if not changed makes maintenance harder because one usually needs to understand it, even if it is not changed.

But we can of course think about extending the deprecation period. As a suggestion I changed to text to:
`scheduled to be disabled by default after the December 2026 release and scheduled for removal after the December 2028 release.`, which means three years until being disabled by default and five years until support is fully removed.
If a Plugin does not manage to apply the simple change and produce new builds I would consider it dead anyways.

As a side-note, for example since version 7 Guice only supports jakarta annotations:
https://github.com/google/guice/wiki/Guice700#jee-jakarta-transition
and Guice 6 is the last version to support javax-annotations, so they faded out support for javax already. Although I don't know if they plan to release new versions in the 6 stream.

HannesWell:
Btw. does anybody know why tonight the I-builds did not run?

merks:
Maybe the switch back to standard time overnight has changed when you should expect that build to happen by 1 hour...

merks:
BTW, I'd like the/a build to happen after I merge this so that I can use it in PDE:

https://github.com/eclipse-equinox/p2/pull/377

iloveeclipse:
> Btw. does anybody know why tonight the I-builds did not run?

It looks lile the builds were changed by @MohananRahul, but I fail to understand the logic behind the config: 
https://ci.eclipse.org/releng/job/Builds/job/I-build-4.30/jobConfigHistory/showDiffFiles?timestamp1=2023-11-04_12-06-41&timestamp2=2023-11-04_12-11-42

iloveeclipse:
Looking closer into the config, it says : "would next run at Sunday, November 5, 2023 at 6:00:59 AM Eastern Standard Time."

image

eclipse.platform.common/bundles/org.eclipse.platform.doc.isv/porting/removals.html
laeubi: > As a side-note, for example since version 7
> Guice 6 is the last version to support javax-annotations

The problem is that Eclipse do not perform such major releases and people usually do not upgrade, also it is usually not feasible to run two eclipse instances just for one plugin.

> For example with https://github.com/eclipse-platform/eclipse.platform/pull/791 a whole bunch of tests was added and what I see they are not in the best shape.

Also note that this extra complications where introduced to make this optional what I personally don't recommended so its a bit strange to argue we made it more complicated lets throw it away because its complicated ;-)

I just know that in the past there where major efforts to keep compatibility with "useful" old plugins and this touches the most basic parts of E4.

MohananRahul:
> > Btw. does anybody know why tonight the I-builds did not run?
> 
> It looks lile the builds were changed by @MohananRahul, but I fail to understand the logic behind the config: https://ci.eclipse.org/releng/job/Builds/job/I-build-4.30/jobConfigHistory/showDiffFiles?timestamp1=2023-11-04_12-06-41&timestamp2=2023-11-04_12-11-42

Milestone week will have 2 builds.

eclipse.platform.common/bundles/org.eclipse.platform.doc.isv/porting/removals.html
HannesWell: > > For example with [eclipse-platform/eclipse.platform#791](https://github.com/eclipse-platform/eclipse.platform/pull/791) a whole bunch of tests was added and what I see they are not in the best shape.
> 
> Also note that this extra complications where introduced to make this optional what I personally don't recommended so its a bit strange to argue we made it more complicated lets throw it away because its complicated ;-)

Testing javax and jakarta should have been done regardless of if javax is optional or not.
The extra compilation to make it optional was not an argument. :)
But without that extra complication everyone would require the javax bundle even when 99.9% are not using it anymore (whenever this will be). So making it optional actually makes it less pressing to remove support for it entirely

HannesWell:
Back to the actual topic: How should we proceed with the deprecation and eventual removal of javax-support?

> But we can of course think about extending the deprecation period. As a suggestion I changed to text to:
scheduled to be disabled by default after the December 2026 release and scheduled for removal after the December 2028 release., which means three years until being disabled by default and five years until support is fully removed.
If a Plugin does not manage to apply the simple change and produce new builds I would consider it dead anyways.

What do you think about that? If there is real demand the javax-support could also stay (disabled by default?) until it causes real pain.

akurtakov:
> Back to the actual topic: How should we proceed with the deprecation and eventual removal of javax-support?
> 
> > But we can of course think about extending the deprecation period. As a suggestion I changed to text to:
> > scheduled to be disabled by default after the December 2026 release and scheduled for removal after the December 2028 release., which means three years until being disabled by default and five years until support is fully removed.
> > If a Plugin does not manage to apply the simple change and produce new builds I would consider it dead anyways.
> 
> What do you think about that? If there is real demand the javax-support could also stay (disabled by default?) until it causes real pain.

I think it's already what we have . The document says that something is deprecated and will stay until XXX date *at least* . E.g. if you look at ICU4J removal item - it's marked as "API removals after June 2022" but it's still there. Similar thing happened with Navigator view. 
Setting longer deprecation period will impose self-limit on our self for no good reason in case an issue arise, current workflow doesn't mandate that *we remove in January 2025" but that *we can remove after January 2025 when we see fit*.

HannesWell:
> current workflow doesn't mandate that we remove in January 2025 but that we can remove after January 2025 when we see fit.

That makes sense too.
If there is agreement on this, I can Update the notice to state that compabitlity is retainted/guaranteed for the next two years and from this point on might be removed at any time without further notice.

Btw. I created an update for the N&N entry:
https://github.com/eclipse-platform/www.eclipse.org-eclipse/pull/85

Can we references that in the deprecation notice?

HannesWell:
> Setting longer deprecation period will impose self-limit on our self for no good reason in case an issue arise, current workflow doesn't mandate that _**we remove in January 2025**_ but that _**we can remove after January 2025 when we see fit**_.

I updated the deprecation notice to reflect that.
@akurtakov, @laeubi please have a look. If you are fine with this I would like to submit that soon so that it will be in the doc of the upcomming release.

laeubi (approved):

HannesWell:
Slightly enhanced the section about how to explicitly enable and disable javax-annotation processing.

HannesWell:
> > Setting longer deprecation period will impose self-limit on our self for no good reason in case an issue arise, current workflow doesn't mandate that _**we remove in January 2025**_ but that _**we can remove after January 2025 when we see fit**_.
> 
> I updated the deprecation notice to reflect that.

Since the updated message is (I assume) in the sense of @akurtakov I'm going to merge this now.

PR #1512, opened 2023-11-05
Update org.junit.platform 1.10.1 and org.junit.vintage 5.10.1

merks:


iloveeclipse:
eclipse-junit-tests can't be built. Is that coming from JDT / do we need https://github.com/eclipse-jdt/eclipse.jdt/pull/78, or is it something different ?

merks:
Let's see how this next build fails.  Maybe the JDT change isn't needed...

merks:
If the I-Build is good, then I assume we don't need the JDT feature bump...

PR #1519, opened 2023-11-07
Update Eclipse.icns to match Apple UI Guideline

BeckerWdf:
Fixes: https://github.com/eclipse-platform/eclipse.platform.ui/issues/1247


merks (approved):
Thanks for tackling this.

BeckerWdf:
I see that "Today is a freeze day". Is this also relevant for this kind of change? If yes, when can we merge this again?

laeubi:
I don't know who is responsible for eclipse artwork ... and when such things should be merged now or for next release, but in general one should not have **code** changes so maybe images are ok?

merks:
@akurtakov 

I would personally prefer to see this merged now.  EPP the analogous changes pending which I assume will also be done for M3...

akurtakov:
I have zero Mac knowledge to estimate the importance of it. As Eclipse is supposed to be follow native LnF the change makes sense to me so take that as +0.5 .

merks:
@BeckerWdf 

Go for it!  👍

PR #1531, opened 2023-11-10
Move JDT javadoc build from ant to maven

laeubi:
Similar to what was done for PDE:

- https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/pull/1501


laeubi:
The resulting jar looks good now so this can be merged as soon as mater is open and jdt javadoc issues are fixed.

akurtakov:
Would you please resolve the conflict so we can see where we stand with this one now?

laeubi:
> Would you please resolve the conflict so we can see where we stand with this one now?

done.

laeubi:
Still has javadoc issues:
- https://github.com/eclipse-jdt/eclipse.jdt.core/issues/1582

eclipse-platform-parent/pom.xml
```diff
@@ -687,6 +687,16 @@
                         <placement>a</placement>
                         <head>Implementation Requirements:</head>
                     </tag>
+                    <tag>
```
akurtakov, line 690: Why is this needed?

laeubi: JDT uses these tags on some places and otherwise javadoc would complain about unknown tags.

laeubi:
> Is there anything still to be done here? It's good to go from my side.

There is one improvement for this regarding usage of third party sources but I'd like to do it on top of this change, so from my side it is ready to go.

PR #1533, opened 2023-11-10
Move Platform javadoc build from ant to maven

laeubi:
Same as
- https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/pull/1531
- https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/pull/1501

also here some javadoc issues have to be fixed before this can really be used.


laeubi:
Beside that, the zip compare looks good.

laeubi:
There is also some manual copy that needs to be replicated...

akurtakov:
@laeubi Do you plan finishing this one?

laeubi:
Now that the JDT parts are merge I'll go on with this but maybe next year if there is no immediate need.

akurtakov:
Next year is fine. The sooner javadoc issues become visible earlier in the process, the better.

laeubi:
@akurtakov rebased and adjusted, lets see what build says here...

laeubi:
@akurtakov this is now somewhat "ready" but revealed some javadoc errors that need to be fixed before.

akurtakov:
We went through that for pde and jdt already. We should use this PR as verification and fix issues one by one again. I'll take on some :).

HannesWell:
Looks like https://github.com/eclipse-platform/eclipse.platform.swt/pull/974 fixed at least the SWT javadoc errors.

laeubi:
@akurtakov build is green now!

akurtakov:
Thanks! I like how releng starts to really shape up.

HannesWell:
Great.
Can we enable some verification in the corresponding repos to see new errors already in verification builds?

laeubi:
> Great. Can we enable some verification in the corresponding repos to see new errors already in verification builds?

You can enable `-Pjavadoc` + configure it to fail on errors...

akurtakov:
My concern with enabling fail on error here is that platform committer would not be able to verify his change in this repo if jdt or equinox javadoc is broken and he is not committer there. One can always open PR and wait but this adds a bit too much responsibility on platform contributors.

merks:
Does/can the IDE detected exactly the same errors?  Should they be configured via project specific preference to produce the same errors?  Better to catch errors in the IDE than to find out later and have it hinder the process of contributing...

laeubi:
> My concern with enabling fail on error here is that platform committer would not be able to verify his change in this repo if jdt or equinox javadoc is broken and he is not committer there. One can always open PR and wait but this adds a bit too much responsibility on platform contributors.

I think it is meant for individual rep's only.... **this one will already fail as aggregator uses fail on error already!**

akurtakov:
Ah yes , with this PR in.

laeubi:
> Does/can the IDE detected exactly the same errors? Should they be configured via project specific preference to produce the same errors? Better to catch errors in the IDE than to find out later and have it hinder the process of contributing...

JDT can, but I reported some things it currently does not detect here:

- https://github.com/eclipse-jdt/eclipse.jdt.core/issues/1858
- https://github.com/eclipse-jdt/eclipse.jdt.core/issues/1582
- https://github.com/eclipse-jdt/eclipse.jdt.ui/issues/960

akurtakov:
JDT's support for javadoc validation is pretty basic and doesn't catch many of the things javadoc tool complains about. It sure can be enhanced but that would require quite some effort so IMHO we can't rely on it short term.

akurtakov:
Platform's javadoc is missing in tonight's I-build https://www.eclipse.org/downloads/download.php?file=/eclipse/updates/4.31-I-builds/I20240113-1800/plugins/org.eclipse.platform.doc.isv_4.31.0.v20240113-0758.jar

akurtakov:
Thinking more about it - no javadoc should break the build.

HannesWell:
> Better to catch errors in the IDE than to find out later and have it hinder the process of contributing...

Absolutely, but IMHO better break in the PR verification than only in the I-build.

laeubi:
It currently don't break because https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/blob/master/eclipse.platform.common/bundles/org.eclipse.platform.doc.isv/build.properties has only an entry for `references` (that is always there) but not one for references/api (what is missing) and the baseline compare is ignored.

Issue #1390, opened 2023-09-23
Eclipse 4.30 Prerequisites: ECF


MohananRahul:
Tracking issue for ECF. M1 is scheduled for Sep 29th

merks:
Currently a snapshot version is being used:

https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/blob/ce9f0a7d18277c7a1d737549a24fc84add3a9eca/eclipse.platform.releng.prereqs.sdk/eclipse-sdk-prereqs.target#L81-L82

There appears to be a problem in that version that will likely need another snaphot build to fix it as noted here:

https://github.com/eclipse-equinox/p2/issues/330#issuecomment-1731008456

/cc @scottslewis

merks:
I'm concerned that ECF is current dragging in quite old dependencies via the httpclient feature:

![image](https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/assets/208716/5625db55-7a35-447b-bd9a-a19facfc189e)

I will try to help @scottslewis get this updated; we need a final release build by next Tuesday and ideally would already have a new build today, latest tomorrow I think.

merks:
FYI, these changes from this PR update the above dependencies to the latest:

https://github.com/eclipse/ecf/pull/78#issuecomment-1810071296

scottslewis:
Hi @merks .   I'm probably able to do a snapshot build today 11/14/2023 in the afternoon/evening pacific time, but I'm unclear about 

a) What (if any) ECF code/build data need to be changed?  (I haven't been able to keep up on all the issue notifications over past 24 hours as have real work)
b) Has anyone done any checking/testing of the new httpclient5 versions that you upgraded too?   I usually do so by running the ECF filetransfer test suite locally (at least), but I wouldn't want this to go right into Eclipse without some sort of smoke testing of the new/updated versions of httpclient5

I'll try to divine the current state of everything wrt ECF filetransfer by looking at the notifications.

merks:
@scottslewis 

a) The PR verification build ran fine so I don't think you need to change anything in your build other than what's in the commit itself.
b) I tested the installer built with the test repository with these versions installed in the installer:

![image](https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/assets/208716/ceedc133-226b-42cb-a3d7-227e925be2b2)

That was able to produce an installation without a glitch that way.

So I think we are in good shape from that point of view...

[...]

PR #1536, opened 2023-11-12
Add the ecf http5 feature back to the target platform

laeubi:
For fallback configuration we want to still ship both implementations in 2023-12 release.

@merks @akurtakov Requires PMC/PL approval, still using @mickaelistria 

See also
- https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/issues/1390
- https://github.com/eclipse-equinox/p2/issues/381#issuecomment-1807081486


merks (approved):
+1 from the PMC

merks:
In the IDE, this pulls in these bundles including the Apache dependencies

image

PR #1542, opened 2023-11-14
Don't publish org.eclipse.orbit.xml-apis-ext to Maven

merks:
It should be mapped to two following as batik generally does:

https://repo1.maven.org/maven2/xml-apis/xml-apis-ext/1.3.04/


merks:
@akurtakov 

I think it's imported we change this soon so that we can verify that the Maven publishing process is working properly.  Given this has no other impact, I assume you're okay to merge this now?

akurtakov:
Please merge it in. This has zero effect on deliverables for which the freeze is but can have pretty bad effect in the last minute if not done proper as we have seen with maven publishing multiple times.

PR #1563, opened 2023-11-22
Add missing output specifications in build.properties

HeikoKlare:


HeikoKlare:
Version bump for test feature was missing. Added in https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/commit/275675798fea221e0cd0c3c5d9c76fa662f218c3

akurtakov:
There is manual rebase needed to fix the conflicts.

akurtakov:
There is a bunch of work that has been duplicated today but IMHO we will have less of these in the future with more and more automation.

PR #1566, opened 2023-11-23
Change compliance level and BREE to 17 in o.e.test.* projects

HeikoKlare:
Since dependencies require Java 17, these projects should target Java 17 as well.


HannesWell (approved):
Totally makes sense.

akurtakov:
Any reason why this is still a draft ?

HeikoKlare:
No, it depended on #1563, but should now also be fine due to #1606.

PR #1575, opened 2023-11-24
Remove PDE user docs from aggregator

laeubi:
See https://github.com/eclipse-pde/eclipse.pde/pull/860


akurtakov:
Pde.doc.user bundle has been moved and build will fail until there is new I-build to pick up latest pde.

PR #1589, opened 2023-11-27
Version bumps for 4.31 stream

akurtakov:


akurtakov:
Tracked in https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/issues/1590

Issue #1590, opened 2023-11-27
Version bumps for 4.31 stream


akurtakov:
Placeholder for the version bumps for 4.31 stream.

akurtakov:
[I20231128-1030](https://download.eclipse.org/eclipse/downloads/drops4/I20231128-1030) is clean and doesn't need any version bumps. New version bumps will be caused by respective changes.

PR #1593, opened 2023-11-27
Version bumps for 4.31 stream

akurtakov:
Tracked in eclipse-platform/eclipse.platform.releng.aggregator#1590


akurtakov:
Merge without waiting so the next I-build is a bit saner.

PR #1594, opened 2023-11-28
IBuild schedule update and Performance test fixing

MohananRahul:
Java home update to java17


akurtakov:
@MohananRahul Why do you download and use Java 17.0.2? And even hardcoded in the script. Latest Java 17 is 17.0.9 and what you used is full of CVEs - https://security.snyk.io/package/linux/oracle:8/java-17-openjdk . 
Furthermore, there should be a good list of jvms already available for usage in builds without the need to download https://wiki.eclipse.org/Jenkins#Eclipse_Temurin . Please change that!

PR #1597, opened 2023-11-28
Update smoke tests

akurtakov:
Get rid of old unused configs and Java 20 configs.


akurtakov:
Verification build will not test this one thus merging.

PR #1600, opened 2023-11-28
Document all Milestones being lightweight

akurtakov:
This has been decided in the PMC call at the November 2023 call as it's extra work for releng.

### DIFF
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -9,22 +9,18 @@
 ### Friday before release week:
  * Update [I-builds](JenkinsJobs/Builds/I_build.groovy) to build on the milestone schedule (Twice daily at 06:00 EST and 18:00 EST except Thursday).
  * Create or update prerequisite issues for tracking ECF, EMF and Orbit
- * Send reminder email for upcoming milestone week to platform-releng-dev@eclipse.org, platform-dev@eclipse.org, eclipse-dev@eclipse.org and equinox-dev@eclipse.org
-   * [Example from 4.23 M1](https://www.eclipse.org/lists/platform-releng-dev/msg38067.html) but the usual schedule:
-     * Monday: Last day of development.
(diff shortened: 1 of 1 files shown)


merks:
So we mean specifically m1, m2, and m3 but not rc1 and rc2.  Yes, we all agreed that this reflects the reality of how this has mostly been working already.

HannesWell:
In my opinion this is the right decision.
Nevertheless I think it would be good to announce even light-weight milestones on the usual Friday before, just to make everyone aware of it.
Otherwise it might happen that milestones are accidentally missed and devs have adjusted their priorities if they would have been aware.

akurtakov:
I would appreciate if someone can automate this somehow based on the calendar .

HannesWell:
> I would appreciate if someone can automate this somehow based on the calendar .

I think that would be possible using a milestone that (e.g.) runs every friday and then fetches the next milestone, using the `octokit.rest.issues.listMilestones()` GH API method (https://octokit.github.io/rest.js/v20#issues) in a github-script action.
And if the milestone is due one week later, it sends the corresponding mails to the mailing lists.

To send a mail the [Send mail Github Action](https://github.com/marketplace/actions/send-email) could be used, but it would require us to have a corresponding bot account for which we have the credentials and that is subscribed to the mailing lists.
Alternatively if the action is not trust worthy enough the infra-team maybe has other means to send such a mail.
@fredg02 can you tell if there is a preferred way to send from a GH workflow a mail to an Eclipse mailing list?

fredg02:
> @fredg02 can you tell if there is a preferred way to send from a GH workflow a mail to an Eclipse mailing list?

We recommend that the project creates a GMail account for this bot. We are discussing how this bot user can be added to the mailing lists in a pragmatic way. Stay tuned for an update about this.

PR #1603, opened 2023-11-29
Make use of BREE configurable and fix documentation link

laeubi:
See
- https://github.com/eclipse-pde/eclipse.pde/pull/913
- https://gitlab.eclipse.org/eclipsefdn/helpdesk/-/issues/4011

With only a few plugins actually using a lower BREE < 9 we could also think about dropping this completely over time and instead rely on the `-release` option of modern JDKs

### DIFF
--- a/eclipse-platform-parent/pom.xml
+++ b/eclipse-platform-parent/pom.xml
@@ -138,6 +138,8 @@
     <tycho.addMavenDescriptor>false</tycho.addMavenDescriptor>
     <!-- allows submodules to enable the mapping of p2 artifacts to maven artifacts -->
     <tycho.mapP2Dependencies>false</tycho.mapP2Dependencies>
+    <!-- allows submodules to disable the usage of BREE for their builds -->
+    <tycho.useJDK>BREE</tycho.useJDK>
     
     <compare-version-with-baselines.skip>true</compare-version-with-baselines.skip>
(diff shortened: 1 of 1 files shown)


akurtakov:
What about local builds using newer JVMs e.g. Java 21? Bundles using useJDK=SYSTEM will all be built with Java 21 compliance?

laeubi:
> What about local builds using newer JVMs e.g. Java 21? Bundles using useJDK=SYSTEM will all be built with Java 21 compliance?

They will be build with Java 21 but compiled with Java 17 bytecode/target see
- https://docs.oracle.com/en/java/javase/17/docs/specs/man/javac.html#option-target

akurtakov:
> > What about local builds using newer JVMs e.g. Java 21? Bundles using useJDK=SYSTEM will all be built with Java 21 compliance?
> 
> They will be build with Java 21 but compiled with Java 17 bytecode/target see
> 
>     * https://docs.oracle.com/en/java/javase/17/docs/specs/man/javac.html#option-target

Ah, I totally misunderstood the goal here. So the goal is to reduce the need for older JVMs lurking around. That's a goal I like a lot!

laeubi:
Yes currently we compile with **THAT** Jre and require it on the toolchains but this is only to have bundles to choose not to use that jdk.

PR #1605, opened 2023-11-29
Update the platform setup to redirect 4.29 to 4.30 and 4.30 to 4.31

merks:


merks:
To update the your existing IDE so that it's set for 4.31 development...

Pull all the Git repository.  To Help -> Perform Setup Tasks...

This variable will be prompted:

image

The default is check marked...

Continue and perform all the tasks:

image

This will install the 4.31 version of the IDE.  A restart will be required at which point the other tasks will perform.

I found the API baseline errors didn't all go away.


In the preferences, delete it

image

Then Help -> Perform Setup Tasks... will create new.

PDE UI has some errors. I guess they will be fixed soon:

image

PR #1612, opened 2023-11-30
use 3GB Heap on Platform Workspace

jukzi:
After few days of doing much work in platform workspace across several repositories i use to end up in OOME.
I never could find a single point of failure, memory is allocated by jdt, platform, jgit, pde, classloader ... it seems like it just sums up
image

If anybody wants that heapdump just tell me


HannesWell:
I would even go for `4g` are nowadays.

In general specifing the xmx/s in the project setup can cause trouble if one has multiple OOMPH project-streams in the workspace setup.
In such cases the ini is often just changed back and forth.

merks:
4g seems good. The task should be merged.  I’ve not seen it get changed repeatedly. That would be annoying.

HannesWell (approved):

merks:
I do believe the configuration will win.  But note that the Configuration is used once initially in the installer and it's parts are copied into Navigate -> Open Setup -> Installation and Navigate -> Open Setup -> Workspace where they can only be changed in the future by editing those files.

jukzi:
@merks 
the platform workspace setup wants to change my XMX
image
how to update that value? I have not followed it exactly, but my impression is that it randomly changes between 2,3 and 4 GB

jukzi:
> You can try Navigate

can't we just update all conflicting setting to 4G in the repositorie(s)- wherever they are hidden?

HannesWell:
> Yes, these can all be changed:
> 
> image
> 
> 

Instead of adjusting all project specific setups, wouldn't it be simpler to just define it once in a parent?

PR #1621, opened 2023-12-04
Fix the wrong description of the "Without test code" classpath attribute

jarthana:


merks:
That's strange to get a license failure for these changes!

FYI, I am really updating some library versions via this change:

https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/pull/1623

There the license review is proceeding, though I thought from my Orbit builds that all these were approved already....

SarikaSinha (approved):

PR #1642, opened 2023-12-06
Trigger the dash workflow on behalf of a PR

laeubi:


merks:
Woo hoo!

laeubi:
> Woo hoo!

Don't be too impressed, I cheated here a bit by adding the otherwise disabled module temporary to trigger review on behalf for PDE project :-D

PR #1647, opened 2023-12-07
Add new Script for Daily repo pruning

MohananRahul:
Removing ppcle build


JenkinsJobs/Cleanup/pruneDailyRepos.groovy
```diff
@@ -0,0 +1,70 @@
...
+.../sdk/cleaners/dailyCleanBuildMachine.sh
+and other such scripts.
+  ''')
+
+  logRotator {
+    daysToKeep(10)
+    numToKeep(5)
+  }
+
+  jdk('oracle-jdk8-latest')
```
akurtakov, line 15: This looks suspicious. Does it really need Java 8? Actually does it need Java at all?

MohananRahul: Java is not required , will remove it.

PR #1656, opened 2023-12-12
Update org.mortbay.jasper:apache-jsp to 9.0.83.

merks:


akurtakov:
Resolving fails with :
```
Error:    Software being installed: org.eclipse.platform.feature.group 4.31.0.qualifier
Error:    Missing requirement: org.mortbay.jasper.apache-jsp 9.0.83 requires 'java.package; org.apache.el.util [9.0.0,10.0.0)' but it could not be found
Error:    Cannot satisfy dependency: org.eclipse.equinox.jsp.jasper 1.2.0.qualifier depends on: java.package; org.apache.jasper.servlet [9.0.0,10.0.0)
Error:    Cannot satisfy dependency: org.eclipse.help.feature.group 2.3.1700.qualifier depends on: org.eclipse.equinox.p2.iu; org.eclipse.equinox.jsp.jasper 0.0.0
Error:    Cannot satisfy dependency: org.eclipse.platform.feature.group 4.31.0.qualifier depends on: org.eclipse.equinox.p2.iu; org.eclipse.help.feature.group 0.0.0
Error:  
```
thus licensecheck can't gather the list of deps and fails.

merks:
Yes, I noticed that after looking again closely.  The Orbit build checks that all requirements resolve so it appears something needs to be added to the target platform.  I will investigate...

merks:
The licenses are good now.  I will look closely at the code itself because I recalls "strange el dependencies"....

PR #1661, opened 2023-12-14
Config tycho-compiler-plugin to use project settings

akurtakov:
https://github.com/eclipse-platform/eclipse.platform/pull/973 identified that IDE builds and maven builds produce different bytecode due to ignoring project settings which shouldn't be the case

### DIFF
--- a/eclipse-platform-parent/pom.xml
+++ b/eclipse-platform-parent/pom.xml
@@ -482,7 +482,7 @@
             <excludeResources>
               <exclude>**/package.html</exclude>
             </excludeResources>
-            <useProjectSettings>false</useProjectSettings>
+            <useProjectSettings>true</useProjectSettings>
           </configuration>
         </plugin>
         <plugin>


iloveeclipse:
@akurtakov : if you want continue, it would be a good moment. I'm pushing "touch bundles" PR's right now...

akurtakov:
I'm looking into it but I'm not sure I can make it build in time. Please let me know when you're going to start I-build here - just in case.

iloveeclipse:
> Please let me know when you're going to start I-build here - just in case.

Sure, but it will take some time, there will be bundles with not changed manifests etc. There is always something...

laeubi:
@akurtakov maybe we should simply remove "fail on warnings" from all projects and use quality gate instead.

iloveeclipse:
> maybe we should simply remove "fail on warnings" from all projects and use quality gate instead.

Where is that specified? Maven or jenkins file or project preferences? 
if you can, propose a PR & if Alex would not make it in time, we can merge that additionally, and later fix every project that insists to see errors on warnings.

laeubi:
> Where is that specified? Maven or jenkins file or project preferences?

sometimes on Jenkinsfile sometimes in poms as far as I know.

akurtakov:
OK, build succeeded and I'm pushing this one.

Issue #1660, opened 2023-12-13
New app icon somehow doesn't look aligned


soethoff:
The new icon introduced with Pull Request https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/pull/1519 doesn't look right in the center of the white box, at least compared to other applications (as already mentioned in https://github.com/eclipse-platform/eclipse.platform.ui/issues/1247#issuecomment-1798459287). 

![eclipse_icon_switcher](https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/assets/34278657/a7e4e53d-c9aa-4e15-acbc-aa7588ca771d)

What do you think? 

As mentioned in https://github.com/eclipse-platform/eclipse.platform.ui/issues/1247#issuecomment-1798471181 the alignment is currently done based on the bluish circle, but maybe an alignment based on the icon as a whole (bluish circle incl. orange moon-like shadow) would be a better solution?

PR #1666, opened 2023-12-15
Center the macOS App Icon

BeckerWdf:
Center the macOS Eclipse App Icon so that the Eclipse Logo is centered horizontally and vertically in the "white box".

Fixes: https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/issues/1660


merks (approved):
I don't have a mac.  I assume you will do a good job as always.

PR #1672, opened 2023-12-16
Add org.apache.felix.http.servlet-api to resolve maven warnings

laeubi:
This is to temporary resolve the issue with warning about unsatisfied contracts / packages of org.osgi.service.http and
org.osgi.service.http.whiteboard once we have resolved the discrepancy in org.eclipse.osgi.services this can be replaced / removed again.

PR #1677, opened 2023-12-18
Remove org.osgi.service.http and felix.http.servlet-api from SDK TP

HannesWell:
With https://github.com/eclipse-equinox/equinox/pull/403 the sources of org.osgi.service.http are provided in a re-packed bundle from Equinox, that also provides the OSGi capabilities which where temporarily provided by felix.http.servlet-api.

As planned, this also reverts https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/pull/1672 because it is not necessary anymore.

### DIFF
--- a/eclipse.platform.releng.prereqs.sdk/eclipse-sdk-prereqs.target
+++ b/eclipse.platform.releng.prereqs.sdk/eclipse-sdk-prereqs.target
@@ -317,13 +317,6 @@
 				  <version>9.0.83.1</version>
 				  <type>jar</type>
 			  </dependency>
-			  <!-- This is to temporary resolve the issue with warning about unsatisfied contracts / packages of org.osgi.service.http and org.osgi.service.http.whiteboard -->
-			  <dependency>
-				<groupId>org.apache.felix</groupId>
-				<artifactId>org.apache.felix.http.servlet-api</artifactId>
-				<version>1.2.0</version>
-				<type>jar</type>
-			  </dependency>
 		  </dependencies>
 	  </location>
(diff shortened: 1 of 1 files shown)


HannesWell:
Build is green and looking at the build's console output I can see no `[WARNING] Problems resolving provisioning plan` regarding the osgi packages or bundles. So this is good to go.

akurtakov:
@HannesWell Would you please adjust javadocs build ? They are broken now https://download.eclipse.org/eclipse/downloads/drops4/I20231218-1800/compilelogs/platform.doc.isv.javadoc.txt . If/when https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/pull/1533 is finished failing javadocs should fail the build so issues like this one should be uncovered earlier.

laeubi:
@HannesWell I suspect that the platform build now fails de to these changes as well:
https://ci.eclipse.org/platform/job/eclipse.platform/job/master/

it complains about 

> Failed to execute goal org.eclipse.tycho:tycho-p2-director-plugin:4.0.5-SNAPSHOT:materialize-products (create-distributions) on project infocenter-product: Installation of product infocenter for environment linux/gtk/x86_64 failed

HannesWell:
> it complains about
> 
> > Failed to execute goal org.eclipse.tycho:tycho-p2-director-plugin:4.0.5-SNAPSHOT:materialize-products (create-distributions) on project infocenter-product: Installation of product infocenter for environment linux/gtk/x86_64 failed

I think this is more an issue of the replacement of embedded sources itself, but I'm not entirely sure.
Created https://github.com/eclipse-platform/eclipse.platform/pull/1054, which at least fixed my local build (without tests).

PR #1680, opened 2023-12-19
Add old javax.inject as an additional javadoc dependency

laeubi:
Fix https://github.com/eclipse-platform/eclipse.platform/issues/1017

FYI @HeikoKlare


laeubi:
Outstanding IP reviews are not related.

PR #1681, opened 2023-12-19
Add old javax.annotation as an additional javadoc dependency

laeubi:
See https://github.com/eclipse-platform/eclipse.platform/issues/1017

FYI @HeikoKlare


merks:
Strange that the BND things keep complaining now.  Did you commit actual dependencies to those in PDE?

Issue #1689, opened 2023-12-22
Wrong XML report generated when running JUnit(4) tests with EclipseTestRunner


fedejeanne:
If a test **container** (_e.g._ a class/suite) fails while running with the `org.eclipse.test.EclipseTestRunner` then an `error` will be counted but there is no information about it in the generated XML report.

## Steps to reproduce

Create this test class in some plugin `org.eclipse.xyz`:
```java
import org.junit.AfterClass;
import org.junit.Test;
import org.junit.runner.RunWith;
import org.junit.runners.JUnit4;

@RunWith(JUnit4.class)
public class ErrorOnTearDownClass {
	@AfterClass public static void tearDownClass() { throw new RuntimeException(); }

	@Test public void test() {}
}
```
Run it using the `EclipseTestRunner` _i.e._:
* Create a new _Run Configuration_ of type `Eclipse Application`
* Set **Run an application** --> `org.eclipse.test.uitestapplication`
* Add the following **Program Arguments**: `-testpluginname org.eclipse.xyz -classname fully.qualifed.name.to.ErrorOnTearDownClass formatter=,C:\\some_folder`
* Run the application
* The result is an XML file under `C:\\some_folder`, open it

## Current behavior
The resulting XML looks like this:
```xml
<?xml version="1.0" ?>
<testsuite name="fully.qualifed.name.to.ErrorOnTearDownClass" tests="1" failures="0" errors="1" skipped="0" aborted="0">
	<properties>
		...
	</properties>
	<testcase classname="fully.qualifed.name.to.ErrorOnTearDownClass" name="test" time="0.002"/>
</testsuite>
```
The relevant part being: `<testsuite ... errors="1" ...>` and the fact that there is no `<testcase>` showing any error!

## Expected behavior
There should be some `<testcase>` that shows the error mentioned in the `<testsuite>` node.

PR #1690, opened 2023-12-22
Show failed containers in JUnit XML report #1689

fedejeanne:
## What id does
* Change the `LegacyXmlResultFormatter` so it shows failed containers too (_e.g._ test classes)
* Refactor the class: extract 1 new method called `writeIfPresentInMap` and call it from 3 existing methods

Fixes https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/issues/1689

## How to test
* See #1689 for information on how to run the `EclipseTestRunner` and run it
* The generated XML report should have 1 `<testcase>` node showing the failed test class `ErrorOnTearDownClass`, like this:
```xml
	<testcase classname="ErrorOnTearDownClass" name="" time="0.014">
		<error type="java.lang.RuntimeException">java.lang.RuntimeException
	at org.eclipse.test.tests.snippets.resources.ErrorOnTearDownClass.tearDownClass(ErrorOnTearDownClass.java:13)
	at java.base/java.lang.reflect.Method.invoke(Method.java:568)
</error>
	</testcase>
```

### DIFF
--- a/eclipse.platform.releng/bundles/org.eclipse.test/src/org/eclipse/test/LegacyXmlResultFormatter.java
+++ b/eclipse.platform.releng/bundles/org.eclipse.test/src/org/eclipse/test/LegacyXmlResultFormatter.java
@@ -247,8 +247,8 @@ public class LegacyXmlResultFormatter extends AbstractJUnitResultFormatter {
 		void writeTestCase(final XMLStreamWriter writer) throws XMLStreamException {
 			for (final Map.Entry<TestIdentifier, Stats> entry : testIds.entrySet()) {
 				final TestIdentifier testId = entry.getKey();
-				if (!testId.isTest()) {
-					// only interested in test methods
+				if (!testId.isTest() && isSuccessful(testId)) {
+					// only interested in test methods and in failing containers
 					continue;
 				}
 				// find the parent class of this test method
@@ -256,10 +256,16 @@ public class LegacyXmlResultFormatter extends AbstractJUnitResultFormatter {
 				if (!parent.isPresent()) {
 					continue;
 				}
-				final String classname = parent.get().getLegacyReportingName();
+
 				writer.writeStartElement(ELEM_TESTCASE);
-				writer.writeAttribute(ATTR_CLASSNAME, classname);
-				writer.writeAttribute(ATTR_NAME, testId.getDisplayName());
+				if (testId.isTest()) {
+					final String classname = parent.get().getLegacyReportingName();
+					writer.writeAttribute(ATTR_CLASSNAME, classname);
+					writer.writeAttribute(ATTR_NAME, testId.getDisplayName());
+				} else { // is a container
+					writer.writeAttribute(ATTR_CLASSNAME, testId.getDisplayName());
+					writer.writeAttribute(ATTR_NAME, "");
+				}
 				final Stats stats = entry.getValue();
 				writer.writeAttribute(ATTR_TIME, String.valueOf((stats.endedAt - stats.startedAt) / ONE_SECOND));
 				// skipped element if the test was skipped
@@ -273,6 +279,13 @@ public class LegacyXmlResultFormatter extends AbstractJUnitResultFormatter {
 
 				writer.writeEndElement();
 			}
+		}
+
+		private boolean isSuccessful(final TestIdentifier testId) {
+			return !aborted.containsKey(testId)//
+					&& !failed.containsKey(testId) //
+					&& !errored.containsKey(testId)//
+					&& !skipped.containsKey(testId);
 		}
 
 		private void writeSkipped(final XMLStreamWriter writer, final TestIdentifier testIdentifier)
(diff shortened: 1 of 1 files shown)


eclipse.platform.releng/bundles/org.eclipse.test/src/org/eclipse/test/LegacyXmlResultFormatter.java
```diff
@@ -247,19 +247,25 @@ void writeProperties(final XMLStreamWriter writer) throws XMLStreamException {
...
 				if (!parent.isPresent()) {
 					continue;
 				}
-				final String classname = parent.get().getLegacyReportingName();
+
 				writer.writeStartElement(ELEM_TESTCASE);
-				writer.writeAttribute(ATTR_CLASSNAME, classname);
-				writer.writeAttribute(ATTR_NAME, testId.getDisplayName());
+				if (testId.isTest()) {
+					final String classname = parent.get().getLegacyReportingName();
```
iloveeclipse, line 262: Not sure if `parent.get()` can be null, so please check that.

In general, please be very defensive in changing this code, as an error here might affect entire SDK build.

eclipse.platform.releng/bundles/org.eclipse.test/src/org/eclipse/test/LegacyXmlResultFormatter.java
```diff
@@ -290,56 +303,38 @@ private void writeSkipped(final XMLStreamWriter writer, final TestIdentifier tes
...
+		private static void writeIfPresentInMap(XMLStreamWriter writer, TestIdentifier testIdentifier,
+				String elemFailure, Map<TestIdentifier, Optional<Throwable>> map, boolean writeExceptionStackTrace)
+				throws XMLStreamException {
+			if (!map.containsKey(testIdentifier)) {
 				return;
 			}
-			writer.writeStartElement(ELEM_ABORTED);
-			final Optional<Throwable> cause = aborted.get(testIdentifier);
+			writer.writeStartElement(elemFailure);
+			final Optional<Throwable> cause = map.get(testIdentifier);
```
iloveeclipse, line 326: `cause` is fetched from map and so can be null. Please be careful and do not dereference it without null check below.

iloveeclipse:
@fedejeanne : I'm actually on vacation already, so I will be not able to revert this if that breaks the build.

If that fixes https://github.com/eclipse-jdt/eclipse.jdt.ui/issues/999, it is nice, but it can also break entire SDK reporting.

So if you plan to merge it, be sure that you will be available on the next day to check SDK build results and if anything would break, make sure to fix it or revert this change.

eclipse.platform.releng/bundles/org.eclipse.test/src/org/eclipse/test/LegacyXmlResultFormatter.java
fedejeanne: The `null`-check happens in line 322 and the presence of the value inside the optional is checked in line 327

eclipse.platform.releng/bundles/org.eclipse.test/src/org/eclipse/test/LegacyXmlResultFormatter.java
fedejeanne: That's being checked in line 256 so `parent.get()` can't be `null`

fedejeanne:
> @fedejeanne : I'm actually on vacation already, so I will be not able to revert this if that breaks the build.
> 
> If that fixes [eclipse-jdt/eclipse.jdt.ui#999](https://github.com/eclipse-jdt/eclipse.jdt.ui/issues/999), it is nice, but it can also break entire SDK reporting.
> 
> So if you plan to merge it, be sure that you will be available on the next day to check SDK build results and if anything would break, make sure to fix it or revert this change.

I'm actually going on holidays in like 5 minutes so I'd say we can wait to merge this :-)

In fact I'm going to be on parental leave for 3 months so I'll only be checking my e-mails/notifications occasionally. I'd say we leave this PR open for a few days and see if someone has a big objection to it before merging it since I won't be able to react quickly to requested changes. Would that be ok for you?

eclipse.platform.releng/bundles/org.eclipse.test/src/org/eclipse/test/LegacyXmlResultFormatter.java
laeubi: > That's being checked in line 256 so `parent.get()` can't be `null`

It usually makes things more clear to extract the method call into a local variable in such cases.

iloveeclipse:
> In fact I'm going to be on parental leave for 3 months

Congrats :)

> I'd say we leave this PR open for a few days and see if someone has a big objection to it before merging it since I won't be able to react quickly to requested changes. Would that be ok for you?

Sure. Simeon will take over, he wanted to work next week.

@trancexpress : could you please review & merge that on your next working day next week, and in case SDK test reports would be broken in some unexpected way, revert it again?

This should fix the "wrong" error reported for JDT UI tests, see https://github.com/eclipse-jdt/eclipse.jdt.ui/issues/999 .

fedejeanne:
> > In fact I'm going to be on parental leave for 3 months
> 
> Congrats :)

Thank you :)

> This should fix the "wrong" error reported for JDT UI tests, see [eclipse-jdt/eclipse.jdt.ui#999](https://github.com/eclipse-jdt/eclipse.jdt.ui/issues/999) .

Just to make sure we're on the same page: this PR will *not* get rid of the "error" shown in the report mentioned in https://github.com/eclipse-jdt/eclipse.jdt.ui/issues/999, it will simply add one `<testcase>` entry to the resulting XML report (and therefore to the HTML report generated from it) and that entry will show the failing container. Whether or not that entry is related to the _Aborted_ test introduced by me in https://github.com/eclipse-jdt/eclipse.jdt.ui/pull/953 remains to be seen.

eclipse.platform.releng/bundles/org.eclipse.test/src/org/eclipse/test/LegacyXmlResultFormatter.java
fedejeanne: I'd rather leave it as it is, I prefer early returns/breaks over nesting.

trancexpress:
Merging this, I'll monitor build results for a few days to see if we want to revert the change. Code-wise it looks fine.

trancexpress:
The JDT UI test error is now shown in the `xml` report:

```
org.eclipse.jdt.junit.tests.JUnit5TestFinderJupiterTest		Error	testresources/testClasses/JupiterTests.java

java.nio.file.NoSuchFileException: testresources/testClasses/JupiterTests.java
at java.base/sun.nio.fs.UnixException.translateToIOException(UnixException.java:92)
at java.base/sun.nio.fs.UnixException.rethrowAsIOException(UnixException.java:106)
at java.base/sun.nio.fs.UnixException.rethrowAsIOException(UnixException.java:111)
at java.base/sun.nio.fs.UnixFileSystemProvider.newByteChannel(UnixFileSystemProvider.java:218)
at java.base/java.nio.file.Files.newByteChannel(Files.java:380)
at java.base/java.nio.file.Files.newByteChannel(Files.java:432)
at java.base/java.nio.file.Files.readAllBytes(Files.java:3288)
at java.base/java.nio.file.Files.readString(Files.java:3366)
at java.base/java.nio.file.Files.readString(Files.java:3325)
at org.eclipse.jdt.junit.tests.JUnit5TestFinderJupiterTest.createCompilationUnit(JUnit5TestFinderJupiterTest.java:144)
at org.eclipse.jdt.junit.tests.JUnit5TestFinderJupiterTest.beforeClass(JUnit5TestFinderJupiterTest.java:107)
at java.base/java.lang.reflect.Method.invoke(Method.java:568) 
```

https://download.eclipse.org/eclipse/downloads/drops4/I20231226-1800/testresults/html/org.eclipse.jdt.ui.tests_ep431I-unit-cen64-gtk3-java17_linux.gtk.x86_64_17.html#org.eclipse.jdt.ui.tests

I'm not noticing anything else changed in the test reports for the latest build (`I20231226-1800`). So from my POV the change is fine and doesn't need a revert. I'll continue checking test results from integration builds for a few days, just in case.

fedejeanne:
@trancexpress thank you very much for reviewing and merging!

PR #1691, opened 2023-12-25
Print exception when executing maven

laeubi:
Maven by default do not print stack traces when a mojo fails ,this makes it quite hard to understand build problems that are caused by unexpected failures.

This adds the -e switch to maven to print full stack trace.


akurtakov (approved):
There should be no harm in this one.

PR #1709, opened 2024-01-11
globally set the jdt.buildtime flag in the parent target-configuration

laeubi:


laeubi:
Merging as this will already fix some problems in the individual builds, for ibuild I need to check why Tycho do not "see" the requirement anymore in the feature.

PR #1718, opened 2024-01-13
Add back readmetool example to platform.doc.isv

akurtakov:
It was missed in the conversion to maven.


laeubi:
Thanks for finding/fixing this...

PR #1719, opened 2024-01-13
[Setup] Add JGit LFS to Platform Aggregator setup

HannesWell:
The eclipse.platform.swt sub-module contains Git LFS content. Therefore in order to fully pull it with EGit/JGit the LFS plugin is necessary.

@merks or should we just add it to the parent Platform project? Then it would not be necessary to have it in the SWT setup and the Platform Aggregator setup.
Having it installed is usually not a problem if not used.

@laeubi FYI

### DIFF
--- a/oomph/Platform.setup
+++ b/oomph/Platform.setup
@@ -60,16 +60,26 @@
           key="/instance/org.eclipse.pde.api.tools/MISSING_EE_DESCRIPTIONS"
           value="Warning"/>
     </setupTask>
+    <setupTask
+        xsi:type="setup:CompoundTask"
+        name="org.eclipse.egit.ui">
+      <setupTask
+          xsi:type="setup:PreferenceTask"
+          key="/instance/org.eclipse.egit.ui/auto_lfs_config"
+          value="true"/>
+    </setupTask>
   </setupTask>
   <setupTask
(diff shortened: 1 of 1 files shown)


merks:
There is no harm in adding/installing it more generally.

HannesWell:
> There is no harm in adding/installing it more generally.

Moved it to the existing `Platform` P2 Director and renamed it to `Platform and Git LFS`. Or would you prefer to have two P2-Director tasks here?

merks:
> Or would you prefer to have two P2-Director tasks here?

Either way is fine.

merks (changes requested):
Note that there is no need to specify an EGit site because if there is any Git Clone task, this will be added automatically:

image

So probably better not to add it.

HannesWell:
> Note that there is no need to specify an EGit site because if there is any Git Clone task, this will be added automatically:

Thanks for the hint. I saw it in my setup that the EGit site is available but since I didn't found a P2 director in the parents I assumed it is due to my local user, workspace or installation setup. Didn't know or at least wasn't sure that the Git clone tasks adds it implicitly.

HannesWell:
@merks I assume you are now fine with this change, aren't you?

merks (approved):
Yes, looks good.

PR #1700, opened 2024-01-06
Remove eclipse.platform.swt.binaries sub-module to reflect merge with eclipse.platform.swt

HannesWell:
With https://github.com/eclipse-platform/eclipse.platform.swt/pull/956 the swt native binaries are stored in the `eclipse.platform.swt` repository using git-lfs. The `eclipse.platform.swt.binaries` repository is therefore obsolete.

Part of https://github.com/eclipse-platform/eclipse.platform.swt.binaries/issues/2

PR #1721, opened 2024-01-14
Fix or remove references to obsolete eclipse.platform.swt.binaries repo

HannesWell:
Fix paths that are still relevant and remove obsolete occurrences. Especially this fixes the provision of the SWT.zip files at the Eclipse download page.
Currently for example https://download.eclipse.org/eclipse/downloads/drops4/I20240108-1340/ only shows N/A for the size of the SWT.zips and if one tries to download them, the server says they are not available.

Part of https://github.com/eclipse-platform/eclipse.platform.swt.binaries/issues/2
This should have be done in https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/pull/1700, but I forgot it.


HannesWell:
There are a few other hits for eclipse.platform.swt.binaries left in this repository, but I assume they are fine:

grafik

In general I wonder if it is correct to have pde_basedir.properties for jdt-doc?
Or is this just ignored?

HannesWell:
@akurtakov does this look good to you?

akurtakov (approved):
Looks good to me

Issue #1724, opened 2024-01-15
4.31 I-Build: I20240114-1800 - BUILD FAILED


MohananRahul:
See https://ci.eclipse.org/releng/job/Builds/job/I-build-4.31/91/console

jukzi:
082c1024f41529428293828e287860b9f4a19d0a
```
00:07:42  [INFO] --- antrun:3.1.0:run (collect-java-sources) @ org.eclipse.swt.gtk.linux.x86_64 ---
00:07:42  [INFO] Executing tasks
00:07:42  [INFO]      [exec] build-scripts/CollectSources.java:35: error: illegal start of expression
00:07:42  [INFO]      [exec] 		private static ScriptEnvironment read(String[] args) {
00:07:42  [INFO]      [exec] 		^

```
@HannesWell

HannesWell:
That's unexpected, because in general it works to launch the Java script, as one can see at the SWT master branch build.
But I have seen similar errors in early versions of the PR when the java executable used by the Jenkins executor by default was a Java-11 one. But this script requires Java-17, mainly due to the use of records (but maybe also other library calls).

I can investigate this in detail this evening, unless somebody else has fixed it in the meantime. But my current guess is that the Maven build is launched with an explicit java exe and not the default one on the system, which I assume is still Java-11.
Maybe it is sufficient to add a tools section to the build stage that installs java-17.

HannesWell:
I replayed the I build with and without the new tools section echoing the value of `PATH` and with

```
  tools {
      jdk 'openjdk-jdk17-latest'
      maven 'apache-maven-latest'
  }
```
PATH is
```
/opt/tools/java/openjdk/jdk-17/latest/bin:
/opt/tools/apache-maven/latest/bin:
/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin
```
and without PATH is just
```
/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin
```
So the tools are prepended as expected, but somehow the `sshagent` step does not like the newer java version (I assume maven is not relevant here)?
If the tools are removed again, the `Setup intial configuration` works...

akurtakov:
Is the path really multiline ?

HannesWell:
> Is the path really multiline ?

No it is linear. I just split it for better comparability.

I now did some replays (with failure mails disabled to not spam everybody) and it looks like that the 
[...]

PR #1727, opened 2024-01-15
Declare required jdk and maven as top-level pipeline tools in I-builds

HannesWell:
Since https://github.com/eclipse-platform/eclipse.platform.swt/pull/973 the I-build (like the SWT build) requires a java-17 (or later) jdk on the `PATH` system environment-variable in order to launch a `jdk>=17` with the 'java' command from the CLI. Just setting the `JAVA_HOME` variable to a path pointing to a jdk-17 is not sufficient since the the java executable is usually located in `$JAVA_HOME/bin`. And while the maven launch script searches in `$JAVA_HOME/bin` for a java executable to run the build itself, a simple java command on the CLI isn't that smart. 
Previously running the `java` command would just run the default default java version on the Jenkins executor.

In general this simplifies the build-configuration because the JDK to use is only defined once and used everywhere.

Hopefully fixes https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/issues/1724. I have no way to test this in advance but the explanation above is the only one I found why the Maven build itself runs fine (which requires Java-17 because of Tycho), but fails as soon as a `java` command is launched that requires java>=17.

### DIFF
--- a/JenkinsJobs/Builds/I_build.groovy
+++ b/JenkinsJobs/Builds/I_build.groovy
@@ -147,13 +147,15 @@ spec:
 """
     }
   }
+  tools {
+      jdk 'openjdk-jdk17-latest'
+      maven 'apache-maven-latest'
+  }
   environment {
       MAVEN_OPTS = "-Xmx6G"
       CJE_ROOT = "${WORKSPACE}/eclipse.platform.releng.aggregator/eclipse.platform.releng.aggregator/cje-production"
-      PATH = "$PATH:/opt/tools/apache-maven/latest/bin"
       logDir = "$CJE_ROOT/buildlogs"
     }
-  
   stages {
       stage('Clean Workspace'){
           steps {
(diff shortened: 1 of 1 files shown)


akurtakov:
This is step in the right direction. Please merge and start new I-build so results are seen sooner.

HannesWell:
> This is step in the right direction. Please merge and start new I-build so results are seen sooner.

Done. Is there anything to do to synchronize the change in the groovy file with the actual job-configuration?
If I understand it correctly https://ci.eclipse.org/releng/job/Create%20Jobs/ does that? But it failed in the last two runs?

HannesWell:
> > This is step in the right direction. Please merge and start new I-build so results are seen sooner.
> 
> Done. Is there anything to do to synchronize the change in the groovy file with the actual job-configuration? If I understand it correctly https://ci.eclipse.org/releng/job/Create%20Jobs/ does that? But it failed in the last two runs?

The error was:`
ERROR: You must configure the DSL job to run as a specific user in order to use the Groovy sandbox.`
and it happened for all timer-based executions.
The one manually started by @MohananRahul succeeded so I assumed it only works if a user explicitly triggers the job and just triggered it too. The run succeed and looking at the config of `https://ci.eclipse.org/releng/job/Builds/job/I-build-4.31/` now it looks like this change was successfully applied.
I just triggered [I-build-4.31#92](https://ci.eclipse.org/releng/job/Builds/job/I-build-4.31/92) to verify it.

HannesWell:
> And while the maven launch script searches in `$JAVA_HOME/bin` for a java executable to run the build itself, a simple java command on the CLI isn't that smart.

For the record, that behavior of Maven is mentioned in the [Maven Installation Guide](https://maven.apache.org/install.html).

PR #1728, opened 2024-01-15
Revert declaration of maven tool to unbreak sshagent step in I-build

HannesWell:
Fixes https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/issues/1724#issuecomment-1892602584


HannesWell:
The [I-build-4.31#97](https://ci.eclipse.org/releng/job/Builds/job/I-build-4.31/97/), that already contains this changes through a Replay, successfully built the SWT native fragments. Therefore I think this change is sound and I will apply it so that tonight's I-build can pick it up.

PR #1731, opened 2024-01-16
Declare required jdk as top-level pipeline tool for P- and Y-builds

HannesWell:
Applies the following changes to P- and Y-builds:
- https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/pull/1727
- https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/pull/1728

### DIFF
--- a/JenkinsJobs/YBuilds/P_build.groovy
+++ b/JenkinsJobs/YBuilds/P_build.groovy
@@ -121,6 +121,9 @@ spec:
 """
     }
   }
+  tools {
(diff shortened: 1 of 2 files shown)


HannesWell:
Changes have been applied successfully.

In general I have the impression that I-, Y- and P-builds have a lot of configuration in common just like the test job configuration in `AutomatedTests` and `YBuilds` and I wonder if it would be desirable to unify this to have one template for the main build and the tests from which all the jobs are generated?

PR #1741, opened 2024-01-22
make path names unix friendly - eclipse workbench bundle

elsazac:
Refs: https://github.com/eclipse-platform/eclipse.platform.ui/pull/1460


akurtakov:
The referenced PR is closed so this one is not needed too.

PR #1742, opened 2024-01-22
Update org.apache.felix.scr to 2.2.10

merks:


merks:
@akurtakov 

FYI, I expect this will fail because this change is needed:

https://github.com/eclipse-equinox/equinox/pull/486

Because this feature includes felix.scr:

image

I'm not sure if I can complete this process because I have to leave for dinner in about 1 hour.  Maybe when I get back later, but if you have time, maybe you can push it along.  If not, no problem...

merks:
@akurtakov Thanks for pushing this to completion.

HannesWell:
> Because this feature includes felix.scr:
> 
> image

We could consider to remove `felix.scr` from the feature. Bundles that use have DS-Component should actually have a required capability `Require-Capability: osgi.extender;filter:="(&(osgi.extender=osgi.component)(version>=1.2)(!(version>=2.0)))"` or similar generated into their Manifest so that they should require/pull-in an SCR anyway.

laeubi:
Not all bundles.... but you can require it in the feature instead.

HannesWell:
> Not all bundles.... but you can require it in the feature instead.

The tooling in PDE/Tycho could add it and set the required version, based on the preferences.

laeubi:
> > Not all bundles.... but you can require it in the feature instead.
> 
> The tooling in PDE/Tycho could add it and set the required version, based on the preferences.

It could and Tycho even support this, but this is not mandatory, so if a specific SCR should be part of SDK its best to mention it in the feature.

PR #1747, opened 2024-01-26
Don't skip api analysis if artifact was replaced to always get messages

laeubi:
As people are ["steady confused"](https://github.com/eclipse-platform/eclipse.platform.ui/pull/1557#issuecomment-1910559050) we better should have slower builds than confusing even experienced committers so they constantly complain about "randomness". Also requested [here](https://github.com/eclipse-platform/eclipse.platform.ui/pull/1557#issuecomment-1910099859).

### DIFF
--- a/eclipse-platform-parent/pom.xml
+++ b/eclipse-platform-parent/pom.xml
@@ -954,6 +954,7 @@
                                  </repository>
                              </baselines>
                              <skip>${skipAPIAnalysis}</skip>
+                             <skipIfReplaced>false</skipIfReplaced>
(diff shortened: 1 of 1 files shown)


laeubi:
@akurtakov another problem in the calendar?

akurtakov:
> @akurtakov another problem in the calendar?

Yes, it gives freeze on promotion day in calendar for both milestones and RCs and we have to keep the promotion in the calendar to not forget it so a change in the script to not show freeze on milestones promotion is needed.

PR #1752, opened 2024-01-29
Use Tycho 4.0.5 release

laeubi:


akurtakov:
Please update it in https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/blob/master/.mvn/extensions.xml too.

PR #1758, opened 2024-01-31
Use 4.0.6-SNAPSHOT for latest fixes/enhancements

laeubi:
See
- https://github.com/eclipse-tycho/tycho/pull/3430
- https://github.com/eclipse-tycho/tycho/pull/3447
- https://github.com/eclipse-tycho/tycho/pull/3451
- https://github.com/eclipse-tycho/tycho/pull/3459
- https://github.com/eclipse-tycho/tycho/pull/3460


laeubi:
Searching for 'baseline and build artifacts have same version but different contents' shows some bundles that would need to be touched when this is applied.

PR #1769, opened 2024-02-05
Target platform bundle updates

merks:
Update assertj-core to 3.25.3.
Update junit-jupiter-* and junit-vintage-engine to 5.10.2. Update junit-platform-* to 1.10.2.

### DIFF
--- a/eclipse.platform.releng.prereqs.sdk/eclipse-sdk-prereqs.target
+++ b/eclipse.platform.releng.prereqs.sdk/eclipse-sdk-prereqs.target
@@ -336,7 +336,7 @@
 			  <dependency>
 				  <groupId>org.assertj</groupId>
 				  <artifactId>assertj-core</artifactId>
-				  <version>3.25.2</version>
+				  <version>3.25.3</version>
 				  <type>jar</type>
 			  </dependency>
 			  <dependency>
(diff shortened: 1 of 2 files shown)


merks:
FYI @akurtakov 

Probably JDT's feature will need a forced qualifier change.

image

I'll create a PR in expectation of that.

merks:
@akurtakov 

I see this in the log:
```
09:05:48  [INFO] --- tycho-packaging:4.0.5:package-feature (default-package-feature) @ org.eclipse.jdt ---
09:05:48  [INFO] Building jar: /home/jenkins/agent/workspace/atform.releng.aggregator_PR-1769/eclipse.jdt/org.eclipse.jdt-feature/target/org.eclipse.jdt-3.19.400-SNAPSHOT.jar
09:05:48  [INFO] 
09:05:48  [INFO] --- tycho-p2:4.0.5:p2-metadata-default (default-p2-metadata-default) @ org.eclipse.jdt ---
09:05:48  [INFO] No baseline version MavenProject: org.eclipse.jdt.feature:org.eclipse.jdt:3.19.400-SNAPSHOT @ /home/jenkins/agent/workspace/atform.releng.aggregator_PR-1769/eclipse.jdt/org.eclipse.jdt-feature/pom.xml
```
So it appears that there is no need because the version already changed since the release baseline, which makes me wonder why the test feature always needs special attention...

akurtakov:
I can not explain it as in the past features in another repository needed a touch of the feature. Test feature is a standalone feature (not included in product) while jdt is in the product and a top-down approach might show that the newly built feature is needed by the product and thus auto updated but test feature doesn't have such way.

laeubi:
The feature qualifier aggregator works in quite an "odd" way I have started in improving it but was not yet able to finish with that:
- https://github.com/eclipse-tycho/tycho/pull/3364

PR #1798, opened 2024-02-10
Print the jvm flags and increase max memory to 2gb

laeubi:


laeubi:
Build shows now `MaxHeapSize = 2147483648`

@sravanlakkimsetti I noticed that P and Y builds have some reference in Jenkins/text files for `MAVEN_OPTS="-Xmx6G"` this should be reviewed now, when the build needs really that much memory we better configure it like here and then we need larger runners!

sravanlakkimsetti:
Eclipse platform build requires little bit more than 4GB. Because of this we increased max heap size to 6GB. Thats the reason we have those references. Y-build is similar to I-build so same parameters are used in Y. For P-build you can reduce.

You can verify the memory usage for I-vuild using https://ci.eclipse.org/platform/job/eclipse-aggregator-master/ Currently this uses centos-7-6gb template. If we use centos-8 template this job fails intermittently with running out of memory

Also `MAVEN_OPTS="-Xmx6G"` used in I, Y and P-builds only. We created larger runners for these using build scripts. See
1. https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/blob/master/JenkinsJobs/Builds/I_build.groovy#L76
2. https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/blob/b90ebe8b44d271574d52affa74f5df03890be665/JenkinsJobs/YBuilds/P_build.groovy#L50
3. https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/blob/b90ebe8b44d271574d52affa74f5df03890be665/JenkinsJobs/YBuilds/Y_build.groovy#L70

sravanlakkimsetti:
@laeubi Please give it a try with lower settings. Please try with a smaller container. The `centos-8` template will create a template with 4GB ram and `centos-7-6gb` creates it with 6GB. 

You may need to reduce the container limits on k8s defined in https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/blob/master/JenkinsJobs/Builds/I_build.groovy#L76

Issue #1767, opened 2024-02-02
No ASM source bundles in SDK


iloveeclipse:
Just stumbled (in https://github.com/eclipse-pde/eclipse.pde/issues/1111) over the problem that ASM sources are not included into SDK, so developing PDE API tooling is harder as it should be (no Javadoc, no proper call hierarchy etc). Also same problem with JDT UI that uses ASM...

We should include src bundles for ASM libraries into SDK.

iloveeclipse:
@laeubi : this is something that could be achieved with some clever tycho config, or that need to be a added as a dependency to some sources feature? Are there maven source bundles that can be used in Eclipse at all, or is it only possible with Orbit bundles?

merks:
It would be unhelpful to point out that the target platform with the Oomph setup has the source:

![image](https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/assets/208716/e7b8b380-0b66-402e-bfbd-aa629d88f13f)

If you want sources in the actual SDK product, you should probably include the .source version of these features which I think will pull in the sources of all the requirements of those features.

![image](https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/assets/208716/190fddbb-4661-483f-a316-84c44d2964f4)

Note that sources have been gradually disappearing for 3rd party bundles that are no longer mentioned in (required by) any feature, and that has become the general trend to avoid locking in a specific version of such a dependency and to avoid redundancy of feature requirements that are already properly specified via requirements in bundles.

laeubi:
It depends where asm is from and how the sdk is build, Tycho has an option to include all sources if they are available.

Beside that we added support to PDE to find sources if they are colocated so explicit inclusion should not be needed, also m2e can find sources if they are originating from a maven repository automatically.

merks:
Is that include all sources for a product or a repo?  I assume the latter while @iloveeclipse wants the former.

laeubi:
I never tried it for a product, but in case one can probably add support for products as well, or customize things on the product than on the feature level now we have mixed products.

merks:
Well, I think products should include only what the requirements say they should include so that a director application can install the product from the repo, i.e., the publisher should produce the appropriate p2 metadata, but we digress...

iloveeclipse:
Ed, do I understand you right, you proposed to add something like 
```
   <includes id="org.objectweb.asm.source" version="0.0.0"/>
```
[...]

PR #1800, opened 2024-02-12
Install available sources with the SDK product

laeubi:
Currently only some sources are included in the SDK product these are those that explicitly specified in source features. As removing more and more third party dependencies from features, they will also vanish from source features and therefore missing in the SDK product. Beside that, managing source features itself is a tedious task.

This enables the `<installSources>` option for the SDK product that will install any available source with a bundle automatically.

grafik

FYI @merks @iloveeclipse 

Fix https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/issues/1767

### DIFF
--- a/eclipse.platform.releng.tychoeclipsebuilder/eclipse.platform.repository/pom.xml
+++ b/eclipse.platform.releng.tychoeclipsebuilder/eclipse.platform.repository/pom.xml
@@ -382,6 +382,7 @@ UNmHBQdtflW5G1L5fQggWG7V
             </goals>
             <configuration>
               <profile>SDKProfile</profile>
+              <installSources>true</installSources>
             </configuration>
           </execution>
           <execution>


laeubi:
Technically this adds `<property name="org.eclipse.update.install.sources" value="true"/>` to the profile, that means if one installs additional software into the SDK and the updatesite is using a recent P2/Tycho even these sources will be installed along with the product.

akurtakov (approved):
Looks very promising. https://download.eclipse.org/eclipse/downloads/drops4/I20240211-1800/buildlogs/reporeports/reports/versionChecks.html (with the correct  new Ibuild) is what should be checked to verify what has been added.

laeubi:
@akurtakov I didn't see this comparator erros in verification builds, so probabbly a compiler change that needs deployment of new compiler as well.

Issue #1815, opened 2024-02-21
Bundle org.eclipse.unittest.ui is missing on central


laeubi:
I just noticed `org.eclipse.unittest.ui` is missing on central:

https://mvnrepository.com/search?q=org.eclipse.unittest.ui

it might be because it hast `test` in its name and therefore assumed to only contain test only.

merks:
These are the three bundles that are no longer removed:

![image](https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/assets/208716/abc2a758-8f6a-4fb1-81a6-e49c092ee798)

The validator is running now which tests that each pom's dependencies all be fully/transitively satisfied.

merks:
Validation worked.

https://ci.eclipse.org/releng/view/Publish%20to%20Maven/job/RepositoryValidator/

I will monitor that these three things are published to here, which will take at least 1/2 hour:

https://repo.eclipse.org/content/repositories/eclipse-snapshots/

akurtakov:
https://repo.eclipse.org/content/repositories/eclipse-snapshots/org/eclipse/platform/org.eclipse.unittest.ui/
https://repo.eclipse.org/content/repositories/eclipse-snapshots/org/eclipse/jdt/org.eclipse.jdt.ui.unittest.junit/
https://repo.eclipse.org/content/repositories/eclipse-snapshots/org/eclipse/pde/org.eclipse.pde.unittest.junit/
All there now.

PR #1816, opened 2024-02-21
Avoid removing "unittest" bundles from the Maven publishing

merks:
https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/issues/1815


merks:
This does not impact the build but only the Maven publishing which I can only test after this is merged.

Issue #1825, opened 2024-02-27
Tip of the Day cannot connect to http: because of security policy


Phillipus:
- [x] I verified I can reproduce this issue against [latest Integration Build of Eclipse SDK](https://download.eclipse.org/eclipse/downloads/)

## Steps to reproduce

1. Open Tip of the Day
2. Dialog opens
3. Error Message:

<img width="372" alt="Screenshot 2024-02-27 at 10 25 30" src="https://github.com/eclipse-platform/eclipse.platform/assets/600504/3c746b25-d4ec-4cb1-bda6-095e3de2800b">

It should use `https:` protocol:

https://www.eclipse.org/downloads/download.php?r=1&file=/eclipse/tips/org.eclipse.jdt.tips.user/content/941624342282108929/941624342282108929.html

Eclipse SDK I20240226-1800 (4.31)

## Community

- [x] I understand reporting an issue to this OSS project does not mandate anyone to fix it. Other contributors may consider the issue, or not, at their own convenience. The most efficient way to get it fixed is that I fix it myself and contribute it back as a good quality patch to the project.

Phillipus:
See https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/blob/78b5b46ade77774b421655a41a05e4b4d2a3c68a/eclipse.platform.common/bundles/org.eclipse.jdt.tips.user/src/org/eclipse/jdt/tips/user/internal/JDTTipProvider.java#L37

Phillipus:
This makes Tip of the Day unusable. Is this worth fixing before 4.31 final? @merks

iloveeclipse:
> Is this worth fixing before 4.31 final

I will provide PR in a moment.

merks:
I imagine the fix is risk free.

iloveeclipse:
@Phillipus : could you please check if https://download.eclipse.org/eclipse/downloads/drops4/I20240227-0600 fixes the problem?

Phillipus:
> @Phillipus : could you please check if https://download.eclipse.org/eclipse/downloads/drops4/I20240227-0600 fixes the problem?

That build doesn't have this fix in, so no the problem exists in that build. I guess we have to wait for the next build?

iloveeclipse:
> That build doesn't have this fix in

Damn, I was a bit too late. Sorry, please check with the next one.

iloveeclipse:
[...]

PR #1826, opened 2024-02-27
Change http://www.eclipse.org to https://www.eclipse.org in tips URLs

iloveeclipse:
Fixes https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/issues/1825


akurtakov:
PMC approved. 
I see a good number of a href=http://.... in documenb bundles which will most likely fail similarly with such strong security settings but I don't think this could be fixed for 4.31.

akurtakov (approved):

PR #1848, opened 2024-03-01
Update Parent pom and target sdk deployment jobs for R4_31_maintenance branch

elsazac:
Fixes: https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/issues/1847

### DIFF
--- a/JenkinsJobs/Releng/deployPlatformParentPom.groovy
+++ b/JenkinsJobs/Releng/deployPlatformParentPom.groovy
@@ -22,16 +22,8 @@ For this to be used by consuming projects, they must have their local maven repo
       remote{
         url('https://github.com/eclipse-platform/eclipse.platform.releng.aggregator.git')
       }
-      branch('master')
-      branch('R4_7_maintenance')
-      branch('R4_8_maintenance')
-      branch('R4_19_maintenance')
-      branch('R4_23_maintenance')
-      branch('R4_26_maintenance')
-      branch('R4_27_maintenance')
-      branch('R4_28_maintenance')
-      branch('R4_29_maintenance')
-      branch('R4_30_maintenance')                  
+      branch('master') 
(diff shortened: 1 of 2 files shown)


JenkinsJobs/Releng/deployPlatformParentPom.groovy
```diff
@@ -31,7 +31,8 @@ For this to be used by consuming projects, they must have their local maven repo
       branch('R4_27_maintenance')
```
akurtakov, line 31: Please remove deployment of old versions. That should create less churn for old releases.

JenkinsJobs/Releng/deploySdkPom.groovy
```diff
@@ -32,6 +32,7 @@ For this to be used by consuming projects, they must have their local maven repo
       branch('R4_28_maintenance')
```
akurtakov, line 32: Same

akurtakov (commented):
Deploying only master and 4.31 is probably what needed here.

JenkinsJobs/Releng/deployPlatformParentPom.groovy
elsazac: @akurtakov Thanks for the review. Just to confirm, can remove all the old branches from R4_7_maintenance up till R4_30_maintenance right?

akurtakov: I believe it's best. These branches are done and changes to parent poms risk to break builds entirely. Furthermore, there are no verification builds for these branches so deploying untested content is risky.

elsazac: Sure. I will remove the same.

akurtakov (approved):

PR #1856, opened 2024-03-02
fix typos in ref-startup.htm

chirontt:


BeckerWdf:
Thanks to @chirontt for fixing and to @laeubi for merging.

Issue #1882, opened 2024-03-08
Publish to Maven central


elsazac:

elsazac:
@MohananRahul  
when the final release artefacts are created, could you update it here with details ?

MohananRahul:
Eclipse downloads:
	https://download.eclipse.org/eclipse/downloads/drops4/R-4.31-202402290520/

	New and Noteworthy:
	https://www.eclipse.org/eclipse/news/4.31/

	Update existing (non-production) installs:
	https://download.eclipse.org/eclipse/updates/4.31/

	Specific repository good for building against:
	https://download.eclipse.org/eclipse/updates/4.31/R-4.31-202402290520/

	Equinox specific downloads:
	https://download.eclipse.org/equinox/drops/R-4.31-202402290520/

merks:
@elsazac @MohananRahul 

I made the changes and restarted your release build to start this build:

https://ci.eclipse.org/releng/job/CBIaggregator/894

The job completed and the validation is running now:

https://ci.eclipse.org/releng/view/Publish%20to%20Maven/

So I think this  is on track now.

merks:
😱 

I see that validation failed now. 

>14:49:03 [ERROR] Failed to execute goal org.apache.maven.plugins:maven-dependency-plugin:3.6.0:get (default-cli) on project standalone-pom: Couldn't download artifact: org.eclipse.aether.resolution.DependencyResolutionException: The following artifacts could not be resolved: xml-apis:xml-apis-ext:jar:1.0.0.v20230923-0644 (absent): Could not find artifact xml-apis:xml-apis-ext:jar:1.0.0.v20230923-0644 in eclipse.maven.central.mirror (https://repo.eclipse.org/content/repositories/maven_central/) -> [Help 1]


 I will need to investigate and that might take significant time.

In the meantime, I will kick off a snapshot job to see if that fails differently.

merks:
😱 

I see that validation failed now. 

[...]

PR #1885, opened 2024-03-08
Publish 4.31 to Maven central

elsazac:
Update SDK4Mvn.aggr and properties.sh file for publishing 4.31 to maven central.

Refs: https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/issues/1882

cc/ @MohananRahul


merks:
FYI, these intermittent failures are super annoying: 😱 

https://ci.eclipse.org/releng/job/CBIaggregator/

I think I might finally force the aggregator to avoid using the java client like I've done for SimRel:
```
-Dorg.eclipse.ecf.provider.filetransfer.excludeContributors=org.eclipse.ecf.provider.filetransfer.httpclientjava
```

elsazac:
@merks Thanks for looking into this.I can try this but I couldn't find an appropriate location in the configuration to add this flag.

PR #1893, opened 2024-03-10
Aggregate 4.32

merks:


merks:
@MohananRahul @elsazac 

I assume that the Maven publishing for 4.31 is on track and so we need to change this for 4.32.  Otherwise the snap shot publishing keeps doing 4.31 and keeps failing validation:

https://ci.eclipse.org/releng/view/Publish%20to%20Maven/job/RepositoryValidator/

PR #1900, opened 2024-03-13
Revert "Prevent two PRs for same update from dependabot"

akurtakov:
Reverts eclipse-platform/eclipse.platform.releng.aggregator#1643

This broke dependabot opening PRs for maven plugins in eclipse-platform-parent


akurtakov:
Push as verification build will not execute dependabot and this actually reverts to the old working state, although it used to create duplications.

PR #1913, opened 2024-03-18
Remove unused tycho-eclipserun-plugin configuration

HannesWell:
It is probably a left-over from the old setup to run the API-analysis, which is now not necessary anymore.
@laeubi, @akurtakov I guess it is also not used implicitly anymore, is it?


akurtakov:
No referenced to the plugin so it looks like you're correct. Merging.

PR #1915, opened 2024-03-19
Update P2 director documentation with new option -addJREIU

raghucssit:
A new option is introduced by
https://github.com/eclipse-equinox/p2/issues/484.


merks:
@raghucssit 

Thanks for your responsiveness addressing issues in the review, even the really picky ones and update documentation.  🏆

Issue #577, opened 2022-09-19
Native support for Eclipse SDKs on Windows Arm64


chirontt:
Update: now that adding support for Windows on Arm64 is under way for various Eclipse repositories, here's the central TODO list to keep track of various tasks done and/or under way:
* [x] java-native-access/jna#1466
* [x] add support for Windows on Arm64 to `eclipse.platform.swt` submodule
  - [x] eclipse-platform/eclipse.platform.swt#1048
  - [x] eclipse-platform/eclipse.platform.swt#1045
  - [x] eclipse-platform/eclipse.platform.swt#1137
  - [x] eclipse-platform/eclipse.platform.swt#1158. This is temporary measure to have successful build; to be reverted later.
    - [x] eclipse-platform/eclipse.platform.swt#1222
  - [x] eclipse-platform/eclipse.platform.swt#1172
* [x] add support for Windows on Arm64 to `equinox` submodule
  - [x] eclipse-equinox/equinox#559
  - [x] eclipse-equinox/equinox#564
  - [x] eclipse-equinox/equinox#575
  - [x] eclipse-equinox/equinox#603
  - [x] eclipse-equinox/equinox#621
* [x] add support for Windows on Arm64 to `rt.equinox.p2` submodule
  - [x] eclipse-equinox/p2#492
* [x] add support for Windows on Arm64 to `eclipse.platform` submodule
  - [x] eclipse-platform/eclipse.platform#1311
  - [x] eclipse-platform/eclipse.platform#1350
  - [x] eclipse-platform/eclipse.platform#1351
  - [x] eclipse-platform/eclipse.platform#1394
* [x] add support for Windows on Arm64 to `eclipse.platform.ui` submodule
  - [x] eclipse-platform/eclipse.platform.ui#1789
  - [x] eclipse-platform/eclipse.platform.ui#1787
* [x] add support for Windows on Arm64 to `eclipse.platform.releng.aggregator`
  - [x] eclipse-platform/eclipse.platform.releng.aggregator#1942
  - [x] eclipse-platform/eclipse.platform.releng.aggregator#1953
  - [x] eclipse-platform/eclipse.platform.releng.aggregator#1917
  - [x] eclipse-platform/eclipse.platform.releng.aggregator#2017
* [x] eclipse-platform/www.eclipse.org-eclipse#169
* [x] eclipse-packaging/packages#162

---
[...]

PR #1917, opened 2024-03-19
Add win32.aarch64 target environment to eclipse-platform-parent/pom.xml

HannesWell:
This is (probably) required to support building the Eclipse-SDK on Windows for ARM.

See https://github.com/eclipse-platform/eclipse.platform.swt/pull/1045#discussion_r1529883121


HannesWell:
Looks like we need a launcher for win32.aarch64 first.

laeubi:
> Looks like we need a launcher for win32.aarch64 first.

Yes that would be the next logical step, @chirontt would you mind to open a PR against equinox to bring in the launcher parts?

laeubi:
btw something looks strange here:

Could not resolve plugin org.eclipse.equinox.launcher.win32.win32.aarch64_null;

chirontt:
> > Looks like we need a launcher for win32.aarch64 first.
> 
> Yes that would be the next logical step, @chirontt would you mind to open a PR against equinox to bring in the launcher parts?

eclipse-equinox/equinox#559

chirontt:
> > Could not resolve plugin org.eclipse.equinox.launcher.win32.win32.aarch64_null;
> 
> looks like there is no version and then Tycho prints it as `null`

It's too early for this submit - all submodules (well, 5 of them) should have `win32.aarch64` support first, before this aggregator repo can begin the `win32.aarch64` support.

I think I will have to create a TODO list of items for all the various PRs for various repos, here in this aggregator as a central space. How should I go about with this TODO list? via an `Issue`? or via `Discussion`? or via a `Pull Request`?

laeubi:
I think you can use this issue here:
- https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/issues/577

you can simply edit your inital posting and add a todo list.

HannesWell:
> @HannesWell as [discussed](https://github.com/eclipse-platform/eclipse.platform.ui/pull/1787#issuecomment-2030305378), here's my branch to merge with this PR, thanks:
> 
> https://github.com/chirontt/eclipse.platform.releng.aggregator/tree/add-win32.aarch64-environment

I have now added your commit to this PR but removed the addition of the `org.eclipse.core.resources.win32.aarch64` and `org.eclipse.core.filesystem.win32.aarch64` fragments since they are not (yet) available and adjusted the commit message since your commit is then mainly about publishing the new artifacts and products.

Unfortunately this build currently fails due to https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/issues/2014.

@iloveeclipse FYI, without this there will probably be comparator errors again, similar to https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/issues/2011 since the Equinox binaries were re-compiled, but this change touches `rcp.config`.
Actually I wanted to submit this today to avoid such error, but https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/issues/2014 is blocking that unfortunately.

PR #1919, opened 2024-03-20
Make Java 21 the primary JVM to execute the maven build

laeubi:
Java 21 is the next LTS and projects are allowed to use Java 21 in this release, therefore it should become the new primary Java version used to execute the build.

This also will allow us to catch any possible issues with Java 21 used as build VM.

### DIFF
--- a/.github/workflows/mavenBuild.yml
+++ b/.github/workflows/mavenBuild.yml
@@ -56,10 +56,12 @@ jobs:
           8
           11
           17
+          21
         mvn-toolchain-id: |
           JavaSE-1.8
           JavaSE-11
           JavaSE-17
+          JavaSE-21
         distribution: 'temurin'
         cache: maven
     - name: Set up Maven
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -10,7 +10,7 @@ pipeline {
 	}
 	tools {
 		maven 'apache-maven-latest'
-		jdk 'temurin-jdk17-latest'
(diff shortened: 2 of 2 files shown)


akurtakov:
Rebased to see whether the failuse is random one or a general one to be investigated.

merks:
@akurtakov  @laeubi 

I think this need to be reverted.  All builds fail since this change:

https://ci.eclipse.org/platform/job/eclipse.platform.releng.aggregator/job/master/

The build will continue to fail until JDT addresses this issue:

https://github.com/eclipse-jdt/eclipse.jdt.core/issues/2217

Okay?

laeubi:
Why not wait "until JDT addresses this issue" it look easy enough to fix instead of reverting, also I think not *all* builds fail, maybe aggregator builds and JDT builds but not pde, platform(ui), equinox, p2 and so on...

merks:
Because:

1.  No verification build works right now so **nothing else can be verified**.  
2. Until JDT is fixed, it will simply keep **telling us about the same problem over and over.**  

I.e., the change has bought the only value it can bring right now  with the ongoing result that the build itself provides no value for the purpose for which it was designed, which is to verify changes.

akurtakov:
https://github.com/eclipse-jdt/eclipse.jdt.core/pull/2214 should  fix the issue. Assuming it finishes without issues today I'll merge it and we should move on. Please give it some time for issues to be fixed.

PR #1931, opened 2024-03-21
Improved TouchBundles to automatically bump affected bundle versions

iloveeclipse:
See https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/issues/1923

### DIFF
--- a/eclipse.platform.releng.prereqs.sdk/eclipse-sdk-prereqs.target
+++ b/eclipse.platform.releng.prereqs.sdk/eclipse-sdk-prereqs.target
@@ -207,6 +207,16 @@
 			  </dependency>
 		  </dependencies>
 	  </location>
+	  <location includeDependencyDepth="infinite" includeDependencyScopes="compile" includeSource="true" label="JGit" missingManifest="error" type="Maven">
+		  <dependencies>
+			  <dependency>
+				  <groupId>org.eclipse.jgit</groupId>
+				  <artifactId>org.eclipse.jgit</artifactId>
+				  <version>6.9.0.202403050737-r</version>
+				  <type>jar</type>
+			  </dependency>
+		  </dependencies>
+	  </location>
 	  <location includeDependencyDepth="none" includeDependencyScopes="compile" includeSource="true" label="Jetty and servlet API" missingManifest="error" type="Maven">
 		  <dependencies>
 			  <dependency>
--- a/eclipse.platform.releng/bundles/org.eclipse.releng.tests/META-INF/MANIFEST.MF
+++ b/eclipse.platform.releng/bundles/org.eclipse.releng.tests/META-INF/MANIFEST.MF
@@ -2,12 +2,14 @@ Manifest-Version: 1.0
 Bundle-ManifestVersion: 2
 Bundle-Name: %Plugin.name
 Bundle-SymbolicName: org.eclipse.releng.tests;singleton:=true
-Bundle-Version: 3.6.400.qualifier
+Bundle-Version: 3.6.500.qualifier
 Eclipse-BundleShape: dir
 Bundle-Vendor: %Plugin.providerName
 Bundle-Localization: plugin
 Require-Bundle: org.eclipse.core.runtime,
- org.junit
+ org.junit,
+ org.eclipse.jgit,
+ slf4j.api;resolution:=optional
 Export-Package: org.eclipse.releng.tests
 Bundle-RequiredExecutionEnvironment: JavaSE-17
 Bundle-ClassPath: .
--- a/eclipse.platform.releng/bundles/org.eclipse.releng.tests/src/org/eclipse/releng/tests/VersionBump.java
+++ b/eclipse.platform.releng/bundles/org.eclipse.releng.tests/src/org/eclipse/releng/tests/VersionBump.java
@@ -1,0 +1,172 @@
+/*******************************************************************************
+ * Copyright (c) 2024 Andrey Loskutov <loskutov@gmx.de> and others.
+ *
+ * This program and the accompanying materials
+ * are made available under the terms of the Eclipse Public License 2.0
+ * which accompanies this distribution, and is available at
+ * https://www.eclipse.org/legal/epl-2.0/
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ *
+ * Contributors:
+ * Andrey Loskutov <loskutov@gmx.de> - initial API and implementation
+ *******************************************************************************/
+package org.eclipse.releng.tests;
+
+import java.io.IOException;
+import java.nio.charset.StandardCharsets;
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.nio.file.Paths;
+import java.util.Objects;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+
+import org.eclipse.jgit.errors.RevisionSyntaxException;
+import org.eclipse.jgit.lib.ObjectId;
+import org.eclipse.jgit.lib.ObjectLoader;
+import org.eclipse.jgit.lib.ObjectReader;
+import org.eclipse.jgit.lib.Repository;
+import org.eclipse.jgit.revwalk.RevCommit;
+import org.eclipse.jgit.revwalk.RevWalk;
+import org.eclipse.jgit.storage.file.FileRepositoryBuilder;
+import org.eclipse.jgit.treewalk.TreeWalk;
+import org.osgi.framework.Version;
+
+public class VersionBump {
+
+	/**
+	 * Checks if the version bump needed for given bundle, and if needed, bumps
+	 * micro segment in manifest and pom
+	 *
+	 * @param args first argument is bundle root path to check, second version SDK
+	 *             release tag like "R4_31"
+	 */
+	public static void main(String[] args) {
+		if (args.length != 2) {
+			System.out.println("Arguments: first is path to the bundle root, \n" + "second is the release tag");
+			System.exit(1);
+		}
+		run(Paths.get(args[0]), args[1]);
+	}
+
+	/**
+	 *
+	 * @param bundleRoot
+	 * @param releaseTag
+	 * @return true if manifest and pom (if exists) were updated
+	 */
+	public static boolean run(Path bundleRoot, String releaseTag) {
+		if (!Files.isDirectory(bundleRoot)) {
+			System.err.println("Bundle root is not a directory: " + bundleRoot);
+			return false;
+		}
+		if (releaseTag == null || releaseTag.isBlank()) {
+			System.err.println("Release tag missing");
+			return false;
+		}
+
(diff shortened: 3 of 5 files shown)


eclipse.platform.releng/bundles/org.eclipse.releng.tests/META-INF/MANIFEST.MF
```diff
@@ -2,12 +2,14 @@ Manifest-Version: 1.0
...
 Bundle-SymbolicName: org.eclipse.releng.tests;singleton:=true
-Bundle-Version: 3.6.400.qualifier
+Bundle-Version: 3.6.500.qualifier
 Eclipse-BundleShape: dir
 Bundle-Vendor: %Plugin.providerName
 Bundle-Localization: plugin
 Require-Bundle: org.eclipse.core.runtime,
- org.junit
+ org.junit,
+ org.eclipse.jgit,
```
iloveeclipse, line 11: @laeubi : is there a way to make the build happy - currently it can't find jgit in the target platform (I assume). Is this something like extra dependency in pom?

laeubi: You need to add jgit here (either P2 or maven):
https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/blob/master/eclipse.platform.releng.prereqs.sdk/eclipse-sdk-prereqs.target

laeubi: by the way there is already a [bump-versions](https://tycho.eclipseprojects.io/doc/latest/tycho-versions-plugin/bump-versions-mojo.html) mojo at Tycho just in case one wants to enhance that in some way.

eclipse.platform.releng/bundles/org.eclipse.releng.tests/src/org/eclipse/releng/tests/VersionBump.java
```diff
@@ -0,0 +1,172 @@
+/*******************************************************************************
+ * Copyright (c) 2023 Andrey Loskutov <loskutov@gmx.de> and others.
```
laeubi, line 2:
```suggestion
 * Copyright (c) 2024 Andrey Loskutov <loskutov@gmx.de> and others.
```

iloveeclipse:
@laeubi : any idea why license check fails?

laeubi:
```
Error:    Software being installed: org.eclipse.releng.tests 3.6.500.qualifier
Error:    Missing requirement: org.eclipse.jgit 6.9.0.202403050737-r requires 'java.package; com.googlecode.javaewah [1.1.6,2.0.0)' but it could not be found
Error:    Cannot satisfy dependency: org.eclipse.releng.tests 3.6.500.qualifier depends on: osgi.bundle; org.eclipse.jgit 0.0.0
```

eclipse.platform.releng.prereqs.sdk/eclipse-sdk-prereqs.target
```diff
@@ -205,6 +205,12 @@
 				  <version>1.9</version>
 				  <type>jar</type>
 			  </dependency>
+			  <dependency>
+				  <groupId>org.eclipse.jgit</groupId>
```
trancexpress, line 209: Fails to compile:

```
org.apache.maven.lifecycle.LifecycleExecutionException: Cannot resolve dependencies of project org.eclipse.releng:org.eclipse.releng.tests:eclipse-test-plugin:3.6.500-SNAPSHOT
 with context {osgi.os=linux, org.eclipse.update.install.features=true, osgi.arch=x86_64, org.eclipse.update.install.sources=true, osgi.ws=gtk, org.eclipse.jdt.buildtime=true}
  Software being installed: org.eclipse.releng.tests 3.6.500.qualifier
  Missing requirement: org.eclipse.releng.tests 3.6.500.qualifier requires 'osgi.bundle; org.eclipse.jgit 0.0.0' but it could not be found
```

eclipse.platform.releng.prereqs.sdk/eclipse-sdk-prereqs.target
```diff
@@ -205,6 +205,12 @@
 				  <version>1.9</version>
 				  <type>jar</type>
 			  </dependency>
+			  <dependency>
```
laeubi, line 208: you need to declare dependencies or we need to put it into a location that includes dependencies automatically.

eclipse.platform.releng.prereqs.sdk/eclipse-sdk-prereqs.target
trancexpress: Oddly I didn't get this by cherry-picking...

laeubi:
I think it needs this one: https://mvnrepository.com/artifact/com.googlecode.javaewah/JavaEWAH/1.2.3

eclipse.platform.releng.prereqs.sdk/eclipse-sdk-prereqs.target
iloveeclipse: Simeon, you can just open the project in the IDE and run from IDE, as we have jgit there.

iloveeclipse:
> ```
> Error:  Cannot resolve project dependencies:
> Error:    Software being installed: org.eclipse.releng.tests 3.6.500.qualifier
> Error:    Missing requirement: org.eclipse.jgit 6.9.0.202403050737-r requires 'java.package; com.googlecode.javaewah [1.1.6,2.0.0)' but it could not be found
> Error:    Cannot satisfy dependency: org.eclipse.releng.tests 3.6.500.qualifier depends on: osgi.bundle; org.eclipse.jgit 0.0.0
> ```

Does it mean, all transitive jgit dependencies must be added to the target?

laeubi:
> Does it mean, all transitive jgit dependencies must be added to the target?

Currently the location is configured to not include transitive dependencies, but It seems jgit does not have many so adding https://mvnrepository.com/artifact/com.googlecode.javaewah/JavaEWAH/1.2.3 could probably be enough.

As an alternative don't add jgit to an existing but a new location and choose infinite as dependency depth.

merks:
It makes a little uncomfortable that we add jgit and its transitive closure to the target platform and hope that none of those dependencies leak into the SDK product we ship:

image

Just saying...

iloveeclipse:
> It makes a little uncomfortable that we add jgit and its transitive closure to the target platform and hope that none of those dependencies leak into the SDK product we ship

Is there any other way? The code is only needed to be in a repo so one can trigger the task from a command line or IDE if needed. So it is not a strong bundle requirement in a "usual" understanding, it just needs to be compiled.

akurtakov:
I don't understand why is jgit in the target platform at all. It looks like remnant of the past to me.

iloveeclipse:
> > > I don't understand why is jgit in the target platform at all. It looks like remnant of the past to me.
> > 
> > 
> > It is needed for current PR :-)
> 
> I would prefer if the target platform is kept for actual sdk content . The script you try to add seems to fit better in https://github.com/eclipse-platform/eclipse.platform.releng.buildtools/ , 

OK, will check that.

>  note that JGit is already there and used.

How is that one building without jgit in platform?

merks:
For that repo's Oomph setup, egit is added to the target platform like this:

image

I don't know how folks manually work with that repo.

akurtakov:
> How is that one building without jgit in platform?

It has it's own build https://ci.eclipse.org/releng/job/EclipseBuildTools/, https://ci.eclipse.org/platform/job/eclipse.platform.releng.buildtools/ solely configured via https://github.com/eclipse-platform/eclipse.platform.releng.buildtools/blob/master/pom.xml to be standalone and not interfere with SDK build,  it actually defines its own target platform in pure maven way.

trancexpress:
https://github.com/eclipse-pde/eclipse.pde/pull/1209 is now updated with the changes made by the improved `TouchBundles`. Running `TouchBundles` from command line now requires a few jars on the classpath (whereas before it could be compiled and executed on its own):

```
java -cp ./eclipse.platform.releng/bundles/org.eclipse.releng.tests/target/org.eclipse.releng.tests-3.6.500-SNAPSHOT.jar:./org/eclipse/jgit/org.eclipse.jgit/6.9.0.202403050737-r/org.eclipse.jgit-6.9.0.202403050737-r.jar:./org/slf4j/slf4j-api/2.0.12/slf4j-api-2.0.12.jar:./org/osgi/org.osgi.framework/1.8.0/org.osgi.framework-1.8.0.jar org.eclipse.releng.tests.TouchBundles /data/tmp/comparator_issues/ /data/git/eclipse/eclipse.platform.releng.aggregator/ "https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/issues/1923"
```

laeubi:
By the way since Java 11 a single java file can even be run without compiling:

https://openjdk.org/jeps/330

iloveeclipse:
> [eclipse-pde/eclipse.pde#1209](https://github.com/eclipse-pde/eclipse.pde/pull/1209) is now updated with the changes made by the improved `TouchBundles`. Running `TouchBundles` from command line now requires a few jars on the classpath (whereas before it could be compiled and executed on its own):
> 
> ```
> java -cp ./eclipse.platform.releng/bundles/org.eclipse.releng.tests/target/org.eclipse.releng.tests-3.6.500-SNAPSHOT.jar:./org/eclipse/jgit/org.eclipse.jgit/6.9.0.202403050737-r/org.eclipse.jgit-6.9.0.202403050737-r.jar:./org/slf4j/slf4j-api/2.0.12/slf4j-api-2.0.12.jar:./org/osgi/org.osgi.framework/1.8.0/org.osgi.framework-1.8.0.jar org.eclipse.releng.tests.TouchBundles /data/tmp/comparator_issues/ /data/git/eclipse/eclipse.platform.releng.aggregator/ "https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/issues/1923"
> ```

As said, you can run it from IDE without compilation, just import the project where the code is located.
Anyway, I will move the code to other repo later on because of dependencies.

trancexpress:
> By the way since Java 11 a single java file can even be run without compiling:
> 
> https://openjdk.org/jeps/330

That worked for `TouchBundles.java` prior to changes in this PR. As commented above; now it needs `jgit` and the `jgit` code needs other things.

iloveeclipse:
> > By the way since Java 11 a single java file can even be run without compiling:
> > https://openjdk.org/jeps/330
> 
> That worked for `TouchBundles.java` prior to changes in this PR. As commented above; now it needs `jgit` and the `jgit` code needs other things.

Sure. But now it does more as before.

laeubi:
Just theoretically speaking one would but the code in **one** java file and only will call it either form commandline or IDE where the dependencies are there, one would not need to *compile* the code as part of the regular build....

iloveeclipse:
Moved code to https://github.com/eclipse-platform/eclipse.platform.releng.buildtools/pull/61

PR #1942, opened 2024-03-26
Adapt to equinox.security.win32 renaming and adjust arch requirements

HannesWell:


HannesWell:
The verification build of this PR cannot succeed without https://github.com/eclipse-equinox/equinox/pull/564 being part of an I-build, but without this an I-build cannot pass after https://github.com/eclipse-equinox/equinox/pull/564 has been submitted. So we have to submit this, without verification.

PR #1953, opened 2024-04-02
The SDK4Mvn.aggr needs a win32/win32/aarch64 configuration

merks:


HannesWell:
Thank you Ed.
For reference, this is part of https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/issues/577.

CC @chirontt.

PR #1986, opened 2024-04-19
Update bouncycastle to 1.78.1

merks:


merks:
The build fails like this:
```
Error:    Software being installed: org.eclipse.jdt.feature.group 3.19.500.qualifier
Error:    Missing requirement: bcpg 1.78.1 requires 'java.package; org.bouncycastle.asn1.cryptlib [1.78.1,1.79.0)' but it could not be found
Error:    Cannot satisfy dependency: org.eclipse.equinox.p2.engine 2.10.100.qualifier depends on: java.package; org.bouncycastle.openpgp 1.65.0
Error:    Cannot satisfy dependency: org.eclipse.jdt.debug.ui 3.13.400.qualifier depends on: osgi.bundle; org.eclipse.ui.ide [3.5.0,4.0.0)
Error:    Cannot satisfy dependency: org.eclipse.jdt.feature.group 3.19.500.qualifier depends on: org.eclipse.equinox.p2.iu; org.eclipse.jdt.debug.ui 0.0.0
Error:    Cannot satisfy dependency: org.eclipse.ui.ide 3.22.200.qualifier depends on: osgi.bundle; org.eclipse.equinox.p2.engine [2.8.0,3.0.0)
```
So we need to add the bundle that provides the org.bouncycastle.asn1.cryptlib  package, i.e.,  bcutil-jdk18on.

merks:
@iloveeclipse  @akurtakov  @tomaswolf 

FYI, it appears necessary to add a dependency on bcutil-jdk18on to use the 1.78.1 version of bcpg.

Note that for testing purposes, I created an installation with only 1.78.1 versions and that resolved/installed fine:

image

merks:
I will assume later today that no one has any concerns about me merging these changes.

akurtakov:
I have no concerns.

iloveeclipse:
Latest Egit still works on this platform ?

merks:
Yes, I was able to clone:

image

I'm not sure that exercises the bouncycastle functionality; I'm not sure what does.  I tried updating their target platform to run the tests.  That worked, but I had to add a package import to eliminate a class not found exception:

image

My understanding is that this is only used to to GPG signing of commits, which is not something the Eclipse SDK development process uses.

PR #1997, opened 2024-04-25
[I-Build-Tests] Simplify tool provisioning for Linux tests

HannesWell:
Leverage the standard Jenkins pipeline tools section to set up the JDK and ANT instead of downloading the JDK 'manually'.
This uses the JDKs provided in the EF Jenkins instances as listed in
- https://wiki.eclipse.org/Jenkins#JDK

@akurtakov am I right that the previous, more complicated implementation is a left-over from the more complicated past or is there a compelling reason to keep it as it is?
I tested the changes for Java-17 and 22 by replaying the corresponding jobs with the changes applied and AFAICT everything looked good.

PR #1999, opened 2024-04-29
[I-Build-Tests] Fix breakage of fundamental commands by tools-sections

HannesWell:
Declaring a jdk and ant the usual way in the 'tools' section, because of unknown reasons, breaks the usage of fundamental commands like `xvnc`, `pkill` and `sh`.

See https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/pull/1997#issuecomment-2080385005

### DIFF
--- a/JenkinsJobs/AutomatedTests/I_unit_cen64_gtk3_java17.groovy
+++ b/JenkinsJobs/AutomatedTests/I_unit_cen64_gtk3_java17.groovy
@@ -85,9 +85,11 @@ spec:
 
   stages {
       stage('Run tests'){
-          tools {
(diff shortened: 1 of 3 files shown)


HannesWell:
Replayed all Linux test jobs with this test applied and all looked good again.
I'll submit this and start another I-build. The verification build does not test this at all.

HannesWell:
> I'll submit this and start another I-build.

Deployed the changed job definitions and just started:
https://ci.eclipse.org/releng/job/Builds/job/I-build-4.32/84/

If it is still not working as expected, I can look into again it this evening.

PR #2004, opened 2024-04-29
[I-Build-Tests] Define Mac and Windows jobs as Jenkins-pipeline

HannesWell:
This allows to unify the job definition with the test-jobs for Linux and also allows to simplify some parts of the job a lot.

It also removes the `testSuite` parameter only available in mac tests, since anybody that wants to run another suite than all can just `Replay` the build in the Jenkins web-interface and adjust the suite as desired.

Especially the Windows pipeline needs some more local testing, e.g. with regards to spanning a command over multiple lines, but nevertheless I still would like to gather some early feedback.

### DIFF
--- a/JenkinsJobs/AutomatedTests/I_unit_mac64_java17.groovy
+++ b/JenkinsJobs/AutomatedTests/I_unit_mac64_java17.groovy
@@ -5,80 +5,40 @@ for (STREAM in STREAMS){
   def MAJOR = STREAM.split('\\.')[0]
   def MINOR = STREAM.split('\\.')[1]
 
-  job('AutomatedTests/ep' + MAJOR + MINOR + 'I-unit-mac64-java17'){
+  pipelineJob('AutomatedTests/ep' + MAJOR + MINOR + 'I-unit-mac64-java17'){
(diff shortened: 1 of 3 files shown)


HannesWell:
With the testing I did in my local Jenkins instance I'm relatively confident that this change is fine (at least syntax-wise).

I'll submit this now, deploy the new test-job definitions and trigger the affected ones to see if they at least start up as expected. Tonight's I-build will then verify if everything is completely working as expected.

HannesWell:
In order to re-generate the jobs with another type (`pipelineJob` instead of `job`) no item with the same id was permitted to exist before. Therefore I renamed the existing configuration and appended the suffix `-old`.
- https://ci.eclipse.org/releng/job/AutomatedTests/job/ep432I-unit-mac64-java17-old/
- https://ci.eclipse.org/releng/job/AutomatedTests/job/ep432I-unit-macM1-java17-old/
- https://ci.eclipse.org/releng/job/AutomatedTests/job/ep432I-unit-win32-java17-old/

I assume the execution history is not that important and we can delete them as soon as the new configuration has proofen itself fully functional.

Until now I have already verified that the jobs start as expected and as planned, tonight's I-build will verify the entire pipeline.

PR #2005, opened 2024-04-30
[I-Build-Tests] Merge multiple definitions into one per OS

HannesWell:
This is currently a draft since I want to await the verification that the latest changes work as expected.
Furthermore the windows job could be prepared for a test on Windows on ARM too.

@akurtakov what do you think?


JenkinsJobs/AutomatedTests/I_unit_mac_java17.groovy
```diff
@@ -1,11 +1,15 @@
 def config = new groovy.json.JsonSlurper().parseText(readFileFromWorkspace('JenkinsJobs/JobDSL.json'))
 def STREAMS = config.Streams
 
+def ARCHS = ['aarch64', 'x86_64']
+def ARCHS_AGENT_LABEL = ['aarch64': 'nc1ht-macos11-arm64', 'x86_64': 'nc1ht-macos11-arm64']
```
HannesWell, line 5: I wonder if it is intentional that the mac-tests for both arches, x86_64 and arm64 run on the arm-machines for over a year now?
And if it is intentional, is there any chance to run it on the x86_64 machine again? That would also improve the distribution of the I-build test execution load.

akurtakov:
I like the simplification a lot. Feel free to push this one.
I have to defer Mac questions to @sravanlakkimsetti

JenkinsJobs/AutomatedTests/I_unit_mac_java17.groovy
sravanlakkimsetti: This is done to optimize resource packs.

HannesWell: > Actually we don't have a x86_84 test machine. Because of this we are using arm64(m1) machine to test both architectures.

What's about https://ci.eclipse.org/releng/computer/b9h15%2Dmacos11%2Dx86%5F64/? That it at least used to build the SWT and Equinox native binaries for MacOS x86_64.

HannesWell:
> I like the simplification a lot. Feel free to push this one.

Great. I have a few more minor adjustments, that I want to add and that will also prepare a possible Windows on ARM I-build.

JenkinsJobs/AutomatedTests/I_unit_mac_java17.groovy
HannesWell: But I'll try a replay of the mac64 I-build test this evening where I change the node label to see if that works as well.

sravanlakkimsetti: There are couple of reasons.
1. I remember  https://ci.eclipse.org/releng/computer/b9h15%2Dmacos11%2Dx86%5F64/ as not having GUI available. I may be wrong here.
2. Don't want to block swt and Launcher builds as the machine is being used for test runs(test run can go up to 4 hours)

HannesWell: > There are couple of reasons.
> 1. I remember  https://ci.eclipse.org/releng/computer/b9h15%2Dmacos11%2Dx86%5F64/ as not having GUI available. I may be wrong here.

Do you think this could be activated?


> 2. Don't want to block swt and Launcher builds as the machine is being used for test runs(test run can go up to 4 hours) 

Yes, but the mac-aarch64 machine is also used to build the natives of SWT and equinox while at the same time being used for I-build tests. The current situation makes it even worse since the mac-aarch64 machine is now blocked by the tests for two architectures and therefore for twice the time.

sravanlakkimsetti: if https://ci.eclipse.org/releng/computer/b9h15%2Dmacos11%2Dx86%5F64/ can be activated that would be great help. I would rather use this for building the SWT and equinox to unblock other machine for testing.

HannesWell: I just tested running the mac-x86_64 test on https://ci.eclipse.org/releng/computer/b9h15%2Dmacos11%2Dx86%5F64/ by replaying with a changed label but it currently already fails because ANT is not available:
https://ci.eclipse.org/releng/job/AutomatedTests/job/ep432I-unit-mac64-java17/14/console

So probably there is a bit more to set up. I'll leave that task for a future PR but in general we should implement that to leverage the available resources better.

> Please check with foundation on this one.

Yes, I have already created https://gitlab.eclipse.org/eclipsefdn/helpdesk/-/issues/4231 some time ago to avoid blocking of native building tasks by I-build tests. The other way round is less problematic because the native build jobs usually complete within one, two minutes or even less, so the I-build tests are not really blocked (one won't really notice if they complete two minutes later).
I also had a little chat with Frederic Gurr about it and I think we will make progress on that topic in the future.

akurtakov:
What is the status of this one?

HannesWell:

Still the same as in my previews comment and I first wanted to to some smoke testing with my local Jenkins.
Since M3 is approaching, I'm currently working on getting the most important bugs fixed or some last enhancements submitted.
I plan to continue this in the upcoming quite-periods, but maybe already at the weekend.

akurtakov:
What is the status of this one? I see many conflicts now.

HannesWell:
> What is the status of this one? I see many conflicts now.

I rebased it and resolved the conflicts. I definitively plan to finish this, but the final try out with my local Jenkins (that at least it leads to the desired job configs) is still pending. If time permits I'll do it at next weekend or the one after it, after EclipseCon.
Completing it at the weekend also has the advantage that in case of problems one can take a bit more time to fix it, since the development is usually not so buy. :)

HannesWell:
All my local testing looked good. All test jobs are still generated with the expected configuration.
The only change is that the name of the mac jobs is preserved as it is in order to avoid adjustments of down-stream scripts for now.

Since the verification build does not test this I'll submit this now so that tonight's I-build can run verify this.

HannesWell:
With https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/pull/2477 the job re-generation has now worked.

The I-build has started already, but since the test-jobs are kicked in its end, that I-build should still test the new configuration (which should actually not changed).

HannesWell:
Due to a second follow-up, https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/pull/2478, the I-build was now aborted and restarted to be triggered completely with the newly generated jobs being available.

PR #2009, opened 2024-05-06
Update objenesis to 3.4

merks:


merks:
@akurtakov  @laeubi 

I won't have time to merge this when the build is done.  If one of you would be so kind, that will be appreciated.

merks:
One can do so much by phone even on a speeding train. 🚂

PR #2017, opened 2024-05-08
[I-Build] Copy built products for Windows on Arm to publication area

HannesWell:
This last step was forgotten in PR https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/pull/1917, activating the publication of Eclipse SDK/platform products for the win32.aarch64 environment.

Part of https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/issues/577


HannesWell:
The verification build does not test this change.

In general I think it would be good to just copy the products for all platforms without the need to list them one-by-one.

HannesWell:
Reverted the addition of win32.aarch64 in `mb310_gatherEquinoxParts.sh` because currently the OSGi starter-kit is only built for x86_64 arch for all OS.

PR #812, opened 2023-01-19
Update Y and P builds for BETA_JAVA20 branch

SDawley:
First pass at adding Y and P builds for Java 20

@akurtakov
I probably missed something.

PR #2019, opened 2024-05-09
Remove unused duplicates of getEBuilder.xml and runTests2.xml

HannesWell:
Only the identical files in the 'production' sibling folder are used.

Those two duplicates where added in https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/pull/812, but to me it looks like they never have been used.

@akurtakov can you tell more about that or know anybody that who knows more?


akurtakov:
Honestly, I don't remember anything. So go on and remove these and if anything fails we can fix it after that. There are so many such things that I don't believe we can get any sanity without trying out.

HannesWell:
> Honestly, I don't remember anything. So go on and remove these and if anything fails we can fix it after that. There are so many such things that I don't believe we can get any sanity without trying out.

Yes.
In case, contrary to the expectation, it is used anywhere, in that places can simply reference the 'original' in the repo or deployed to https://download.eclipse.org/eclipse/relengScripts/production/testScripts/hudsonBootstrap

PR #2021, opened 2024-05-13
Update coactions to use Node 20

fedejeanne:
GH Actions transitioned to Node 20 in September 2023 and all projects should transition too (see:
https://github.blog/changelog/2023-09-22-github-actions-transitioning-from-node-16-to-node-20/)

This gets rid of the warnings in the runs of GH Actions.

### DIFF
--- a/.github/workflows/codeQLworkflow.yml
+++ b/.github/workflows/codeQLworkflow.yml
@@ -76,7 +76,7 @@ jobs:
       with:
         maven-version: ${{ inputs.mavenVersion }}
     - name: Build with Maven
-      uses: coactions/setup-xvfb@b6b4fcfb9f5a895edadc3bc76318fae0ac17c8b3 # v1.0.1
+      uses: coactions/setup-xvfb@6b00cf1889f4e1d5a48635647013c0508128ee1a
       with:
        run: >- 
         mvn --batch-mode -V -U
--- a/.github/workflows/mavenBuild.yml
+++ b/.github/workflows/mavenBuild.yml
@@ -76,7 +76,7 @@ jobs:
         target: .github/matcher
     - run: echo "::add-matcher::.github/matcher/${{ steps.api-tools-matcher.outputs.filename }}"
     - name: Build with Maven
-      uses: coactions/setup-xvfb@b6b4fcfb9f5a895edadc3bc76318fae0ac17c8b3 # v1.0.1
+      uses: coactions/setup-xvfb@6b00cf1889f4e1d5a48635647013c0508128ee1a
       with:
        run: >- 
         mvn --batch-mode -V -U


fedejeanne:
@mickaelistria , @jukzi I see you touched those lines previously so maybe you could review this?

FTR I compared the [old version](https://github.com/coactions/setup-xvfb/blob/b6b4fcfb9f5a895edadc3bc76318fae0ac17c8b3/action.yml) of the script with the [new version](https://github.com/coactions/setup-xvfb/blob/90473c3ebc69533a1a6e0505c36511b69c8c3135/action.yml) and they are almost identical: only the Node version changed.

Also, the latest version of the action is still `v1.0.1` according to [the marketplace](https://github.com/marketplace/actions/setup-xvfb)

.github/workflows/codeQLworkflow.yml
```diff
@@ -76,7 +76,7 @@ jobs:
       with:
         maven-version: ${{ inputs.mavenVersion }}
     - name: Build with Maven
-      uses: coactions/setup-xvfb@b6b4fcfb9f5a895edadc3bc76318fae0ac17c8b3 # v1.0.1
+      uses: coactions/setup-xvfb@90473c3ebc69533a1a6e0505c36511b69c8c3135 # v1.0.1
```
jukzi, line 79: according to https://github.com/coactions/setup-xvfb/issues/29#issuecomment-2100951480 we are supposed to use head - which is not v1.0.1:
```suggestion
      uses: coactions/setup-xvfb@6b00cf1889f4e1d5a48635647013c0508128ee1a
```

.github/workflows/mavenBuild.yml
```diff
@@ -76,7 +76,7 @@ jobs:
         target: .github/matcher
     - run: echo "::add-matcher::.github/matcher/${{ steps.api-tools-matcher.outputs.filename }}"
     - name: Build with Maven
-      uses: coactions/setup-xvfb@b6b4fcfb9f5a895edadc3bc76318fae0ac17c8b3 # v1.0.1
+      uses: coactions/setup-xvfb@90473c3ebc69533a1a6e0505c36511b69c8c3135 # v1.0.1
```
jukzi, line 79:
```suggestion
      uses: coactions/setup-xvfb@6b00cf1889f4e1d5a48635647013c0508128ee1a
```

.github/workflows/codeQLworkflow.yml
fedejeanne: Done in 40c593b19bdd0c37ba3095f3b4fcdd3299c1dbf0, thank you!

jukzi: this still has the v1.0.1 annotation

fedejeanne: Forget it, I saw my mistake

fedejeanne: Done in 45a2052f7d7ba212e0e179438a72dfcadf67eb38, thank you!

jukzi (approved):

PR #2036, opened 2024-05-21
Update gson to 2.11.0.

merks:


merks:
@akurtakov 

I would very much prefer to update the gson version before the current release cycle...

akurtakov:
I see no reason to not do it in https://github.com/google/gson/releases/tag/gson-parent-2.11.0

rgrunber:
Any reason not to bump again to 2.12.0 based on https://github.com/eclipse-jdtls/eclipse.jdt.ls/issues/3236#issuecomment-2618077392 ?

laeubi:
@rgrunber it usually takes some time for the release to show up in the eclipse mirror, but if then it should get an automatic suggestion for update here with this job here:

https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/actions/workflows/callUpdateTarget.yml

merks:
Note it's still not on Maven Central:

- https://repo1.maven.org/maven2/com/google/code/gson/gson/
- https://repo1.maven.org/maven2/com/google/code/gson/gson/maven-metadata.xml

Orbit too would generally update to it as soon as it's available in Maven Central...

laeubi:
It is now here:
- https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/pull/2794

PR #2037, opened 2024-05-21
Print stacktrace on failure of maven builds

laeubi:
Currently stack trace are suppressed on maven build problems, as those are often the only way to understand unexpected build failures this now enables them by default. The drawback is that the output is a bit harder to read, but as we already have parsers for compile and test logs that present this to the user in a more comfortable way it seems a good compromise.


akurtakov:
My concerns are what you wrote already - readability. But I think stability overcomes these and one can always disable it if/when we see being stable.

Issue #2044, opened 2024-05-23
Declare 4.32 RC1


MohananRahul:
Declare 4.32 RC1
A "go" is needed from all the components below to declare this milestone.

Current Candidate

Eclipse downloads:
https://download.eclipse.org/eclipse/downloads/drops4/I20240522-1800 

Build logs and/or test results (eventually):
https://download.eclipse.org/eclipse/downloads/drops4/I20240522-1800/testResults.php

Software site repository:
https://download.eclipse.org/eclipse/updates/4.32-I-builds

Specific (simple) site repository:
https://download.eclipse.org/eclipse/updates/4.32-I-builds/I20240522-1800

Equinox downloads:
https://download.eclipse.org/equinox/drops/I20240522-1800

------

Deadlines:

Friday, 2024- May -24, around 1 AM (Eastern): deadline for sign-off (or, by then, comment in this issue when you expect to sign-off).
Friday, 2024 -May - 24, around 3 AM (Eastern): promote approved build to S-4.32RC1-*, contribute to simultaneous release repo, and announce to mailing lists.
Remember to investigate and document here any failing JUnit tests.

-----

- [ ] Platform: 
  - [ ]  Resources 
  - [ ]     UI 
  - [ ]  Debug 
  - [ ]  Ant 
  - [ ]  SWT 
  - [x]  Releng
- [ ] JDT:
  - [ ]  Core 
  - [ ]  Debug 
  - [x]  UI
- [x]  PDE
- [ ]  Equinox

iloveeclipse:
JDT Core needs time for analysis of two issue found very late:
- https://github.com/eclipse-jdt/eclipse.jdt.core/issues/2484
- https://github.com/eclipse-jdt/eclipse.jdt.core/issues/2481

We are working on that.

merks:
FYI, we have a bit of a disaster on your hands that pretty much affects **everyone**.

In particular, this build failing alerted me to the problem:

https://ci.eclipse.org/oomph/job/repository-analyzer/lastCompletedBuild/testReport/

The signing certificate has expired and so content is being signed by a certificate after it has expired, making the signature invalid:
```
$/c/Program\ Files/Java/jdk-21.0.2+13/bin/jarsigner.exe  -verbose -verify /d/stuff/org.eclipse.e4.ui.dialogs_1.5.0.v20240424-0957.jar

[...]

PR #2047, opened 2024-05-25
Force qualifier to pick up features containing resigned bundles

HannesWell:
https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/issues/2044


HannesWell:
Actually this is not necessary https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/issues/2044#issuecomment-2131330202

Issue #2074, opened 2024-06-03
POM and product version changes for 4.33 release


MohananRahul:

MohananRahul:
Prepare release 4.33
https://github.com/eclipse-jdt/eclipse.jdt.core.binaries/pull/12
https://github.com/eclipse-jdt/eclipse.jdt.debug/pull/447
https://github.com/eclipse-jdt/eclipse.jdt.core/pull/2515
https://github.com/eclipse-jdt/eclipse.jdt/pull/103
https://github.com/eclipse-jdt/eclipse.jdt.ui/pull/1436
https://github.com/eclipse-equinox/p2/pull/523
https://github.com/eclipse-platform/eclipse.platform.ui/pull/1918
https://github.com/eclipse-platform/eclipse.platform/pull/1402

PR #2076, opened 2024-06-03
Update for release 4.33

MohananRahul:
https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/issues/2074


merks:
FYI the license check fail not because of unvetted licenses but because the build itself fails.

PR #2098, opened 2024-06-05
Update the platform setup to redirect 4.31 to 4.32 and 4.32 to 4.33

merks:


merks:
This has no build impact.

merks:
FYI, if you do a pull on eclipse.platform.releng.aggregator, Help -> Perform Setup tasks will install Eclipse 4.33, which update the API baseline to 4.32 and the target platform to 4.33...

Issue #2101, opened 2024-06-06
No test results for I20240605-1800


iloveeclipse:
See https://download.eclipse.org/eclipse/downloads/drops4/I20240605-1800/
We have no idea about build quality.

At least SDK is starting now on Linux, so that [SWT binaries problem](https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/issues/2085) is not a problem anymore.

MohananRahul:
This issue appears to be related to the automated test , Jenkins parameters. somehow the parameter field is not generated when the test is triggered automatically. However, when we rebuild the test manually, the buildID parameter is available and  passed correctly. Raised :  https://gitlab.eclipse.org/eclipsefdn/helpdesk/-/issues/4717

iloveeclipse:
OK, we have some changes made by Sebastien on https://gitlab.eclipse.org/eclipsefdn/helpdesk/-/issues/4717 and now it **seem** to me that the tests are started properly but we have build? errors with some plugins missing
https://download.eclipse.org/eclipse/downloads/drops4/I20240606-0650/testResults.php

```
Missing Files
org.eclipse.core.tests.resources_ep433I-unit-win32-java17_win32.win32.x86_64_17.xml
org.eclipse.equinox.p2.tests_ep433I-unit-win32-java17_win32.win32.x86_64_17.xml
org.eclipse.jdt.apt.pluggable.tests_ep433I-unit-win32-java17_win32.win32.x86_64_17.xml
org.eclipse.jdt.apt.tests_ep433I-unit-win32-java17_win32.win32.x86_64_17.xml
org.eclipse.jdt.compiler.apt.tests_ep433I-unit-win32-java17_win32.win32.x86_64_17.xml
org.eclipse.jdt.compiler.tool.tests_ep433I-unit-win32-java17_win32.win32.x86_64_17.xml
org.eclipse.jdt.core.tests.builder_ep433I-unit-win32-java17_win32.win32.x86_64_17.xml
org.eclipse.jdt.core.tests.compiler_ep433I-unit-win32-java17_win32.win32.x86_64_17.xml
org.eclipse.jdt.core.tests.model_ep433I-unit-win32-java17_win32.win32.x86_64_17.xml
org.eclipse.jdt.core.tests.performance_ep433I-unit-win32-java17_win32.win32.x86_64_17.xml
[...]

PR #2107, opened 2024-06-06
[I-build] Fix I-build configuration regardin kubernetes container

HannesWell:
This contains the changes to the I-build config applied by @heurtematte directly to the job in Jenkins for
- https://gitlab.eclipse.org/eclipsefdn/helpdesk/-/issues/4717#note_2343210
- https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/issues/2101#issuecomment-2152845763

I tried to derive the changes from a diff of the current job config and the Groovy script in 
https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/blob/6c1b5bda2d4016a0d47dc4687c9b6342ea9cd17a/JenkinsJobs/Builds/I_build.groovy

I have not yet fully understood the change yet and I'm not sure if it is complete.
As far as I understood the [kubernetes Jenkins plugin](https://plugins.jenkins.io/kubernetes/#plugin-content-using-the-pipeline-step) `container('jnlp')` is applied by default if no container is specified. So removing that is just a clean-up?
On the other hand since `defaultContainer 'container'` specifies that the default container is named `container` it could mean that previously only the last stage run in the default container while all other run in another (unspecified?) `jnlp` container?

I also assume that much of the removal is because now the config inherits from `centos-8`?
But since the config before was `image: "eclipsecbi/jiro-agent-centos-8:latest"` and seemed more 'fully-qualified' I wonder if the short name is sufficient?

I have to leave now for about half an hour but will continue to when back.
@heurtematte it would be great if you could confirm or negate my assumptions.

I general I think it would be good to reduce this change to the necessary minimum too fix the builds and do clean-ups separately.

### DIFF
--- a/JenkinsJobs/Builds/I_build.groovy
+++ b/JenkinsJobs/Builds/I_build.groovy
@@ -59,15 +59,13 @@ pipeline {
   agent {
     kubernetes {
       label 'aggrbuild-pod'
-      defaultContainer 'container'
+      inheritFrom 'centos-8'
       yaml """
 apiVersion: v1
 kind: Pod
 spec:
   containers:
   - name: "jnlp"
-    image: "eclipsecbi/jiro-agent-centos-8:latest"
-    imagePullPolicy: "Always"
     resources:
       limits:
         memory: "8192Mi"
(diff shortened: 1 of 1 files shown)


HannesWell:
It just came into my mind that I can also diff the replay config of run 3 and 4:
- https://ci.eclipse.org/releng/job/Builds/job/I-build-4.33/3/replay/
- https://ci.eclipse.org/releng/job/Builds/job/I-build-4.33/4/replay/

And it seems that this PR is complete, but still I would like to reduce this if possible.
Call back later.

heurtematte:
You're right. I apologize for introducing the `inherit` change, which might have added more confusion to the fix.

Fair point: the agent's version is stricter. 

I don't fully understand the root cause myself. I noticed that the JNLP container was specified in each stage except for the one executing tests. This was just an assumption, and the test job confirmed it.

About the defaultContainer, I don't know why Jenkins never raised an issue before.

I haven't checked the Kubernetes plugin changelog, but there are quite a few fixes and features in latest versions. This is my only lead for now.

HannesWell:
> You're right. I apologize for introducing the `inherit` change, which might have added more confusion to the fix.

No problem and thanks for your answer. I just try to understand the change.

Regarding the `jnlp` container:
The [Overview](https://plugins.jenkins.io/kubernetes/#plugin-content-overview) section of the `Kubernetes` Jenkins plugin says:
`Within these pods, there is always one special container jnlp that is running the Jenkins agent. Other containers can run arbitrary processes of your choosing, and it is possible to run commands dynamically in any container in the agent pod.`
So everything inside the `jnlp` container runs in the same container (and process?) as the Jenkins agent.

Section [Using the pipeline step](https://plugins.jenkins.io/kubernetes/#plugin-content-using-the-pipeline-step) says:
`Commands will be executed by default in the jnlp container, where the Jenkins agent is running. (The jnlp name is historical and is retained for compatibility.)`

Furthermore section [Declarative Pipeline](https://plugins.jenkins.io/kubernetes/#plugin-content-declarative-pipeline) explains that `defaultContainer 'container'` defines, that if no container is specified a stage runs in a container with that name.

From that it seems that the `Trigger tests` and the `Trigger publication to Maven snapshots repo` stage used to run in a container named `container`, while all other stages run in the `jnlp` container. But actually no container named `container` was specified (no clue why it worked).
Now all stages in in the implicitly defined `jnlp` container.

That being said, the section [Static pod templates](https://plugins.jenkins.io/kubernetes/#plugin-content-static-pod-templates) says: 
`We do not recommend overriding the jnlp container except under unusual circumstances.`

So actually we should probably not use the `jnlp` but since it was like that since 'ever' I assume there was a special reason for running in the same container as the agent (maybe access rights, special files?).

Aslo the [`Multiple containers support` section](https://plugins.jenkins.io/kubernetes/#plugin-content-multiple-containers-support) says:
`Due to implementation constraints, there can be issues when executing commands in different containers if they run using different uids. It is recommended to use the same uid across the different containers part of the same pod to avoid any issue.`


With all that my current guess is that running it in the jnlp might have fixed the triggering, but using another image could have broke test execution, because the previous image defined a lot more volumes etc.? I have no clue if the default centos-8 does that too?
As another test I have now replayed build 3 (which has the config before your change) but removed the `defaultContainer 'container'` and added the `container('jnlp')` explicitly for the `Trigger tests` and the `Trigger publication to Maven snapshots repo` explicitly (remove the default should have the same effect):
https://ci.eclipse.org/releng/job/Builds/job/I-build-4.33/5/

Lets see if this already helps.

> I haven't checked the Kubernetes plugin changelog, but there are quite a few fixes and features in latest versions. This is my only lead for now.

Can I see somewhere the current version of the Kubernetes Plugin and ideally when it was updated and from which version?

HannesWell:
Build was successful, implemented the applied changes in https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/pull/2108

heurtematte:
You can find the plugins list here: https://github.com/eclipse-cbi/jiro/blob/master/instances/eclipse.platform.releng/target/jenkins/plugins.log

These are the Kubernetes version used on the releng jenkins instance: 

5 months ago:
kubernetes 4179.v3b_88431df708

2 weeks ago with the latest LTS:
kubernetes 4233.vb_67a_0e11a_039

From yesterday update:
kubernetes 4238.v41b_3ef14a_5d8

I have verified all kubernetes plugin changelogs and I haven't find anything that could have caused the issue.

HannesWell:
> I have verified all kubernetes plugin changelogs and I haven't find anything that could have caused the issue.

Nevertheless, thanks for checking.
Just submitted and deployed #2108. Lets see if this was sufficient.

HannesWell:
Closing this as the actual fix of https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/issues/2101 was https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/pull/2129 and the simplification made as part of the investigation are extracted and extended into https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/pull/2149.

PR #2108, opened 2024-06-06
[I-build] Fix I-build kubernetes container configuration

HannesWell:
Ensure all stages run within the 'jnlp' container and remove defaultContainer 'container' because a container named 'container' isn't defined.

See https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/issues/2101#issuecomment-2152845763

### DIFF
--- a/JenkinsJobs/Builds/I_build.groovy
+++ b/JenkinsJobs/Builds/I_build.groovy
@@ -59,7 +59,6 @@ pipeline {
   agent {
     kubernetes {
       label 'aggrbuild-pod'
-      defaultContainer 'container'
       yaml """
 apiVersion: v1
 kind: Pod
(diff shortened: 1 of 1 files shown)


HannesWell:
From https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/pull/2107#issuecomment-2153350353
> With all that my current guess is that running it in the jnlp might have fixed the triggering, but using another image could have broke test execution, because the previous image defined a lot more volumes etc.? I have no clue if the default centos-8 does that too?
> As another test I have now replayed build 3 (which has the config before your change) but removed the `defaultContainer 'container'` and added the `container('jnlp')` explicitly for the `Trigger tests` and the `Trigger publication to Maven snapshots repo` explicitly (remove the default should have the same effect):
> https://ci.eclipse.org/releng/job/Builds/job/I-build-4.33/5/

The build completed successfully and triggered the test jobs also with the `buildId` parameters as expected. This PR records the changes applied to that replayed job. But since this is a manually triggered build and they succeeded in the past I'm not absolutely sure this will fix #2101 and to confirm it we would have to await a cron-triggered build. But at least from my current understanding this is what I assume the best guess.

HannesWell:
I'm about to submit this now and await tonight's I-build result to see if this already fixes #2101 or at least if the missing parameters don't happen again.
#2107 still contains the full change applied manually before.

Issue #2109, opened 2024-06-07
Does the org.eclipse.rcp.configuration feature have a purpose, or can we delete it?


merks:
Yesterday I noticed that BIRT depends on something that is not on the release train, specifically the org.eclipse.rcp.configuration feature.  Then I wondered, can BIRT get rid of that thing, which of course begs the question, what does that thing provide?  So I looked today and see that this thing provides nothing and looking at the history has never provided anything.

![image](https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/assets/208716/37c677ad-6806-458d-8d2c-5a85294c598c)

This suggest to me that it is historical noise that we can eliminate.

laeubi:
I'm not sure if it really is empty, it includes root files (that is these getting placed inside an eclipse installation when installed) so maybe it still serves a purpose but not much anyone else should depend on?

laeubi:
Looking a bit closer it seems to copy the native launcher binaries into some test, can probably be replaced by https://tycho.eclipseprojects.io/doc/latest/tycho-p2-repository-plugin/repo-to-runnable-mojo.html today.

HannesWell:
> Looking a bit closer it seems to copy the native launcher binaries into some test, can probably be replaced by https://tycho.eclipseprojects.io/doc/latest/tycho-p2-repository-plugin/repo-to-runnable-mojo.html today.

Another alternative might be to use the [org.eclipse.equinox.executable.feature](https://github.com/eclipse-equinox/equinox/tree/master/features/org.eclipse.equinox.executable.feature) instead? I haven't looked into the details yet, but there might have been a reason why another feature has been introduced for this?

merks:
Yes, that seems to have similar build.properties:

https://github.com/eclipse-equinox/equinox/blob/master/features/org.eclipse.equinox.executable.feature/build.properties

If there's a reason to duplicated that, one used only for PDE build tests, I don't see it recorded anywhere. 

The build for this passed:

https://github.com/eclipse-pde/eclipse.pde/pull/1293

but apparently that doesn't really test the removal from the test.xml which is disturbing at best.

It seems the only thing to do is to proceed and see what happens.  Or?

HannesWell:
> Yes, that seems to have similar build.properties:
> 
> https://github.com/eclipse-equinox/equinox/blob/master/features/org.eclipse.equinox.executable.feature/build.properties
> 
[...]

PR #2120, opened 2024-06-11
Remove the org.eclipse.rcp.configuration feature and all its references

merks:
https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/issues/2109


merks:
@iloveeclipse  @akurtakov @MohananRahul 

I assume you folks have no concerns about removing this.  With PDE build removing its use of this, I think this has become garbage that can be collected.  Okay?

akurtakov:
None from me.

HannesWell (approved):
Looks good to me. 
I also found no other reference to org.eclipse.rcp.configuration in the SDK and think it's great to get rid of another ancient effectively unused piece of baggage.

merks:
Let's live dangerously and see what happens with the next I-Build.

PR #2125, opened 2024-06-12
Disable the platform setup 4.31 early redirection, and rough in for 4.32

merks:


merks:
Has no build impact and should be done ASAP.

PR #2128, opened 2024-06-12
Rename rt.equinox.binaries sub-module to equinox.binaries

HannesWell:
Adapt to the name changed assumption about the default name in equinox.
See https://github.com/eclipse-equinox/equinox/pull/641

### DIFF
--- a/cje-production/streams/repositories_java23.txt
+++ b/cje-production/streams/repositories_java23.txt
@@ -1,4 +1,4 @@
-rt.equinox.binaries: master
+equinox.binaries: master
 equinox: master
 rt.equinox.p2: master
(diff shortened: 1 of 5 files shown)


cje-production/streams/repositories_java23.txt
```diff
@@ -1,4 +1,4 @@
-rt.equinox.binaries: master
+equinox.binaries: master
```
HannesWell, line 1: @MohananRahul do you know if something has to be redeployed to make this effective?

MohananRahul: @HannesWell It seems that this is fine and has been utilized in Y&P Builds.

PR #2129, opened 2024-06-12
[I-Build-Tests] Declare parameters in job configuration

HannesWell:
Declaring the parameters only within the pipeline definition has the disadvantage that they have to be discovered in the first run and are therefore only available in the second run onwards. This leads to total failures of all I-build tests on the first run each time the seed job to create the Eclipse RelEng jobs ran.

Permanently fixes https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/issues/2101


HannesWell:
Submitting since this change is not tested by verification builds.
Will re-run the job-creation by the seed job immediately to have this in tonight's I-build.

PR #2131, opened 2024-06-13
Use Maven 3.9.7 in GH actions

akurtakov:


akurtakov:
This verification is not testing GH actions thus merging.

Issue #2134, opened 2024-06-13
Disable resigning of already signed jars


netomi:
As you are aware, signing of jars using the Eclipse Foundation signing service is currently quite slow as the service is in degraded mode and only support 1 concurrent signing atm.

While trying to mitigate that for now till we have a more scalable solution, I have noticed in the build at https://ci.eclipse.org/platform/job/eclipse.platform.releng.aggregator/job/master/ that there are lots of jars (> 800) that are signed for each build. Many of these jars are already signed, and you will find entries like that in the log file:

```
14:50:16  [INFO] Signing jar: /home/jenkins/agent/workspace/latform.releng.aggregator_master/equinox/bundles/org.eclipse.equinox.device/target/org.eclipse.equinox.device-1.2.100-SNAPSHOT.jar
14:50:19  [INFO] Jar '/home/jenkins/agent/workspace/latform.releng.aggregator_master/equinox/bundles/org.eclipse.equinox.device/target/org.eclipse.equinox.device-1.2.100-SNAPSHOT-sources.jar' is already signed and will be resigned.
```

I was wondering if it would not make sense to change the resigningStrategy of the eclipse-jarsigner plugin to something like `DO_NOT_RESIGN` instead of keeping at the default which is `RESIGN`.

More information about the configuration can be found at: https://eclipse-cbi.github.io/org.eclipse.cbi/eclipse-jarsigner-plugin/sign-mojo.html

Disable RESIGNING would reduce the number of signing operations considerably imho. 

cc @merks

akurtakov:
This is might be doable but has to be checked deeper by someone as resigning is done before the decision to replace or not with baseline version is taken 
```
15:56:41  [INFO] --- eclipse-jarsigner:1.4.3:sign (sign) @ org.eclipse.osgi.tests ---
15:56:41  [INFO] Signing of nested jars of '/home/jenkins/agent/workspace/latform.releng.aggregator_master/equinox/bundles/org.eclipse.osgi.tests/target/org.eclipse.osgi.tests-3.19.100-SNAPSHOT.jar' is disabled.
15:56:41  [INFO] Jar '/home/jenkins/agent/workspace/latform.releng.aggregator_master/equinox/bundles/org.eclipse.osgi.tests/target/org.eclipse.osgi.tests-3.19.100-SNAPSHOT.jar' is already signed and will be resigned.
[...]

PR #2140, opened 2024-06-17
[Build] Build don't resign signed jars and skip signing in master build

HannesWell:
Reduce the number of actual signing by not signing artifacts on master branch builds in this repository (the built artifacts are not published) and not re-signing artifacts in the I-build that are baseline replaced and therefore already signed.

Fixes https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/issues/2134

In order for `DO_NOT_RESIGN` to have an effect, signing should happen after a baseline-repacement done in `tycho-p2:p2-metadata`. Currently it is done before. Therefore this is a draft.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -13,14 +13,6 @@ pipeline {
 		jdk 'temurin-jdk21-latest'
 	}
 	stages {
-		stage('initialize PGP') {
-			steps {
-				withCredentials([file(credentialsId: 'secret-subkeys.asc', variable: 'KEYRING')]) {
-					sh 'gpg --batch --import "${KEYRING}"'
-					sh 'for fpr in $(gpg --list-keys --with-colons  | awk -F: \'/fpr:/ {print $10}\' | sort -u); do echo -e "5\ny\n" |  gpg --batch --command-fd 0 --expert --edit-key ${fpr} trust; done'
-				}
-			}
-		}
 		stage('Use master') {
 			steps {
 				sh 'git submodule foreach "git fetch origin master; git checkout FETCH_HEAD"'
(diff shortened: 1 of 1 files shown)


akurtakov:
PGP is not related to jarsigning and not using the jarsigner plugin (where the actual throughput is limited) thus I don't see benefit in dropping it. Actually, keeping it will still allow us to have signed content (pgp) without extra burdern on the infrastructure (as you are removing -Peclipse-sign)

HannesWell:
> PGP is not related to jarsigning and not using the jarsigner plugin (where the actual throughput is limited) thus I don't see benefit in dropping it. Actually, keeping it will still allow us to have signed content (pgp) without extra burdern on the infrastructure (as you are removing -Peclipse-sign)

It's right that PGP-signing is unrelated to jar-signing and not causing problems. But since both are activated with the same `eclipse-sign` profile it is currently not possible to activate only one of both without reworking these profiles in general.
Since the modified Jenkins pipeline is only used to run the master and verification builds of this repository I don't see a benefit in keeping the signatures because the artifacts build there are not published anywhere AFAIK.
If we keep the signing (pgp and jar) here and only use `DO_NOT_RESIGN` we more than double the number of signing compared to when not calling the profile. The reason is that each night when the nightly-build pushes the sub-module updates the master build runs and the signings that happen in the I-builds are repeated for that master build, plus the master builds due to manual changes that re-sign all other bundles changed in sub-modules since the last I-build.

HannesWell:
I replayed the build to activate signing for the verification build in order to replicate the I-build situation and it looks like swapping the `eclipse-jarsigner-plugin` and `tycho-p2-plugin:p2-metadata` has the desired effect and unchanged bundles are baseline replaced and not resigned.

But even when not re-signing alrady signed artifacts I see at least 91 artifacts being singed (the build has not yet completed). Some of them are for example not in the baseline, like tests.

akurtakov:
For the sake of minimizing the change (and probably https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/issues/2134#issuecomment-2165610288 ) please revert the reordering and push the disabling of signing altogether in the jenkinsfile to master.

HannesWell:
> For the sake of minimizing the change (and probably [#2134 (comment)](https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/issues/2134#issuecomment-2165610288) ) please revert the reordering and push the disabling of signing altogether in the jenkinsfile to master.

Reduced this PR now to just remove the signing of artifacts at the master/maintenance branch builds of this repository.
Create a dedicated PR to further reduce the number of signing operations by not re-signing already signed artifacts:
- https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/pull/2143

PR #2143, opened 2024-06-19
[Build] Don't resign already signed jars

HannesWell:
Reduce the number of actual signing operations by not re-signing artifacts (in the I-build) are already signed, usually because they have been 'baseline-replaced'.

In order to ensure that signatures of baseline-replaced artifacts can be considered, the p2-baseline-replacement is moved before the `eclipse-jarsigner-plugin`.

Furthermore set 'defaultP2Metadata' to false to skip the default baseline replace.

Fixes https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/issues/2134

This is a follow-up on https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/pull/2140.

### DIFF
--- a/eclipse-platform-parent/pom.xml
+++ b/eclipse-platform-parent/pom.xml
@@ -837,6 +837,7 @@
               </execution>
             </executions>
             <configuration>
+              <resigningStrategy>DO_NOT_RESIGN</resigningStrategy>
               <excludeInnerJars>${defaultSigning-excludeInnerJars}</excludeInnerJars>
               <!--
                The default timeout is 0 seconds which means "wait forever": https://www.eclipse.org/cbi/sitedocs/eclipse-jarsigner-plugin/sign-mojo.html#timeoutMillis


HannesWell:
> In order to ensure that signatures of baseline-replaced artifacts can be considered, the p2-baseline-replacement is moved before the `eclipse-jarsigner-plugin`.
> 
> Furthermore set 'defaultP2Metadata' to false to skip the default baseline replace.

I have analyzed and thought through this again and concluded that all we should change is to set `resigningStrategy=DO_NOT_RESIGN`. The subsequent invocation of the p2-metadata goal is necessary in order to update the metadata like size if the jar was altered through the signing process. Because we have the `p2-metadata-default` activated the baseline replacement that restores previous signings already happens (the first time) earlier.

Because in multiple sub-modules the source feature generation, which requires yet another `p2-metadata` invocation (usually named `attach-p2-metadata`) is, is not guarded in a profile only activated for features, there are in total three p2-metadata invocations for many `eclipse-plugin-projects`.
But that's another issue and ideally we would replace source-features by pure metadata, but that's again another issue.

To make a long story short I think this should be fine now as it is.
@akurtakov do you agree?

akurtakov (approved):

akurtakov:
Yes, no resigning if signed should be fine.

HannesWell:
Build failure is unrelated and should hopefully be fixed with https://github.com/eclipse-equinox/p2/pull/528.

PR #2146, opened 2024-06-20
[Build] Run eclipse sdk aggregator builds in parallel

HannesWell:
This hopefully reduces the runtime of the build. Since no tests are run in these builds, I hope there are no interference problems.


merks:
I think the problem is that some dependencies aren't resolving at all not that there are actual license dependencies with 3rd party libraries:
```
[INFO] --------------------------------[ pom ]---------------------------------
Warning:  The following dependencies could not be resolved at this point of the build but seem to be part of the reactor:
Warning:  o org.eclipse.jdt.ui:org.eclipse.jdt.ui.junit.sampleproject:jar:dist:1.0.0-SNAPSHOT (compile)
Warning:  Try running the build up to the lifecycle phase "package"
```

HannesWell:
In the verification builds of this PR using `-T 1C ` reduced the runtime of the Maven build to almost 50% of the current time.
Lets try this and see what reduction will be achieved in I-builds and if there are any issues.

I assume the additional Maven-dependency of `eclipse.platform.repository` to `eclipse-junit-tests` is completely internal and therefore not a problem.

> I think the problem is that some dependencies aren't resolving at all not that there are actual license dependencies with 3rd party libraries:
> 
> ```
> [INFO] --------------------------------[ pom ]---------------------------------
> Warning:  The following dependencies could not be resolved at this point of the build but seem to be part of the reactor:
> Warning:  o org.eclipse.jdt.ui:org.eclipse.jdt.ui.junit.sampleproject:jar:dist:1.0.0-SNAPSHOT (compile)
> Warning:  Try running the build up to the lifecycle phase "package"
> ```

This warning is also present for the current master build, therefore I assume it is unrelated.
I think it is because the extra mavne-dependency was added. I simply hope that @waynebeaton can resolve https://gitlab.eclipse.org/eclipsefdn/emo-team/iplab/-/issues/15495. IIRC such artifacts were added to a special list in the past or the dash tool were improved to find-out why a reactor project was considered as third-party dependency.

But since we probably cannot anything here, do you mind if I submit this now and we await the resolution of the vetting issue, wherever it happens?

.mvn/jvm.config
```diff
@@ -1,2 +1,2 @@
 -XX:+PrintFlagsFinal
--Xmx2G
+-Xmx3G
```
HannesWell, line 2: The latest verification build ran out of heap space.
Since we are running on the large 8GB executors since https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/pull/2148, I think 3GB should not be a problem.

HannesWell:
> @HannesWell
> 
> I don't think we should let glitches like this hold stuff up so I suggest we merge this when the build is good.

Then lets try this.
From the last runtime of builds in other PRs I'm not so confident anymore that there is really the runtime improvement mentioned above. It might have been heavily influenced by other builds but at least when I checked it my 'test' builds where running alone.

Anyway I think it's worth trying it out and if it eventually does not give sufficient speed-ups or causes too much trouble we can still revert it.

HannesWell:
In the I-build that contained this the Maven build stage took only 24min instead of ~40min. Five days ago, before the signing problem was addressed it took often more than two hours.

PR #2149, opened 2024-06-22
[I/Y/P-Builds] Simplify Kubernetes pod definitions

HannesWell:
Inherit most of the configuration from the centos-8 pod-template.

This extracts the changes made in https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/pull/2107 on the kubernetes spec.

### DIFF
--- a/JenkinsJobs/Builds/I_build.groovy
+++ b/JenkinsJobs/Builds/I_build.groovy
@@ -26,9 +26,6 @@ for (STREAM in STREAMS){
 # Post M1, no nightlies, I-builds only. (Be sure to "turn off" for tests and sign off days)
 #0 6 16-28 8 5-7,1-3
 #0 18 16-28 8 5-7,1-3
-
-
-
             """)
           }
         }
@@ -58,15 +55,13 @@ pipeline {
 	}
   agent {
     kubernetes {
-      label 'aggrbuild-pod'
+      inheritFrom 'centos-8'
       yaml """
 apiVersion: v1
 kind: Pod
 spec:
   containers:
   - name: "jnlp"
-    image: "eclipsecbi/jiro-agent-centos-8:latest"
-    imagePullPolicy: "Always"
     resources:
       limits:
         memory: "8192Mi"
@@ -74,75 +69,6 @@ spec:
       requests:
         memory: "6144Mi"
         cpu: "2000m"
-    securityContext:
-      privileged: false
-    tty: true
-    volumeMounts:
-    - mountPath: "/home/jenkins/agent"
-      name: "workspace-volume"
-      readOnly: false
-    - mountPath: "/home/jenkins/.m2/toolchains.xml"
-      name: "toolchains-xml"
-      readOnly: true
-      subPath: "toolchains.xml"
-    - mountPath: "/opt/tools"
-      name: "volume-0"
-      readOnly: false
-    - mountPath: "/home/jenkins"
-      name: "volume-2"
-      readOnly: false
-    - mountPath: "/home/jenkins/.m2/repository"
-      name: "volume-3"
-      readOnly: false
-    - mountPath: "/home/jenkins/.m2/settings-security.xml"
-      name: "settings-security-xml"
-      readOnly: true
-      subPath: "settings-security.xml"
-    - mountPath: "/home/jenkins/.m2/settings.xml"
-      name: "settings-xml"
-      readOnly: true
-      subPath: "settings.xml"
-    - mountPath: "/home/jenkins/.ssh"
-      name: "volume-1"
-      readOnly: false
-    workingDir: "/home/jenkins/agent"
-  nodeSelector: {}
-  restartPolicy: "Never"
-  volumes:
-  - name: "settings-security-xml"
-    secret:
-      items:
-      - key: "settings-security.xml"
-        path: "settings-security.xml"
-      secretName: "m2-secret-dir"
-  - name: "volume-0"
-    persistentVolumeClaim:
-      claimName: "tools-claim-jiro-releng"
-      readOnly: true
-  - configMap:
-      items:
-      - key: "toolchains.xml"
-        path: "toolchains.xml"
-      name: "m2-dir"
-    name: "toolchains-xml"
-  - emptyDir:
-      medium: ""
-    name: "volume-2"
-  - configMap:
-      name: "known-hosts"
-    name: "volume-1"
-  - name: "settings-xml"
-    secret:
-      items:
-      - key: "settings.xml"
-        path: "settings.xml"
-      secretName: "m2-secret-dir"
-  - emptyDir:
-      medium: ""
-    name: "workspace-volume"
-  - emptyDir:
-      medium: ""
-    name: "volume-3"
 """
     }
   }
--- a/JenkinsJobs/YBuilds/P_build.groovy
+++ b/JenkinsJobs/YBuilds/P_build.groovy
@@ -32,16 +32,13 @@ pipeline {
 	}
   agent {
     kubernetes {
-      label 'aggrbuild-pod'
-      defaultContainer 'container'
+      inheritFrom 'centos-8'
       yaml """
 apiVersion: v1
 kind: Pod
 spec:
   containers:
   - name: "jnlp"
-    image: "eclipsecbi/jiro-agent-centos-8:latest"
-    imagePullPolicy: "Always"
     resources:
       limits:
         memory: "8192Mi"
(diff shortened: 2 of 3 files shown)


JenkinsJobs/Builds/I_build.groovy
```diff
@@ -59,90 +56,20 @@ pipeline {
...
       label 'aggrbuild-pod'
+      inheritFrom 'centos-8'
       yaml """
 apiVersion: v1
 kind: Pod
 spec:
   containers:
   - name: "jnlp"
-    image: "eclipsecbi/jiro-agent-centos-8:latest"
-    imagePullPolicy: "Always"
```
HannesWell, line 69: @heurtematte in https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/pull/2107#issuecomment-2153252150 you said 
> Fair point: the agent's version is stricter.

about my statement before

> I also assume that much of the removal is because now the config inherits from centos-8?
> But since the config before was image: "eclipsecbi/jiro-agent-centos-8:latest" and seemed more 'fully-qualified' I wonder if the short name is sufficient?

Could you please elaborate on that please?
If I understand it correctly, with the previous configuration the image was always pulled from the registry on each build? Is the registry https://hub.docker.com/r/eclipsecbi/jiro-agent-centos-8?
And is the template centos-8 from which we now inherit this?
https://github.com/eclipse-cbi/jiro-agents/tree/master/centos-8

And if that's correct how can I make the template to inherit from more strict or less strict respectivly to what have you been referring exactly? more strict in regard which version of the centos-8 template being used (i.e. latest) or regarding if centos-8 or `centos-7` or `centos-latest` is used?

When reading through https://plugins.jenkins.io/kubernetes/#plugin-content-default-inheritance, I also wonder what the applicable values for `inheritFrom` are? Just `centos-7`, `centos-8`, `centos-latest`, the values we can also use for agents in ordinary Jenkinsfile?

JenkinsJobs/Builds/I_build.groovy
```diff
@@ -59,90 +56,20 @@ pipeline {
   agent {
     kubernetes {
       label 'aggrbuild-pod'
+      inheritFrom 'centos-8'
```
HannesWell, line 59: And when reading through https://plugins.jenkins.io/kubernetes/#plugin-content-warning-label-option-is-deprecated, I assume that this change will fix that warning that I can see in I-builds.

Or do I have to remove the label too? Is that label even still necessary?

JenkinsJobs/Builds/I_build.groovy
HannesWell: > When reading through https://plugins.jenkins.io/kubernetes/#plugin-content-default-inheritance, I also wonder what the applicable values for `inheritFrom` are? Just `centos-7`, `centos-8`, `centos-latest`, the values we can also use for agents in ordinary Jenkinsfile?

I assume that all labels defined in https://github.com/eclipse-cbi/jiro-agents/blob/e57775c2b706354b6bc10db9a5fde6cb6aa9ab7c/agents.jsonnet can be used, can't they?

heurtematte: > Could you please elaborate on that please?

Latest tag version in dockerhub is not the same as the one defined in podtemplate. 
* dockerhub: eclipsecbi/jiro-agent-centos-8:remoting-3160.vd76b_9ddd10cc = eclipsecbi/jiro-agent-centos-8:latest
* podtemplate: docker.io/eclipsecbi/jiro-agent-centos-8:remoting-3206.vb_15dcf73f6a_9

That's probably a mistake on our side, I'll try to figure it out this week.  

In any case, I would recommend using the one from podtemplate.

> Is the registry https://hub.docker.com/r/eclipsecbi/jiro-agent-centos-8?

right.

> https://github.com/eclipse-cbi/jiro-agents/tree/master/centos-8

And yes. 


Remoting is defined here with the latest configuration: https://github.com/eclipse-cbi/jiro-agents/blob/master/remoting/remoting.jsonnet

> And if that's correct how can I make the template to inherit from more strict or less strict respectivly to what have you been referring exactly? more strict in regard which version of the centos-8 template being used (i.e. latest) or regarding if centos-8 or centos-7 or centos-latest is used?

You can only override values defined as yaml in the cloud agent configuration. 
https://github.com/eclipse-cbi/jiro/blob/master/instances/eclipse.platform.releng/target/jenkins/configuration.yml#L142

If a value is defined by web UI, this one can't be overridden. 
I don't know exactly which other properties are overridable. If you need some, ask me and I'll do some tests.

> When reading through https://plugins.jenkins.io/kubernetes/#plugin-content-default-inheritance, I also wonder what the applicable values for inheritFrom are? Just centos-7, centos-8, centos-latest, the values we can also use for agents in ordinary Jenkinsfile?
I assume that all labels defined in https://github.com/eclipse-cbi/jiro-agents/blob/e57775c2b706354b6bc10db9a5fde6cb6aa9ab7c/agents.jsonnet can be used, can't they?

Current podtemplate are: `basic`, `basic-ubuntu`, `centos-7`, `centos-8`. 
And yes for instance, `centos-8` has these both labels: `centos-latest` `centos-8`

It is worth knowing that centos will be in the coming months considered as deprecated. We will communicate soon.

HannesWell: > > And if that's correct how can I make the template to inherit from more strict or less strict respectivly to what have you been referring exactly? more strict in regard which version of the centos-8 template being used (i.e. latest) or regarding if centos-8 or centos-7 or centos-latest is used?
> 
> You can only override values defined as yaml in the cloud agent configuration. https://github.com/eclipse-cbi/jiro/blob/master/instances/eclipse.platform.releng/target/jenkins/configuration.yml#L142

I think the right question to ask would have been, should the version be more specific or is it sufficient as it is. The latter is my guess.

> It is worth knowing that centos will be in the coming months considered as deprecated. We will communicate soon.

Thanks for the heads-up. I heard something similar already somewhere else. I assume we then simplify have to change the `inheritFrom` value? I don't think this will be a big problem.

HannesWell (commented):
Thank you @heurtematte for your answers! I have now removed the labels as suggested.
Do you think the current state is fine?
Since you tested the change already in https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/pull/2107, I think it is relatively save to assume it works as expected?

@sravanlakkimsetti do you want to take a look at this as well?

JenkinsJobs/Builds/I_build.groovy
heurtematte: It is sufficient as it is.

heurtematte:
LGTM!

HannesWell:
The triggered I-build failed with
https://ci.eclipse.org/releng/job/Builds/job/I-build-4.33/35/
```
[ERROR] Failed to execute goal org.eclipse.tycho:tycho-compiler-plugin:4.0.8:compile (default-compile) on project org.eclipse.equinox.http.service.api: useJDK = BREE configured, but no toolchain of type 'jdk' with id 'JavaSE-1.8' found. See https://maven.apache.org/guides/mini/guide-using-toolchains.html -> [Help 1]
 org.apache.maven.lifecycle.LifecycleExecutionException: Failed to execute goal org.eclipse.tycho:tycho-compiler-plugin:4.0.8:compile (default-compile) on project org.eclipse.equinox.http.service.api: useJDK = BREE configured, but no toolchain of type 'jdk' with id 'JavaSE-1.8' found. See https://maven.apache.org/guides/mini/guide-using-toolchains.html
```

My first guess is that some of the `volumeMounts` about `toolchains.xml` that were removed were not useless. But it's unexpected since the build of this repo and other project builds also use `-Pbree-libs` with the `centos-latest` pob-template.
I have replayed the build with some of these mounts restored to check if that's really the cause.

HannesWell:
When comparing the pod definition that is printed on the console in the very beginning of a job, e.g.:
```
Created Pod: kubernetes releng/builds-i-build-4-33-37-n6f4m-32g09-nhhz6
Agent [builds-i-build-4-33-37-n6f4m-32g09-nhhz6](https://ci.eclipse.org/releng/computer/builds-i-build-4-33-37-n6f4m-32g09-nhhz6) is provisioned from template Builds_I-build-4_33_37-n6f4m-32g09
...
```
of the current [I-build-4.33](https://ci.eclipse.org/releng/job/Builds/job/I-build-4.33/) job, for example with [eclipse.platform.releng.aggregator/master](with https://ci.eclipse.org/platform/job/eclipse.platform.releng.aggregator/job/master/) job, where the latter just uses in its pipeline a simple labeled agent `agent { label "centos-latest-8gb" }
`
I see that the following `volumeMounts` and (what I believe are) the corresponding `volumes` are missing:
```
    volumeMounts:
    - mountPath: "/home/jenkins/.m2/toolchains.xml"
      name: "m2-dir"
      readOnly: true
      subPath: "toolchains.xml"
    - mountPath: "/home/jenkins/.m2/settings-security.xml"
      name: "m2-secret-dir"
      readOnly: true
      subPath: "settings-security.xml"
    - mountPath: "/home/jenkins/.m2/settings.xml"
      name: "m2-secret-dir"
      readOnly: true
      subPath: "settings.xml"
```
Besides the `toolchains.xml`, which I added back in #2157, the absence of the other two mounts seem not to be problematic.
But nevertheless I wonder why they are missing?
@heurtematte can you tell that? Does inheriting a pod-template not inherit the `volum(mounts)`? Does it have to be activated explicitly? Or are we eventually using actually another template than the `eclipse.platform.releng.aggregator/master` that only defines a labeled agent?

HannesWell:
> @HannesWell I have updated the jenkins instance with the merge strategy for podtemplate. It is not anymore necessary to replicate those mount volume definitions.

Awesome! Thank you. Just created https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/pull/2158 to remove those extra volume(Mounts) again. Verified it before with a replayed (shortly after its start aborted) I-build that the volume(Mounts) are now inherited as intended.

PR #2150, opened 2024-06-22
Clean-up and generalize testPlatforms from runTests2 ANT script

HannesWell:
- Remove unsupported/unused testPlatform values and configurations and generalize the windows configuration.

- Remove obsolete java version checks and duplicated java-version print-outs
- Remove commented out javascript tasks

This also simplifies to run tests for new architectures on Windows.

PR #2151, opened 2024-06-22
Restore creation of javaversionEscaped property in runTests2 ANT script

HannesWell:
This was removed in https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/pull/2150 although the property `javaversionEscaped` is still in use.


HannesWell:
Submitting this now to not affect the I-build tests and will abort the just started I-build and will trigger a new one.

PR #2152, opened 2024-06-24
Align name of rt.equinox.p2 submodule to actual repository name

HannesWell:
... which is equinox.p2

@akurtakov, @iloveeclipse or anyone else, do you have objections?


iloveeclipse:
> ... which is equinox.p2
> 
> do you have objections?

Not really, but I'm not deep inside SDK module/job/script configuration to be qualified for a review.

akurtakov (approved):
This is long overdue.

HannesWell:
Thank you both.

> > ... which is equinox.p2
> > 
> > do you have objections?
> 
> Not really, but I'm not deep inside SDK module/job/script configuration to be qualified for a review.

This is similar to https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/pull/2128 which worked flawlessly.
In general I searched the entire aggregator for the old submodule name and replaced it where suitable. There are some variable names left, but I leave that for a follow-up.

Therefore I'm relatively confident this works too, but I'll check tonight's I-build if something is missing.

PR #2157, opened 2024-06-25
[I/Y/P-Builds] Restore mount of toolchains.xml in K8 pod definitions

HannesWell:
Fixes a regression from https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/pull/2149


HannesWell:
Verification build does not test this. But I have already tested this with a Replay in Jenkins of a previously failing I-build.

PR #2171, opened 2024-07-08
Add help context for the find/replace overlay

Wittmaxi:
Adds a help context for the find/replace overlay which links to the help context for the dialog.

related to
https://github.com/eclipse-platform/eclipse.platform.ui/pull/2045

### DIFF
--- a/eclipse.platform.common/bundles/org.eclipse.platform.doc.user/contexts_Workbench.xml
+++ b/eclipse.platform.common/bundles/org.eclipse.platform.doc.user/contexts_Workbench.xml
@@ -8,370 +8,386 @@
     </context>
     <context id="dock_on_perspective_action_context">
         <description>This docks the perspective bar in different locations.</description>
-        <topic label="Perspectives" href="concepts/concepts-4.htm"/>
-        <topic label="Perspective Bar" href="reference/ref-3.htm"/>
+        <topic label="Perspectives" href="concepts/concepts-4.htm" />
+        <topic label="Perspective Bar" href="reference/ref-3.htm" />
     </context>
     <context id="toggle_coolbar_action">
         <description>This toggles the visibility of the window toolbar and perspective bar.</description>
-        <topic label="Perspective Bar" href="reference/ref-3.htm"/>
-        <topic label="Window Menu - Customize Perspective" href="reference/ref-60.htm"/>
+        <topic label="Perspective Bar" href="reference/ref-3.htm" />
+        <topic label="Window Menu - Customize Perspective" href="reference/ref-60.htm" />
     </context>
     <context id="show_text_perspective_action_context">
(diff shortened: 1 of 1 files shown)


Wittmaxi:
grafik

eclipse.platform.common/bundles/org.eclipse.platform.doc.user/contexts_Workbench.xml
```diff
@@ -510,6 +510,10 @@
         <description>Dialog to find and replace specified text.</description>
 		<topic label="Find/Replace with regular expressions" href="reference/ref-regex-find-replace.htm" />
     </context>
+    <context id="find_replace_overlay_context">
+        <description>Overlay to find and replace specified text.</description>
+		<topic label="Find/Replace with regular expressions" href="reference/ref-regex-find-replace.htm" />
```
HeikoKlare, line 515: Please correct indentation (and maybe also fix the one in the context where this was copied from).

Wittmaxi: Can you please check that the indentation is wrong in the IDE as well? Sometimes github shows some indentation wrong. This is what I see in the IDE:

grafik

HeikoKlare: The indentation is inconsistent, which is why different viewers render different visible indentations. One line uses whitespaces and one uses tabs. In the IDE, by accident, the number of whitespaces aligns with the tab size (you could even tell the IDE to render tabs differently and would then also see a difference):
image

Wittmaxi: Sorry, I didn't see that!
I have changed *all* tabs into four spaces

HeikoKlare: Thanks. Unfortunately, just transforming tabs into whitespaces is not correct, as there have been combinations of whitespaces and tabs which still behaves like a single tab (e.g. 1 whitespace + 1 tab behaves like 1 tab). So there are now places with too many whitespaces.

Wittmaxi: It looks like Eclipse has no autoformat for XML by default (??!)
I copied the file into vscode (which also doesn't autoformat XMLs by default?! I used the RedHat XML extension) and formatted it there.

HeikoKlare: Yes, if I am not mistaken, both require (additional) extensions to (auto-)format XMLs.

PR #2183, opened 2024-07-11
Update to ECF 3.15.2 m1a

merks:


merks:
@akurtakov 

It looks like the license check:

https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/actions/runs/9903172956

fails to run completely

image

akurtakov:
https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/pull/2184 should fix it.

PR #2195, opened 2024-07-23
Document the new --launcher.noRestart launcher argument.

dhendriks:
This new argument just got added. See https://github.com/eclipse-equinox/equinox/pull/595. Here we add the documentation for it.

### DIFF
--- a/eclipse.platform.common/bundles/org.eclipse.platform.doc.isv/reference/misc/runtime-options.html
+++ b/eclipse.platform.common/bundles/org.eclipse.platform.doc.isv/reference/misc/runtime-options.html
@@ -115,6 +115,10 @@ eclipse.ini, product.exe looks for product.ini)</dd>
   <dd>If specified the executable will not display any error or message
 dialogs.&nbsp; This is useful if the executable is being used in an
 unattended situation.</dd>
+  <dt>--launcher.noRestart (Executable)</dt>
+  <dd>If specified, the special handling of exit codes 23 (restart) and 24
+(relaunch) is disabled and they are treated as regular exit codes. This can be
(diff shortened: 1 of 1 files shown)


eclipse.platform.common/bundles/org.eclipse.platform.doc.isv/reference/misc/runtime-options.html
```diff
@@ -115,6 +115,10 @@ <h2><a name="commandline"></a>Command line arguments </h2>
   <dd>If specified the executable will not display any error or message
 dialogs.&nbsp; This is useful if the executable is being used in an
 unattended situation.</dd>
+  <dt>--launcher.noRestart (Executable)</dt>
+  <dd>If specified, the special handling of exit codes 23 (restart) and 24
+(relaunch) is disabled, treating them like regular exit codes. This is useful
```
HannesWell, line 120: I'm not a native English-speaker so that impression might be wrong, but I'm not sure if the second half of the first sentence is correct?
Furthermore I think the second sentence should be a bit more defensive. 
```suggestion
(relaunch) is disabled and they are treated as regular exit codes. This can be useful
```

HannesWell (commented):
Thank you for this prompt doc-update.

eclipse.platform.common/bundles/org.eclipse.platform.doc.isv/reference/misc/runtime-options.html
dhendriks: I applied it, with rewrapping to 80 character lines, as a forced pushed update to the existing commit.

HannesWell (approved):
Thanks!

PR #2205, opened 2024-08-04
Add workflow that automatically updates target dependencies

laeubi:
Currently we need to keep track of new versions manually what can become quite cumbersome.

This creates an (experimental) automatic job that is triggered each day and check if there are updates available and suggests a PR to update them.

### DIFF
--- a/.github/workflows/updateTarget.yml
+++ b/.github/workflows/updateTarget.yml
@@ -1,0 +1,38 @@
+name: Update Target Platform
+
+on:
+  schedule:
+    - cron:  '0 0 * * *'
+jobs:
+  update:
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@692973e3d937129bcbf40652eb9f2f61becf3332 # v4.1.7
+      with:
+        fetch-depth: 0
+        ref: master
+    - name: Set up Maven
+      uses: stCarolas/setup-maven@d6af6abeda15e98926a57b5aa970a96bb37f97d1 # v5
+      with:
+        maven-version: 3.9.7
+    - name: Set up JDK
+      uses: actions/setup-java@99b8673ff64fbf99d8d325f52d9a5bdedb8483e9 # v4.2.1
+      with:
+        java-version: '17'
+        distribution: 'temurin'
+        cache: maven
+    - name: Update Target Platform
+      run: >-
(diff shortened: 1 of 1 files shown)


mickaelistria:
Let's merge and see what happens.

mickaelistria:
See https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/pull/2209 . This bot isn't providing such helpful input at the moment. I'm going to revert it.

laeubi:
@mickaelistria sorry I was away from the computer last week.

> how does it identify new versions?

It is using the metadata (for example P2 or configured maven repositories) to simply search for the latest version and updates the version string.

> Does it work only for Maven content or also with p2 content? 

Currently it support Maven + P2 but can technically enhanced to support any type of location that provides suitable information

> Can it even already suggest to use a newer p2 repo?

Not in the current state but that is something I'm planning to support

> Let's merge and see what happens.

- https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/pull/2209

>  This bot isn't providing such helpful input at the moment. I'm going to revert it.

That a bit unfortunate as the goal is to get some experience, find bugs and improvements and this won't work if we do this back-and-forth when some issues arises (that don't really affect any production builds), at least we can now see that it produces a PR as desired so we can review / improve and possibly pick whats useful and identify things that need improvements.

mickaelistria:
>  this won't work if we do this back-and-forth when some issues arises 

I think the bot "only" calls a Tycho mojo. And with #2209 we identified some concrete issues with the proposal of this Tycho mojo. Improving the mojo can be done locally, running it can easily be done locally without creating "noise" on the Eclipse project tracker. There is enough feedback here and on #2209 to trigger some development. When all the provided feedback is addressed, then we can reconsider enabling the bot and see what the next version produces. But at the moment, I think there is clearly enough input to drive development and pause the experiment on Platform tracker.

laeubi:
> But at the moment, I think there is clearly enough input to drive development and pause the experiment on Platform tracker.

I don't agree here, the workflow is for the automation of the process and gather feedback from others, nothing of those can be accomplished with local testing and one open PR is negligible compared to having noise in the git history bu adding / removing the same code over and over again on each iteration.

Beside that, workflows can be temporarily being disabled in github, there is no need to delete the file itself.

PR #2214, opened 2024-08-11
Add workflow that automatically suggests updates of target dependencies

laeubi:
Currently we need to keep track of new versions manually what can become quite cumbersome.

This creates an (experimental) automatic job that is triggered each day and check if there are updates available and suggests a PR to update them.

### DIFF
--- a/.github/workflows/updateTarget.yml
+++ b/.github/workflows/updateTarget.yml
@@ -1,0 +1,39 @@
+name: Update Target Platform
+
+on:
+  schedule:
+    - cron:  '0 0 * * *'
+jobs:
+  update:
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@692973e3d937129bcbf40652eb9f2f61becf3332 # v4.1.7
+      with:
+        fetch-depth: 0
+        ref: master
+    - name: Set up Maven
+      uses: stCarolas/setup-maven@d6af6abeda15e98926a57b5aa970a96bb37f97d1 # v5
+      with:
+        maven-version: 3.9.7
+    - name: Set up JDK
+      uses: actions/setup-java@99b8673ff64fbf99d8d325f52d9a5bdedb8483e9 # v4.2.1
+      with:
+        java-version: '17'
+        distribution: 'temurin'
+        cache: maven
+    - name: Update Target Platform
+      run: >-
+          mvn -f eclipse.platform.releng.prereqs.sdk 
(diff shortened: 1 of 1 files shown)


mickaelistria:
Is it expected to fix all the concerns of previous iteration?

laeubi:
As explained the workflow itself can't fix anything (it is already working as expected):

1. call the update-target mojo
2. create (or update) the PR if there are any difference in the target files

If anything of the resulting target is undesired, this has to be fixed/changed at Tycho (and probably update the configuration) and then the workflow updates the PR on the next run with the hopefully better result.

mickaelistria:
> If anything of the resulting target is undesired

#2209 is listing undesired proposals in the change that make this mojo/workflow currently not profitable for Platform.

laeubi:
> > If anything of the resulting target is undesired
> 
> #2209 is listing undesired proposals 

And they will be addressed step-by-step

> in the change that make this mojo/workflow currently not profitable for Platform.

I have explained already why it is "profitable" for platform and the change does not harm as one can simply ignore the pending PR, I would agree that it might be disturbing if we usually have zero open PRs over all platform repositories but this is not the case here, I can change the workflow to only open a draft-pr but that's just syntactic sugar for me.

mickaelistria:
> one can simply ignore the pending PR

Filtering actionable, pending, worthy, ... PRs is actually a huge part of the job. Adding noise there by creating PRs that we know are currently not good is adding noise and requiring all developers to decide how to process such PR.
Platform isn't a sandbox, there are many developers involved and we must all make an effort so all other contributors don't feel distrurbed more than necessary. If you're curious about the outcome of this workflow, please enable it in your fork and you'll get the results you need without adding noise.

> And for me...

As long as you've not convinced any other contributor who regularly deals with updates (I'm thinking mostly of @merks or @akurtakov these days) that the current bot is creating profitable PR for their work too, then please keep it in your own fork. As soon as anyone else agrees with you that we're better with the bot than without, then it's all good to merge it in the project.

akurtakov:
If this will be actively worked on - I am fine with having PRs open and etc. and if it stagnates before reaching usable state we should disable the workflow.

merks:
I agree. I assume @laeubi will monitor these to fine tune the details.

laeubi:
Most concerns are already addressed and there for Tycho4, so lets see if we can find more.
The remaining one is actually correct and we should decide how to scope with that.

merks:
The next time the orbit generator generates an update for the platform target, I’ll contact @laeubi  about it.

PR #2227, opened 2024-08-15
Update jakarta.annotation-api to 3.0.0

laeubi:


laeubi:
Okay it seems we need to make some more steps here:

1. Add 3.0.0 to the target so it is usable
2. update (and check) version ranges to use upper bound 3.0.0
3. then we can remove 2.0.0

merks:
FYI, I looked here:

https://jakarta.ee/specifications/annotations/3.0/

and the only comment I see for why it's a 3.0.x version is this:

image

I'm quite sure we aren't using that so either 2.x or 3.x should be fine for the Platform.

PR #2228, opened 2024-08-15
Remove includes of 3rd party libraries from the org.eclipse.test feature

merks:


merks:
This is similar to https://github.com/eclipse-jdt/eclipse.jdt/pull/107

In this case though, the two selected bundles:

image

are in the transitive closure of dependencies only because of inclusion in the feature:

image

image

Those could be converted to imports instead of include, but then I'm not entirely sure of the purpose of the test feature...

akurtakov:
Actually this feature is used to install all the test deps from p2 repo in order to make working on tests easy. Unfortunately I am not aware of another way of "install tests dependencies from I-build at once" and nothing in the feature actually requires these third party deps.

merks:
> Actually this feature is used to install all the test deps from p2 repo in order to make working on tests easy. Unfortunately I am not aware of another way of "install tests dependencies from I-build at once" and nothing in the feature actually requires these third party deps.

Note that all the other dependencies are transitively required, e.g., 

image

The only goal of this PR is to avoid needing to touch the feature when one of the 3rd party bundles  is updated, so converting those two that are not in the transitive closure would also accomplish that goal.  Or converting all of them to imports would also be fine accomplish the goal.

I'm curious to see the result of the build.

I'll mark this as a draft for the time being.

merks:
The build itself passed, but that doesn't mean haven't made the test feature less useful but these changes?

@akurtakov 

What do you suggest here?   

- Keep these changes?  
- Convert the ones that will be transitively missing to imports?  
- Convert all to imports?   

When to do this?
- Now?
- M1

akurtakov:
Convert the transitively misisng ones to imports for M1 is best IMO.

merks:
I created a new PR failing to update this one:

https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/pull/2332

PR #2230, opened 2024-08-16
Create a shared workflow that allows updating bundle dependencies

laeubi:


laeubi:
I have now successfully tested this here:

- https://github.com/laeubi/p2/pull/9

example PR is:
- https://github.com/laeubi/p2/pull/80

so this is now ready to be merged (and used).

Issue #2166, opened 2024-07-04
Eclipse 4.33 Prerequisites: EMF


MohananRahul:
Tracking issue for EMF.

merks:
EMF's final contribution is complete.

PR #2234, opened 2024-08-20
Update Orbit dependencies

merks:
- Update commons-logging to 1.3.4.
- The orbit repository contains no changes to the included IUs.

https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/issues/2166

### DIFF
--- a/eclipse.platform.releng.prereqs.sdk/eclipse-sdk-prereqs.target
+++ b/eclipse.platform.releng.prereqs.sdk/eclipse-sdk-prereqs.target
@@ -52,7 +52,7 @@
       <unit id="org.commonmark-gfm-tables" version="0.22.0.v20240316-0700"/>
 
       <!-- This is the "normal" Orbit repository is expected to be updated on milestones and releases based on Orbit deliveries. -->
-      <repository location="https://download.eclipse.org/tools/orbit/simrel/orbit-aggregation/milestone/S202408130653"/>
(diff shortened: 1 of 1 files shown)


akurtakov:
+1

eclipse.platform.releng.prereqs.sdk/eclipse-sdk-prereqs.target
```diff
@@ -52,7 +52,7 @@
       <unit id="org.commonmark-gfm-tables" version="0.22.0.v20240316-0700"/>
 
       <!-- This is the "normal" Orbit repository is expected to be updated on milestones and releases based on Orbit deliveries. -->
-      <repository location="https://download.eclipse.org/tools/orbit/simrel/orbit-aggregation/milestone/S202408130653"/>
+      <repository location="https://download.eclipse.org/tools/orbit/simrel/orbit-aggregation/milestone/S202408200619"/>
```
laeubi, line 55: I wonder whats the rationale behind changing just the URL but no versions, does it mean actually nothing has changed?

merks: Because that's the most recent milestone, and does contain updates, but just not ones the Platform is using.  It's not uncommon that other folks look at the Platform to see what "they should use" based on what the Platform is using.

PR #2236, opened 2024-08-20
Add 'incompatibilities' entry about inlined native fragments

HannesWell:
To inform about changes in
- https://github.com/eclipse-platform/eclipse.platform/pull/1394
- https://github.com/eclipse-platform/eclipse.platform/pull/1476
- https://github.com/eclipse-platform/eclipse.platform/pull/1440

@akurtakov I think this is the right thing to do, isn't it?


akurtakov (approved):
Thanks for writing it.

PR #2251, opened 2024-08-24
Update Update byte-buddy to 1.15.0

merks:


merks:
@akurtakov 

Do we want to update to this now?  In the past, you mentioned this was only a test dependency.  In this case though it's a new minor version...

akurtakov:
Go for it.

PR #2259, opened 2024-08-27
Add gtk4-devel to centos-9 for swt and remove gtk2

laeubi:


cje-production/dockerfiles/centos-gtk4-mutter/9-swtBuild/Dockerfile
```diff
@@ -31,10 +31,9 @@ RUN dnf -y update && dnf -y install \
       webkit2gtk3 \
       webkit2gtk3-devel \
       gtk3-devel \
-      gtk2 \
-      gtk2-devel \
+      gtk4-devel \
       java-17-openjdk-devel \
-      webkit2gtk3-devel \
+      webkit2gtk4-devel \
```
akurtakov, line 36: There seems no webkit2gtk4-devel at https://mirror.stream.centos.org/9-stream/AppStream/x86_64/os/Packages/ so that will break the build. 
It should be possible to build webkit part without having the headers as everything from it is dynamically loaded due to the huge range of versions.
Please keep the gtk3 version - in case it's used to transitively install webkit dependencies and thus provide everything needed for tests.

akurtakov (approved):
Looks good

Issue #2299, opened 2024-09-03
Version bumps for 4.34 stream


MohananRahul:

akurtakov:
As per https://download.eclipse.org/eclipse/downloads/drops4/I20240903-1200/buildlogs/reporeports/reports/versionChecks.html there are no more bundles with qualifier bump only.

PR #2305, opened 2024-09-03
Version bump for 4.34 stream

akurtakov:
Contributes to
https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/issues/2299


akurtakov:
Verification can't succeed before all other repos are clean from version bumps.

PR #2313, opened 2024-09-04
Update the platform setup to redirect 4.32 to 4.33 and 4.33 to 4.34

merks:


merks:
The installer will prompt for this:

image

Then the director task will install/update to 4.34:

image

The API baseline will be the current RC2a build:

image

The target platform will be based on 4.34:

image

PR #2322, opened 2024-09-06
add org.eclipse.jdt.debug.jdi.tests to daily build

jukzi:
https://github.com/eclipse-jdt/eclipse.jdt.debug/issues/484

### DIFF
--- a/production/testScripts/configuration/sdk.tests/testScripts/test.xml
+++ b/production/testScripts/configuration/sdk.tests/testScripts/test.xml
@@ -2137,6 +2137,12 @@
   </target>
 
   <target
+    name="jdtjdi"
+    depends="init">
+    <runTests testPlugin="org.eclipse.jdt.debug.jdi.tests" />
+  </target>
+
+  <target
     name="jdtui"
     depends="init">
     <runTests testPlugin="org.eclipse.jdt.ui.tests" />
@@ -2730,6 +2736,7 @@
     <antcall target="jdtcoreperf" />
     <antcall target="jdtcorebuilder" />
     <antcall target="jdtdebug" />
+    <antcall target="jdtjdi" />  	
     <antcall target="jdtapt" />
     <antcall target="jdtaptpluggable" />
   </target>
--- a/eclipse.platform.releng.tychoeclipsebuilder/eclipse/publishingFiles/testManifest.xml
+++ b/eclipse.platform.releng.tychoeclipsebuilder/eclipse/publishingFiles/testManifest.xml
@@ -641,6 +641,9 @@
       name="org.eclipse.jdt.debug.tests"
       type="test" />
     <logFile
+      name="org.eclipse.jdt.debug.jdi.tests"
+      type="test" />
+    <logFile
       name="org.eclipse.jdt.text.tests"
       type="test" />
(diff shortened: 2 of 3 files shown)


jukzi:
@jarthana

jukzi:
i have no clue what these files are about, but they seem to be missing jdi. i added them like the org.eclipse.jdt.debug.tests tests. i could need some help/review here! @HeikoKlare ?

production/testScripts/configuration/sdk.tests/testScripts/test.xml
```diff
@@ -2136,6 +2136,12 @@
     <runTests testPlugin="org.eclipse.jdt.debug.tests" />
   </target>
 
+  <target
+    name="jdtjdi"
+    depends="init">
+    <runTests testPlugin="org.eclipse.jdt.debug.jdi.tests" />
+  </target>
```
HeikoKlare, line 2143: You also have to call this target from somewhere to have it executed in the build. Maybe a good position could be in this existing target for JDT, adding just another antcall target with the id "jdtjdi":
https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/blob/b7c704a494d7a091933ec5638583463d8fa5ba19/production/testScripts/configuration/sdk.tests/testScripts/test.xml#L2724-L2735

eclipse.platform.releng.tychoeclipsebuilder/eclipse/publishingFiles/performanceTestManifest.xml
```diff
@@ -17,6 +17,9 @@
     <logFile
       name="org.eclipse.jdt.debug.tests"
       type="test" />
+    <logFile
+      name="org.eclipse.jdt.debug.jdi.tests"
+      type="test" />
```
HeikoKlare, line 22: I am not sure whether this one is necessary. I think it depends on whether you want to have performance tests from that plug-in to be executed. In that case, you will probably also need to add the plug-in to the performance target in the `test.xml`, but I am not sure whether that's your intention. Otherwise, I would simply revert this change to avoid that builds look for non-existing log files.

HeikoKlare (commented):
From my understanding, you found all files that need to be touched to address the problem. Only a call of the added Ant target seems to be missing and the performance test log addition seems to be obsolete (see my comments). Once that's addressed, the PR should be good to go.

FYI: For validation, I have also taken a look at a previous plug-in addition to daily build (https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/pull/2035).

eclipse.platform.releng.tychoeclipsebuilder/eclipse/publishingFiles/performanceTestManifest.xml
jukzi: ok, removed

production/testScripts/configuration/sdk.tests/testScripts/test.xml
jukzi: ok, did that, thanks!

HeikoKlare (approved):
Thank you for the update. Looks good now.

PR #2331, opened 2024-09-10
Update target platform to Orbit M0.

merks:
- Update org.apache.ant to 1.10.15.


merks:
I'm not sure if the include in the org.eclipse.platform feature will build errors problems.  We shall see.

PR #2332, opened 2024-09-10
Remove includes of 3rd party libraries from the org.eclipse.test feature

merks:


merks:
I separate out the commit of the version increment from the commit of the changes to the feature.

PR #2338, opened 2024-09-12
Use EF Ubuntu 24.04 image for smoke tests

akurtakov:
Platform releng one is still on 22.04 and now that foundation provides 24.04 as a default one we better rely on it.


akurtakov:
Only I-builds run smoke tests thus merging directly.

PR #2341, opened 2024-09-12
Fix jdt.feature version in P build and add commonmark bundles to the patch

jarthana:

PR #2344, opened 2024-09-13
Adding markdown lib to P-build

sravanlakkimsetti:


MohananRahul (approved):

merks:
Perhaps in the future it can be avoided to work on multiple different PRs to address the same problem where I'm helping on one but folks are working on a different one that ignores what I'm doing.  I'm especially concerned about the org.apache.commons.commons-logging inclusion because that has caused major problems that cost me days of work:

https://github.com/eclipse-orbit/orbit-simrel/issues/40

Why exactly is it needed when it is best avoided?

merks:
I strongly advise that you remove it, especially because you added to the feature instead of to the category so you are forcing it to be installed.  If EGit stops working after installing your patch, don't say you weren't warned.

mpalat:
> It's in the 4.33 repository, so again, why is it needed?
> 
@stephan-herrmann any comments here? [ ref: https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/pull/2341#discussion_r1756738614]

stephan-herrmann:
> > It's in the 4.33 repository, so again, why is it needed?
> 
> @stephan-herrmann any comments here? [ ref: https://github.com/[/pull/2341](https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/pull/2341)#discussion_r1756738614]

Answer copied from the other PR:

I'm sorry if commons-logging was a red herring. I believed that was the reason why at one point p2 claimed the feature were not applicable although commonmark was already installed.

If we have a patch build that is installable in pristine SDK 4.33 without stuffing commons-logging into the patch: much better!

Many thanks!!

PR #2352, opened 2024-09-16
Add workflow for automated version increments in pull-requests

HannesWell:
If an Eclipse Plugin project is changed the first time in a development cycle build failures due to missing version bumps are very common, for example like in the following case:
```
 [ERROR] Failed to execute goal org.eclipse.tycho.extras:tycho-p2-extras-plugin:4.0.8:compare-version-with-baselines (compare-attached-artifacts-with-release) on project org.eclipse.pde.api.tools.tests:
Only qualifier changed for (org.eclipse.pde.api.tools.tests/1.3.600.v20240905-2015). Expected to have bigger x.y.z than what is available in baseline (1.3.600.v20240806-1811) -> [Help 1]
```

This PR adds a workflow to automatically detect the need for such service version increments whenever a PR is created or updated, using a fast running Maven build that invokes the `tycho-p2-extras-plugin:compare-version-with-baselines` goal. The required version bumps are then applied by invoking Tycho's `tycho-versions-plugin:bump-versions` goal. The result is committed with the expected message and pushed to the PR's branch using a configurable Eclipse bot account.
Furthermore the configured bot also adds a comment to the affected PR to inform about what happened, to list the files that were changed and to provide a git-patch of the change.

In order to make this possible a personal access token] (PAT) with scope `public_repo`:
- https://docs.github.com/en/authentication/keeping-your-account-and-data-secure/about-authentication-to-github#authenticating-with-the-command-line
- https://docs.github.com/en/authentication/keeping-your-account-and-data-secure/managing-your-personal-access-tokens#personal-access-tokens-classic

With this all set up and running, one does not have to apply version bumps due to qualifier differences anymore and we also don't have to explain it to new users because everything just happens automatically.

An example of the current state can be seen in my PDE fork where I configured my own main account to act as bot-account:
- https://github.com/HannesWell/eclipse.pde/pull/3

This is the major implementation part of https://gitlab.eclipse.org/eclipse-wg/ide-wg/community/-/issues/56. The other parts are to set up the repositories where the new workflows should be called to use the described functionality.
I'll start with the PDE repository and also use it as a verification that this really works. Afterwards I'll set this up for all repositories that are interested.

The next steps are to ask the EF infra team to set up the corresponding tokens described above.

### DIFF
--- a/.github/workflows/checkVersions.yml
+++ b/.github/workflows/checkVersions.yml
@@ -1,0 +1,80 @@
+name: Check for version increments and apply them if necessary 
+
+on:
+  workflow_call:
+    inputs:
+      botName:
+        description: The name of the bot that adds the necessary version increment changes
+        type: string
+        required: true
+      botMail:
+        description: The name of the bot that adds the necessary version increment changes
+        type: string
+        required: true
+
+permissions: {} # all none
+
+jobs:
+  versions-check-and-increment:
+    name: Check and increment service versions
+    runs-on: ubuntu-latest
+    steps:
+
+    - uses: actions/checkout@692973e3d937129bcbf40652eb9f2f61becf3332 # v4.1.7
+      with:
+        fetch-depth: 0 # required for jgit timestamp provider to work
+
+    - name: Set up Java
+      uses: actions/setup-java@6a0805fcefea3d4657a47ac4c165951e33482018 # v4.2.2
+      with:
+        java-version: 17
+        distribution: 'temurin'
+        cache: maven
+
+    - name: Set up Maven
+      uses: stCarolas/setup-maven@d6af6abeda15e98926a57b5aa970a96bb37f97d1 # v5
+      with:
+        maven-version: 3.9.9
+
+    - name: Check and increment versions
+      uses: Wandalen/wretry.action@6feedb7dedadeb826de0f45ff482b53b379a7844 # master
+      with:
+        attempt_delay: 200
+        attempt_limit: 10
+        command: >
+          mvn verify -DskipTests -Dcompare-version-with-baselines.skip=false
+          org.eclipse.tycho:tycho-versions-plugin:bump-versions -Dtycho.bump-versions.increment=100
+          --threads 1C --fail-at-end --batch-mode --no-transfer-progress --show-version
+
+    - name: Commit version increments, if any
+      run: |
+        set -x
+        if [[ $(git status --ignore-submodules --porcelain) != '' ]]; then
+          # Workspace is not clean, i.e. some version were changed
+          
+          # Read 'releaseNumberSDK' property as stream version
+          streamVersion=$(mvn help:evaluate -Dexpression=releaseNumberSDK --quiet -DforceStdout)
+          
+          git config --global user.email '${{ inputs.botMail }}'
+          git config --global user.name '${{ inputs.botName }}'
+          git add --all
+          git status
+          git commit -m "Version bump(s) for ${streamVersion} stream"
+          
+          git format-patch -1 HEAD --no-stat --output 'version_increments.patch'
+          
+          echo '${{ github.event.pull_request.number }}' > 'github_pull_request_number.txt'
+          
+          echo "::error title=Version increments are missing::Required version increments are missing and a commit to apply them is about to be pushed to your PR's branch."
+          exit 1
+        else
+          echo 'No version increments required'
+        fi
+
+    - uses: actions/upload-artifact@50769540e7f4bd5e21e526ee35c689e35e0d6874 # v4.4.0
+      if: always()
+      with:
+        name: versions-git-patch
+        path: |
+          version_increments.patch
+          github_pull_request_number.txt
--- a/.github/workflows/publishVersionCheckResults.yml
+++ b/.github/workflows/publishVersionCheckResults.yml
@@ -1,0 +1,136 @@
+name: Publish Version Check Results
+on:
+  workflow_call:
+    inputs:
+      botGithubId:
+        description: The id of the bot's github account that adds the information comment
+        type: string
+        required: true
+
+    secrets:
+      githubBotPAT:
+        description: The personal access token (with scope 'public_repo') of the bot to push a required change to a PR branch in a fork.
+        required: true
+
+permissions: {} # all none
+
+env:
+  COMMENT_FIRST_LINE: 'This pull request changes some projects for the first time in this development cycle'
+
+jobs:
+  versions-check-result:
+    name: Publish Version Check Results
+    runs-on: ubuntu-latest
+    if: github.event.workflow_run.conclusion != 'skipped'
+    steps:
+
+    - name: Search version increment git patch
+      uses: actions/github-script@60a0d83039c74a4aee543508d2ffcb1c3799cdea # v7.0.1
+      id: search-patch
+      with:
+        script: |
+          let allArtifacts = await github.rest.actions.listWorkflowRunArtifacts({
+             run_id: context.payload.workflow_run.id,
+             ...context.repo
+          })
+          let artifact = allArtifacts.data.artifacts.find(artifact => artifact.name == 'versions-git-patch')
+          return artifact?.id
+
+    - uses: actions/checkout@692973e3d937129bcbf40652eb9f2f61becf3332 # v4.1.7
+      if: steps.search-patch.outputs.result
+      with:
+        ref: '${{ github.event.workflow_run.head_sha }}'
+        persist-credentials: false #Opt out from persisting the default Github-token authentication in order to enable use of the bot's PAT when pushing below
+
+    - name: Download version increment git patch
+      uses: actions/github-script@60a0d83039c74a4aee543508d2ffcb1c3799cdea # v7.0.1
+      id: fetch-patch
+      if: steps.search-patch.outputs.result
+      with:
+        script: |
+          let download = await github.rest.actions.downloadArtifact({
+             artifact_id: ${{ steps.search-patch.outputs.result }},
+             archive_format: 'zip',
+             ...context.repo
+          })
+          let fs = require('fs')
+          fs.writeFileSync(`${process.env.GITHUB_WORKSPACE}/patch.zip`, Buffer.from(download.data))
+          await exec.exec('unzip', ['patch.zip'])
+          let pr_number = Number(fs.readFileSync('github_pull_request_number.txt'))
+          core.setOutput('pull_request_number', pr_number)
+          await io.rmRF('patch.zip')
+          await io.rmRF('github_pull_request_number.txt')
+
+    - name: Apply and push version increment
+      id: git-commit
+      if: steps.search-patch.outputs.result
+      run: |
+          set -x
+          # Set initial placeholder name/mail and read it from the patch later
+          git config --global user.email 'foo@bar'
+          git config --global user.name 'Foo Bar'
+          
+          git am version_increments.patch
+          
+          # Read the author's name+mail from the just applied patch and recommit it with both set as committer
+          botMail=$(git log -1 --pretty=format:'%ae')
+          botName=$(git log -1 --pretty=format:'%an')
+          git config --global user.email "${botMail}"
+          git config --global user.name "${botName}"
+          git commit --amend --no-edit
+          
+          fileList=$(git diff-tree --no-commit-id --name-only HEAD -r)
+          echo "file-list<<EOF" >> $GITHUB_OUTPUT
+          echo "$fileList" >> $GITHUB_OUTPUT
+          echo "EOF" >> $GITHUB_OUTPUT
+          
+          git push \
+            "https://oauth2:${BOT_PA_TOKEN}@github.com/${{ github.event.workflow_run.head_repository.full_name }}.git" \
+            'HEAD:refs/heads/${{ github.event.workflow_run.head_branch }}'
+      env:
+        BOT_PA_TOKEN: ${{ secrets.githubBotPAT }}
+
+    - name: Find existing information comment
+      uses: peter-evans/find-comment@3eae4d37986fb5a8592848f6a574fdf654e61f9e # v3.1.0
+      id: search-comment
+      if: steps.fetch-patch.outputs.pull_request_number
+      with:
+        body-regex: '^${{ env.COMMENT_FIRST_LINE }}'
+        issue-number: ${{ steps.fetch-patch.outputs.pull_request_number }}
+        comment-author: ${{ inputs.botGithubId }}
+        direction: last
+
+    - name: Add or update information comment
+      uses: actions/github-script@60a0d83039c74a4aee543508d2ffcb1c3799cdea # v7.0.1
+      if: always()
+      with:
+        github-token: ${{ secrets.githubBotPAT }}
+        script: |
+          const fs = require('fs')
+          const fileList = `${{ steps.git-commit.outputs.file-list }}`
+          if (fileList) { // if list is empty, no versions were changed
+            const commentBody = `
+          ${{ env.COMMENT_FIRST_LINE }}.
+          Therefore the following files need a version increment:
+          \`\`\`
+          ${fileList}
+          \`\`\`
+          An additional commit containing all the necessary changes was pushed to the top of this PR's branch. To obtain these changes (for example if you want to push more changes) either fetch from your fork or apply the _git patch_.
+          <details>
+          <summary>Git patch</summary>
+          
+          \`\`\`
+          ${ fs.readFileSync( process.env.GITHUB_WORKSPACE + '/version_increments.patch', {encoding: 'utf8'}).trim() }
+          \`\`\`
+          </details>
+          
+          Further information are available in [Common Build Issues - Missing version increments](https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/wiki/Common-Build-Issues#missing-version-increments).
+          `.trim()
+            const prNumber = '${{ steps.fetch-patch.outputs.pull_request_number }}'
+            const existingCommentId = '${{ steps.search-comment.outputs.comment-id  }}'
+            if (existingCommentId) {
+              github.rest.issues.updateComment({...context.repo, comment_id: existingCommentId, body: commentBody })
+            } else {
+              github.rest.issues.createComment({...context.repo, issue_number: prNumber, body: commentBody })
+            }
+          }


eclipse-platform-parent/pom.xml
```diff
@@ -937,6 +922,131 @@
...
+                    <include>feature.xml</include>
+                    <include>MANIFEST.MF</include>
+                  </includes>
+                  <archive>
+                    <manifestFile>${project.build.directory}/MANIFEST.MF</manifestFile>
+                  </archive>
+                </configuration>
+              </execution>
+            </executions>
+          </plugin>
```
HannesWell, line 1012: When https://github.com/eclipse-tycho/tycho/pull/4279 and https://github.com/eclipse-tycho/tycho/pull/4280 respectively their back-ports are available this can be much simpler and specially can work without explicit file manipulations.

.github/workflows/publishVersionCheckResults.yml
```diff
@@ -0,0 +1,124 @@
...
+          }
+
+
+# TODO: Add an FAQ about version bumps?
+# - Why are the version bumps necessary at all?
+#   According to the rules of semantic versioning at least the service version of a bundle should be increased if anything changes between releases.
+# - Why is the service version bumped by +100
+#   Allows service version increments on maintanance branches that stay semnatically smaller than the next releases version
+# - Why a separate commit?
+#   Easier to identify in Git history and the actual change is easier to revert (version must not be reverted)
```
HannesWell, line 123: I intend to add an folded FAQ to the comment added by the bot so that especially newcommers can understand better why these version bumps are necessary and why they are done this way.

The list is currently a brief summary, but if you think questions are missing, please let me know.

.github/workflows/publishVersionCheckResults.yml
```diff
@@ -0,0 +1,124 @@
...
+# TODO: make comment updatable?
```
HannesWell, line 124: I'm contemplating to make the comment added by the bot updatable, to avoid new comments being added after each (force-)push, if the version bump isn't fetched and pushed as well.
But this also means that updates are less visible.
It can for example also happen that due to an update no version bumps are necessary anymore (but the workflow doesn't do anything in this case anyways).

What's your preference?

eclipse-platform-parent/pom.xml
laeubi: What is the benefit of packing a "pseudo-jar" here? Usually skipping tests (and maybe using multiple threads) makes the packing fast enough, trying to replicate the packing to safe a few seconds does not seems very useful given that we perform non trivial task there.

.github/workflows/publishVersionCheckResults.yml
laeubi: I think it must not be configurable, but most confess it a bit unfortunate that we optimize for backports (happen very seldom) and reverts (should actually never happen), anyways a link to the FAQ would be suitable enough for me we don't need a summary replicated on each PR.

.github/workflows/prepareRelease.yml
```diff
@@ -32,9 +32,9 @@ jobs:
         distribution: 'temurin'
         cache: maven
     - name: Update Main Versions
-      run: mvn -U -ntp -f eclipse-platform-parent org.eclipse.tycho:tycho-versions-plugin:4.0.6:set-version -DnewVersion=${{ github.event.milestone.title }}.0-SNAPSHOT -Dmodules=../eclipse.platform.releng.prereqs.sdk
+      run: mvn -U -ntp -f eclipse-platform-parent org.eclipse.tycho:tycho-versions-plugin:set-version -DnewVersion=${{ github.event.milestone.title }}.0-SNAPSHOT -Dmodules=../eclipse.platform.releng.prereqs.sdk
```
laeubi, line 35: Are these changes related to this PR?

.github/workflows/checkVersions.yml
```diff
@@ -0,0 +1,80 @@
...
+
+    - name: Commit version increments, if any
+      run: |
+        set -x
+        if [[ $(git status --ignore-submodules --porcelain) != '' ]]; then
+          # Workspace is not clean, i.e. some version were changed
+          
+          # Read pom project parent version or project version as development stream version
+          sudo apt-get install -qq -y libxml2-utils
+          pomVersion=$(xmllint --xpath "(/*[local-name()='project']/*[local-name()='parent'] | /*[local-name()='project'])/*[local-name()='version']/text()" pom.xml)
```
laeubi, line 58: Why do we need to use special tools, can't we simply use 

https://maven.apache.org/plugins/maven-help-plugin/evaluate-mojo.html

(here is an example https://stackoverflow.com/a/50906059/9503281)

an use either

https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/blob/71bdc744408995313b832a3b9dd1d8224d88b3bf/eclipse-platform-parent/pom.xml#L52

or

https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/blob/71bdc744408995313b832a3b9dd1d8224d88b3bf/eclipse-platform-parent/pom.xml#L67

laeubi:
Instead of having an own build I think one should adopt to what other workflows do (e.g. the junit-publishing workflow), where they work on the result of a previous workflow. e.g if one looks here:

grafik

there is an upload event file step already that contains all the meta-data of the pull request.

Later on, unit test results are uploaded:
grafik

now the [Publish test results workflow](https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/blob/master/.github/workflows/publishTestResults.yml) can read them and act upon these.

The only thing needed is then that we probably upload additional information e.g required version bumps can be written to a file in whatever format suitable, maybe even as a patch file by Tycho and don't need to trigger additional builds at all and we make sure we always reuse what Tycho already offers, this could also probably require less permissions and make the workflow easier to understand.

.github/workflows/publishVersionCheckResults.yml
```diff
@@ -0,0 +1,124 @@
...
+             ...context.repo
+          })
+          let artifact = allArtifacts.data.artifacts.find(artifact => artifact.name == 'versions-git-patch')
+          return artifact?.id
+
+    - uses: actions/checkout@692973e3d937129bcbf40652eb9f2f61becf3332 # v4.1.7
+      if: steps.search-patch.outputs.result
+      with:
+        ref: '${{ github.event.workflow_run.head_sha }}'
+        persist-credentials: true # (Explicit) opt-in for persisting the bot's PAT for authentication when pushing below
```
laeubi, line 37: I'm not sure this is something we want here if it runs on PRs submitted by users from [this comment](https://github.com/actions/checkout/issues/485#issuecomment-1197047674):

> because it means any workflow using actions/checkout basically leaks the token to any process/action in that workflow which can just read it from .git/config.

Can't we use

https://github.com/marketplace/actions/create-pull-request

 instead that already supports all the handling of tokens? According to the docs one can also prepare the commit manually:
https://github.com/marketplace/actions/create-pull-request#create-your-own-commits

one also can push to forked repositories:
https://github.com/peter-evans/create-pull-request/blob/main/docs/concepts-guidelines.md#push-pull-request-branches-to-a-fork

.github/workflows/updateRelease.yml
```diff
@@ -24,19 +24,8 @@ jobs:
...
-        attempt_delay: 120000
-        attempt_limit: 10
-        command: >-
-            mvn -U -Pbuild-individual-bundles -ntp
-            clean verify
-            -DskipTests
-            -Dcompare-version-with-baselines.skip=false
-            org.eclipse.tycho:tycho-versions-plugin:4.0.6:bump-versions -Dtycho.bump-versions.increment=100
+          org.eclipse.tycho:tycho-versions-plugin:set-version -DnewVersion=${{ github.event.milestone.title }}.0-SNAPSHOT
+          org.eclipse.tycho:tycho-versions-plugin:set-parent-version -DnewParentVersion=${{ github.event.milestone.title }}.0-SNAPSHOT
```
laeubi, line 28: Is this related too this PR?

eclipse-platform-parent/pom.xml
```diff
@@ -768,33 +780,6 @@
        </repository>
      </repositories>
    </profile>
-   <profile>
```
laeubi, line 771: Why is this removed?

.github/workflows/prepareRelease.yml
HannesWell: Indirectly.
It becomes possible due to a change that's required for this PR. Respectivly I want to avoid for the new workflow having the need to hard-code the version again.

But I have moved this into a separate, isolated PR:
- https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/pull/2354

.github/workflows/checkVersions.yml
HannesWell: Yes we can absolutely use that. I just didn't knew that this exists.
Actually I wanted to use the value of `releaseNumberSDK` in the first place, but could access it in an elegant way, so the current approach is more a work-around since these versions are usually in sync.

Thanks for pointing this out, will use your suggestion instead.

.github/workflows/updateRelease.yml
HannesWell: Similar to https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/pull/2352#discussion_r1763576838 ?

.github/workflows/publishVersionCheckResults.yml
HannesWell: > I'm not sure this is something we want here if it runs on PRs submitted by users from [this comment](https://github.com/actions/checkout/issues/485#issuecomment-1197047674):
> 
> > because it means any workflow using actions/checkout basically leaks the token to any process/action in that workflow which can just read it from .git/config.

Yes this is something I also wanted to discuss, but in general it should not be very dangerous because this happens only in the workflow triggered upon the completion of the first workflow that runs the build to determine the required changes. And that second workflow only applies the git-patch of these changes and pushes it.
So no changes supplied from external persons _should_ be executed in any form.

> Can't we use
> 
> https://github.com/marketplace/actions/create-pull-request

Unfortunately not because this workflow pushes an extra commit to an already existing PR and does not aim to create a new one (as far as I understand that's not possible with the action you mention?).
However there are other actions just for pushing, but in the end they just push via https using basic-auth:
https://github.com/ad-m/github-push-action/blob/77c5b412c50b723d2a4fbc6d71fb5723bcd439aa/start.sh#L52

I have now changed this part to do the same when pushing. This way the token is only supplied to the `run` step as an environment variable. Initially I implemented it this way, but changed it to persisted credentials since it seems a bit simpler and as described above, the second workflow should actually not be critical.
If you think it's necessary I can also move the push part to a separate run step to further narrow the availability of the token.

eclipse-platform-parent/pom.xml
HannesWell: It was just a work-around because the two tycho changes mentioned above were not yet available.
Now it's just an invocation of two tycho-plugins configured to do the minimally necessary, plus an `antrun` invocation to set a few properties based on the project's packaging type.

Of course it would be even very nice if just the `tycho-p2-extras-plugin:compare-version-with-baselines` goal could be invoked without a preceding `tycho-p2-plugin` and `tycho-packaging-plugin`, but that seemed to be more work (although it was also more work than expected to strip that all down).

> Usually skipping tests (and maybe using multiple threads) makes the packing fast enough, trying to replicate the packing to safe a few seconds does not seems very useful given that we perform non trivial task there.

This not only saves a few seconds. With the current set up a build takes in the PDE example about 45s, while a full build even if executed in parallel and without tests takes about 5min (on my machine). If it is considers that in case of required version bumps at least one more build is required this is already a difference of two vs. 10min response time.
In the PDE example presented above, even three build runs are necessary so the response time that's currently between three/four minutes (the workflow has an overhead) would go up to ~15min. Which is also added to the overall build time.
And I think the response time should be fast.

If this does not work anymore we can still go back to `mvn clean verify -DskipTests -T 1C -Dcompare-version-with-baselines.skip=false` or make `tycho-p2-extras-plugin:compare-version-with-baselines` work with only a qualifier being computed before.

.github/workflows/publishVersionCheckResults.yml
HannesWell: Was this intended for https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/pull/2352#discussion_r1762064757?

But sure providing the FAQ in the wiki and linking it is of course also a good solution. It would be folded, just like the git-patch, but with a link one can always read the latest version.
I can implement it that way.

> but most confess it a bit unfortunate that we optimize for backports (happen very seldom) and reverts (should actually never happen)

I wouldn't say we optimize it, but with the workflow it also just technically simpler. And just as a side-note: since these things happen seldomly it's actually better to have less things to consider because otherwise mistakes are more likely.

HannesWell:
> Instead of having an own build I think one should adopt to what other workflows do (e.g. the junit-publishing workflow), where they work on the result of a previous workflow. e.g if one looks here:
> [...]
> The only thing needed is then that we probably upload additional information e.g required version bumps can be written to a file in whatever format suitable, maybe even as a patch file by Tycho and don't need to trigger additional builds at all and we make sure we always reuse what Tycho already offers, this could also probably require less permissions and make the workflow easier to understand.

First of all I started out simple and wanted to keep it simple, but unfortunately it turned out that if you want to push to a PR-branch in a fork from a GitHub workflow, things become complicated. I tried out a lot and tried to keep it as simple as possible.
The main problems are that first the GITHUB_TOKEN has no permission to push to forks, even if I grant `write-all` permissions, so a PAT is necessary for pushing. And second, workflows triggered by the `pull_request` event from a fork of a non-committer (i.e. an account without write permissions) don't have access to the secret store of a repo.
Therefore a second workflow is necessary that runs upon completion of the `pull_request` even triggered workflow, similar to the workflow to publish test results (although it's a bit different there because that workflow only adds a comment, but that's also not possible for a workflow triggered by a PR from a fork).

It would be possible to have the version bumps applied as part of the regular verification builds, but that would only save us one extra git-checkout step plus the configuration of a JDK and Maven as well as that we have a separate file. But all other steps and inputs etc. done in the first workflow still would have to be done and the second workflow would still be necessary at it is.
This moderate simplification of the configuration would come with the draw-back that each iteration would take the full time of the verification build. In the PDE example this would mean ~30min per iteration instead of 45s. In the linked example this would mean a total wait-time of roughly 1.5h until the PR is completely built instead of ~3-4min extra time.
For me this justifies the current setup.

Having separated builds also helps because it also helps to separate concerns and avoids that we collect everything in the `mavenBuild.yml`.

eclipse-platform-parent/pom.xml
HannesWell: With this new workflow to apply and push required version increments automatically to a PR, this seems not to be necessary anymore since the new workflow will handle this.
So removing this and the version-bump part in the `updateRelease.yml` reduce amount of configuration and avoids that duplicated configuration goes out of sync (e.g. if we want other version-bump deltas in the future).

But I moved all removals to separate commit in this PR to make this more clear.

.github/workflows/publishVersionCheckResults.yml
HannesWell: As discussed in https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/pull/2352#discussion_r1762273259 ff the comment added by this workflow now just links to a new section in the wiki that I have just created:
https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/wiki/Common-Build-Issues#missing-version-increments

HannesWell:
I updated the workflow to implement the missing TODOs, which besides others, include the addition of a link to a new [wiki entry about version-bumps](https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/wiki/Common-Build-Issues#missing-version-increments) and making the workflow update the comment it added in a previous run, instead of adding a new comment for each run (if e.g. a new change is pushed but the auther overwriting/reverting the applied version bump).

I have also done further testing with
- https://github.com/HannesWell/eclipse.pde/pull/3

and everything looks perfectly well.
I also hope that all concerns were addressed.

With that for me this workflow is ready for submission.

To really use it (e.g. like set up in https://github.com/eclipse-pde/eclipse.pde/pull/1408), we have to await the addition of the corresponding tokens by the EF infra team.

jukzi:
I like the idea, but i have doubts about the benefit of this change as long as the API tooling is flaky https://github.com/eclipse-pde/eclipse.pde/issues/1193.
Recent example:
https://github.com/eclipse-platform/eclipse.platform/actions/runs/10961494829/job/30438556401?pr=1562
false positive:
```
IPasswordStore has been removed from org.eclipse.jsch.core_1.5.500
META-INF/MANIFEST.MF:5 The major version should be incremented in version 1.5.500, since API breakage occurred since version 1.5.500
```
As far as i understand this workflow would just add a wrong version bump that would have to be manually reverted.

merks:
I would hope the tool is configurable enough that it wouldn't suggest major increments because we basically never to do that the platform's projects, and the IDE would have complained already.    I expect the +100 bumps are the things the IDE (API tools) never notice...

HannesWell:
This workflow does not use API-Tools at all, it just uses the [tycho-p2-extras:compare-version-with-baselines](https://tycho.eclipseprojects.io/doc/4.0.9/tycho-extras/tycho-p2-extras-plugin/compare-version-with-baselines-mojo.html) goal and therefore detects if anything has changed compared to the baseline (should always be the latest release) and it only suggests the `+100` service/micro version-segment bump.
This works quite reliably and I have never seen a false positive in that.

It would be possible to automatically perform the version increments required by API-tooling too, in fact Tycho already supports this and it would just have to be enabled. But as you said API-Tools are not as reliable and it would also increase the runtime of the check significantly.
But first of all I think version increments due to API changes should be a decision made deliberately and in full awareness plus it's already checked and indicated in the IDE. On the other hand, the service-version increments are something that has to be done on any first change in a release cycle and something one does not really get around and the IDE doesn't check that (yet).

eclipse-platform-parent/pom.xml
laeubi: As one can use the VB profile also locally I don't thing it is really redundant

.github/workflows/updateRelease.yml
laeubi: The idea is to have as much as possible for going for a new release to be included in one commit.

eclipse-platform-parent/pom.xml
HannesWell: That's right. But is this really used (by many)?

Anyway I moved this change to https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/pull/2387, so we can later discuss there if it should be removed or retained.

.github/workflows/updateRelease.yml
HannesWell: Does it really matter if it's in one commit or not?
For everything changed later we are also fine with it and it has some advantages. And even if it doesn't have an advantage to have it separated, having less configuration to maintain is IMHO one.

But anyway I moved this change to https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/pull/2387, so we can later discuss there if it should be removed or retained.

HannesWell:
After some more testing with other repos of the SDK revealed that the extra profile for an expedited version check does not work on all SDK projects I have now changed the workflow to just execute an ordinary verification build with tests execution disabled (and no extra api-checks etc. enabled) that runs in parallel.
I will later work on a Tycho mojo geared for this use-case that should make the workflow faster.

With that this is now ready for submission since also the infrastructure of PDE is ready for a first usage.
Thank you all for the discussion.

HannesWell:
The real world test in PDE looks good:
- https://github.com/eclipse-pde/eclipse.pde/pull/1413

@akurtakov or @iloveeclipse, do you as PL want to have this for Eclipse-Platform?
@tjwatson or @laeubi, do you as PL want to have this in Equinox?
@jarthana do you as PL want to have this for JDT?
Similar like it was set up for PDE in https://github.com/eclipse-pde/eclipse.pde/pull/1408.

I'm asking all of you together because then I can request the necessary changes, i.e. the creation of the Personal-Access-Token for the bot-user, all together from the EF infra-team.

akurtakov:
This sounds like quite handy enhancement so +1 for Platform repositories.

tjwatson:
Sounds good to me but would like @laeubi to comment also before applying to Equinox.  Thanks!

laeubi:
Lets give it a try!

iloveeclipse:
@HannesWell : also please for JDT.

HannesWell:
Great!

Just created for Platform:
- https://github.com/eclipse-platform/eclipse.platform/pull/1568
- https://github.com/eclipse-platform/eclipse.platform.ui/pull/2326
- https://github.com/eclipse-platform/eclipse.platform.swt/pull/1491

an for Equinox:
- https://github.com/eclipse-equinox/equinox/pull/678
- https://github.com/eclipse-equinox/p2/pull/549

We have to await the addition of the PATs which was requested in https://gitlab.eclipse.org/eclipsefdn/helpdesk/-/issues/5020#note_2825941 ff.

> @HannesWell : also please for JDT.

Yes. Just requested the PAT as well and will follow-up with corresponding PRs for JDT tomorrow.

HannesWell:
> @HannesWell : also please for JDT.

@iloveeclipse please see the following PRs for JDT:
- https://github.com/eclipse-jdt/eclipse.jdt.core/pull/3021
- https://github.com/eclipse-jdt/eclipse.jdt.debug/pull/516
- https://github.com/eclipse-jdt/eclipse.jdt.ui/pull/1680

PR #2355, opened 2024-09-18
Fix running smoke_test_ubuntu

akurtakov:


akurtakov:
Verification build will not test it thus merge.

PR #2356, opened 2024-09-18
Update that target platform to a new Orbit milestone

merks:
- Update com.sun.jna/jna-platform to 5.15.0.
- Update org.eclipse.orbit.xml-apis-ext to new qualifier version.
- Update org.commonmark to 0.23.0.


merks:
@stephan-herrmann 

FYI, note the update to the commonmark dependencies.

stephan-herrmann:
> @stephan-herrmann
> 
> FYI, note the update to the commonmark dependencies.

Thanks for the notice!

Doesn't look like it would break anything.

In fact there's a new feature that could perhaps have helped for Java 23: "Support for custom link processing during inline parsing (e.g. [foo])" :smile:

PR #2357, opened 2024-09-18
Get out of ubuntu 22 image building

akurtakov:
EF provided 24.04 is used in smoke tests.


HannesWell:
As discussed privately it would be good to await https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/pull/2271 (which is hopefully submitted tomorrow) so that we have the required change in the Jenkins job to build the images in this PR too.

In the meantime, I think the image definition itself should be removed too, shouldn't it?
I.e. I think the entire directory should be deleted at
https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/tree/master/cje-production/dockerfiles/ubuntu-gtk3-metacity

akurtakov:
> As discussed privately it would be good to await #2271 (which is hopefully submitted tomorrow) so that we have the required change in the Jenkins job to build the images in this PR too.
> 
> In the meantime, I think the image definition itself should be removed too, shouldn't it? I.e. I think the entire directory should be deleted at https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/tree/master/cje-production/dockerfiles/ubuntu-gtk3-metacity

Yes. That one should be deleted too.

HannesWell:
Updated this PR to the top of the master and removed the suggested files too.

HannesWell (approved):
This should now be complete but is not tested by the verification build, so there is no need to wait for a successful build.

PR #2362, opened 2024-09-19
Stop executing perf tests

akurtakov:
As discussed in
https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/discussions/828#discussioncomment-10695397 and https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/discussions/2264 this only takes precious resources and no one actually looks at these results.


akurtakov:
Merge directly as this would not be verified by verification builds.

PR #2363, opened 2024-09-19
Run I-built tests on Java 23 on Linux

akurtakov:


akurtakov:
This is executed only by I-builds thus merging and starting new I-build to see whether it works.

PR #2365, opened 2024-09-19
Enable standard I-build schedule

akurtakov:
Having it commented in the jenkinsfile made it being no standard time build.


akurtakov:
Yet another non-testable by verification build change.

PR #2367, opened 2024-09-20
Stop running and publishing Java 22 tests

akurtakov:
Java 22 is EOL now that Java 23 is released.


akurtakov:
No need to wait for verification build as it doesn't test these jobs.

PR #2368, opened 2024-09-20
provide junit ScreenshotOnFailure TestWatcher

jukzi:


HeikoKlare:
May this be better placed in [`org.eclipse.core.tests.harness`](https://github.com/eclipse-platform/eclipse.platform/tree/master/runtime/tests/org.eclipse.core.tests.harness), since it is not related to performance test?

akurtakov:
> May this be better placed in [`org.eclipse.core.tests.harness`](https://github.com/eclipse-platform/eclipse.platform/tree/master/runtime/tests/org.eclipse.core.tests.harness), since it is not related to performance tests?

Probably even better place is https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/tree/master/eclipse.platform.releng/bundles/org.eclipse.test .

akurtakov (changes requested):
Please move to org.eclipse.test bundle.

jukzi:
Moving it would be a hell of work. Moving the class would be an API breakage. TracingSuite requires it needs to be fixed, Access restrictions need to be fixed... . I don't like how it is either, but to clean up that mess seems too complicated :-(

akurtakov:
Tests are not API incl org.eclipse.test. Does it make the task easier?

jukzi:
technically they are API. Some tests just do not include org.eclipse.test but the performance one. so no it's not helpful.

akurtakov:
Everything should be in o.e.test without any breakage so you can patch Screenshots in o.e.test bundle.

akurtakov:
Tests in I-build seems to be in good shape now.

akurtakov (approved):
Looks good to me now.

jukzi:
Not only seems that build error unrelated but should'nt it have produced an automated version bump @HannesWell ?
`org.eclipse.jdt.doc.user: Only qualifier changed for (org.eclipse.jdt.doc.user/3.15.2300.v20240927-0605). Expected to have bigger x.y.z than what is available in baseline (3.15.2300.v20240823-1014) -> [Help 1]`

akurtakov:
It is caused by https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/pull/2392

HannesWell:
> Not only seems that build error unrelated but should'nt it have produced an automated version bump @HannesWell ? `org.eclipse.jdt.doc.user: Only qualifier changed for (org.eclipse.jdt.doc.user/3.15.2300.v20240927-0605). Expected to have bigger x.y.z than what is available in baseline (3.15.2300.v20240823-1014) -> [Help 1]`

The version-increment workflow is not (yet) setup for this repository. In the current form it would be basically the same as the verification build. I plan to work on a Tycho mojo that should speed that up significantly but in the meantime I think it's not worth the effort to enable it in this repo.

PR #2370, opened 2024-09-20
[Docker-Images] Skip default checkout and limit registry availability

HannesWell:


HannesWell:
Verification build does not test this. Verified it by replaying the affected build.

PR #2373, opened 2024-09-22
Remove .classpath attributes about containing only test source

merks:
- The downstream dependencies in the workspace end up with a sea of red.


merks:
My workspace ended up with a sea of errors:

image

I tried changing downstream projects but the osgi test project I couldn't figure out how to fix.  And I'm not sure the attribute is really correctly set in the first place.  These projects aren't test but provide infrastructure other tests.

Anyway, if this is not the "right-thing-to-do" it can be changed later, but then all the downstream projects need to be revisited too...

PR #2360, opened 2024-09-19
Update to Tycho 4.0.9

laeubi:

PR #2374, opened 2024-09-22
Temporarily use 2024-06 API-Tools for API checks in verification builds

HannesWell:
Revert to use previous release's API-Tools to hopefully unblock verification builds that are currently suffering from https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/pull/2360#issuecomment-2364207760.

If this doesn't help, the problem is probably not directly within PDE's API-Tools but in the integration with Tycho.

### DIFF
--- a/eclipse-platform-parent/pom.xml
+++ b/eclipse-platform-parent/pom.xml
@@ -923,6 +923,9 @@
                             <goal>verify</goal>
                         </goals>
                         <configuration>
+                             <apiToolsRepository>
+                                 <url>https://download.eclipse.org/eclipse/updates/4.32/R-4.32-202406010610/</url>
+                             </apiToolsRepository>
                              <baselines>
                                  <repository>
                                      <url>${previous-release.baseline}</url>


HannesWell:
If this does not help, the only candidates on the Tycho site I see are
- https://github.com/eclipse-tycho/tycho/pull/4105
- https://github.com/eclipse-tycho/tycho/pull/4196

HannesWell:
Ok, so with this applied and the `api-checks` profile enabled it still happens that the `tycho-apitools:verify` mojo hangs and I therefore assume https://github.com/eclipse-tycho/tycho/pull/4105 has something to do with it.
But since I also saw false positive errors/warnings I assume that the underlying PDE Api-Tools als cause some troubles. Therefore I suspect that this could still fix some problems in verification builds. At least it worked to check a few plug-ins and also worked with tycho 4.0.8 and I didn't saw a change that should be incompatible to the old api-tools version.
So I hope that in the worst case, this simply does not help.

laeubi:
> Ok, so with this applied and the api-checks profile enabled it still happens that the tycho-apitools:verify mojo hangs 

Is there a way to reproduce the problem e.g. can one build a single project to see the hang or does it only happen randomly?

HannesWell:
> > Ok, so with this applied and the api-checks profile enabled it still happens that the tycho-apitools:verify mojo hangs 
> 
> Is there a way to reproduce the problem e.g. can one build a single project to see the hang or does it only happen randomly?

I didn't see a systematic schema so I think it's random. But when I run the aggregator build it usually hangs after a few projects have been built.

HannesWell:
Btw. if we don't find another immediate remediation, couldn't we simply use only the apitools plug-in in version 4.0.8?

iloveeclipse:
Could it be, that jenkins builds are failing now because something is not in-sync with the pom config?
See https://ci.eclipse.org/platform/job/eclipse.platform.ui/job/PR-2288/4/console for recent build failure
for PR that was green before rebase: https://github.com/eclipse-platform/eclipse.platform.ui/pull/2288

HannesWell:
> Could it be, that jenkins builds are failing now because something is not in-sync with the pom config? See https://ci.eclipse.org/platform/job/eclipse.platform.ui/job/PR-2288/4/console for recent build failure for PR that was green before rebase: [eclipse-platform/eclipse.platform.ui#2288](https://github.com/eclipse-platform/eclipse.platform.ui/pull/2288)

I cannot really say anything with certainty. I assume this didn't make it worse, but I cannot guarantee it.
But I have started another attempt for a work-around to unblock verification builds:
- https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/pull/2380

iloveeclipse:
> > Could it be, that jenkins builds are failing now because something is not in-sync with the pom config? See https://ci.eclipse.org/platform/job/eclipse.platform.ui/job/PR-2288/4/console for recent build failure for PR that was green before rebase: [eclipse-platform/eclipse.platform.ui#2288](https://github.com/eclipse-platform/eclipse.platform.ui/pull/2288)
> 
> I cannot really say anything with certainty. I assume this didn't make it worse, but I cannot guarantee it. But I have started another attempt for a work-around to unblock verification builds:
> 
> * [[Build] Temporarily use tycho-apitools:verify mojo in version 4.0.8 #2380](https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/pull/2380)

The funny thing is that PR build failed 3 times with different issues (two times due API errors) and after that it succeeded without any change in the code. So I assume we have (had?) some infrastructure issues.

HannesWell:
> The funny thing is that PR build failed 3 times with different issues (two times due API errors) and after that it succeeded without any change in the code. So I assume we have (had?) some infrastructure issues.

That's also possible, but as far as I can tell #2380 improved the situation for now.

PR #2369, opened 2024-09-20
Reduce split packages in o.e.tests

akurtakov:
Split packages cause a lot of troubles and this is test harness code so it's better to live without them.

PR #2377, opened 2024-09-23
Remove package import of o.e.test from o.e.test.performance

merks:
- It already requires bundle org.eclipse.test so the import seems redundant and causes problems running tests.

https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/pull/2369

### DIFF
--- a/eclipse.platform.releng/bundles/org.eclipse.test.performance/META-INF/MANIFEST.MF
+++ b/eclipse.platform.releng/bundles/org.eclipse.test.performance/META-INF/MANIFEST.MF
@@ -7,8 +7,7 @@ Bundle-Activator: org.eclipse.test.internal.performance.PerformanceTestPlugin
 Bundle-Vendor: %Plugin.providerName
 Bundle-Localization: plugin
 Import-Package: org.junit.jupiter.api,
- org.junit.platform.suite.api,
- org.eclipse.test
+ org.junit.platform.suite.api
 Export-Package: org.eclipse.perfmsr.core,
  org.eclipse.test.internal.performance,
  org.eclipse.test.internal.performance.data,


merks:
@jukzi 

Will this build confirm that this fixes the problem?  Or is there another way to confirm that this fixes the problem?

jukzi:
> Will this build confirm that this fixes the problem?

no, the test is in another repo. but i can locally try.

jukzi:
does not fix org.eclipse.core.tests.internal.runtime.PlatformURLSessionTest

jukzi:
`java.lang.IllegalArgumentException: Bundle "org.eclipse.core.tests.harness" not found. Possible causes include missing dependencies, too restrictive version ranges, or a non-matching required execution environment.
`

merks:
So the error message is different (not missing package import) but even more "unhelpful"...

jukzi:
my best guess is that all plugins that did require one of the affected bundles need to require updated version of those bundles.

merks:
This looks suspicious to me, optional, and not greedy:

image

I remove it and there are errors, so I'm not sure how it's optional or what it being optional brings us:

image

merks:
There's really nothing using the o.e.test package as far as I can see except in the plugin.xml of its own bundle:

image

merks:
Does making the dependency non-optional make a difference for the test?  Is there no other failure message that might give a bit more of a clue what it doesn't like?

merks:
@jukzi @akurtakov 

Shall I commit this?  I think the import is redundant and unhelpful...

HeikoKlare:
Looks like the problems are about `ConfigurationSessionTestSuite` instances. That one needs to have explicitly defined the dependencies for the remote Eclipse instance started during test runs. It means, the according dependencies need to be explicitly added in `ConfigurationSessionTestSuite#addMinimalBundleSet()` (via specifying one of the classes in the the plugin you depend on). The `org.eclipse.test` dependency is not mentioned yet, but is required by the `org.eclipse.core.tests.harness` plugin used for the session tests.

laeubi:
@merks often things are required transitive (e.g. though method signatures/return types) that are not obvious in the code, e.g. if I have

`myObject.getX().getY()`

then you won't find a (direct) reference to the return type of `getX` (what we can assume is `X`) but if `X` is from another bundle you still have to import it. ... Making it optional do not seem useful here.

iloveeclipse:
The problem is visible in https://github.com/eclipse-platform/eclipse.platform/issues/1564.
I could see the test fails locally, and this PR doesn't fix them.

iloveeclipse (changes requested):
See https://github.com/eclipse-platform/eclipse.platform/issues/1564#issuecomment-2368642798.

This PR can be closed.

akurtakov:
Closing as the full set of changes to not requrie and reexport is there anymore.

PR #2380, opened 2024-09-23
[Build] Temporarily use tycho-apitools:verify mojo in version 4.0.8

HannesWell:
Temporarily revert the `tycho-apitools:verify` mojo to version 4.0.8 for API checks to hopefully unblock verification builds that are currently suffering from https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/pull/2360#issuecomment-2364207760.


HannesWell:
At least for eclipse.platform and eclipse.platform.ui it looks like the apitools verification now works again.

PR #2381, opened 2024-09-23
Add overwritable indirection for flag to skip version baseline check

HannesWell:
This can for example be used in https://github.com/eclipse-jdt/eclipse.jdt.core/pull/2994 and can be applied to future version-baseline check mojos that are probably necessary for
- https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/pull/2352


akurtakov:
Heap memory issue which shouldn't be related to this change. Merging.

PR #2386, opened 2024-09-24
Use 1.5.1 snapshot cbi plugin

akurtakov:
It's a prep step decoupling cbi plugin from using tycho internals as tycho 5.0 changed them in prep for Maven 4.


akurtakov:
See https://github.com/eclipse-cbi/org.eclipse.cbi/pull/542

PR #2387, opened 2024-09-24
[Builds] Remove elements unnecessary with the new version-bump workflow

HannesWell:
With the new reusable workflow for automated version increments in pull-requests, introduced with https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/pull/2352, it is not necessary anymore to bump versions in advance during release preparation or the enable bumping if the property 'compare-version-with-baselines.skip' is set to false. Everything now happens automatically and these elements can therefore be removed.

Removing these elements avoids partly duplicated configuration and generally reduced the amount of configuration.

### DIFF
--- a/.github/workflows/updateRelease.yml
+++ b/.github/workflows/updateRelease.yml
@@ -26,17 +26,6 @@ jobs:
           mvn -U -Pbuild-individual-bundles -ntp
           org.eclipse.tycho:tycho-versions-plugin:set-version -DnewVersion=${{ github.event.milestone.title }}.0-SNAPSHOT
           org.eclipse.tycho:tycho-versions-plugin:set-parent-version -DnewParentVersion=${{ github.event.milestone.title }}.0-SNAPSHOT
-    - name: Build and Bump Versions
-      uses: Wandalen/wretry.action@ffdd254f4eaf1562b8a2c66aeaa37f1ff2231179 # master
-      with:
-        attempt_delay: 120000
-        attempt_limit: 10
-        command: >-
-            mvn -U -Pbuild-individual-bundles -ntp
-            clean verify
-            -DskipTests
-            -Dcompare-version-with-baselines.skip=false
-            org.eclipse.tycho:tycho-versions-plugin:bump-versions -Dtycho.bump-versions.increment=100
     - name: Create Pull Request for Release ${{ github.event.milestone.title }}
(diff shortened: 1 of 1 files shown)


akurtakov:
As thenew workflow is not yet setup on this repo, this one can not be pushed, right?

HannesWell:
> As thenew workflow is not yet setup on this repo, this one can not be pushed, right?

This repository is not really the problem since the removed elements are (hardly) used here, but we should await until it's active in all other (active) SDK repositories. And while JDT should only be a matter of committer time, Equinox and SWT seem to need more adjustments and might take longer.

eclipse-platform-parent/pom.xml
```diff
@@ -774,32 +774,6 @@
        </repository>
      </repositories>
    </profile>
-   <profile>
-      <!-- This provile enables automatic version bumps when running the build -->
```
laeubi, line 778: As mentioned last time this can be used locally as well, so I would prefer to keep the profile

HannesWell: Yes it can be used. But I wonder if it is really used by a relevant number of people?
Personally I have never used it and didn't even know about before working in this area. Do you use it or know anybody that uses it?

Now that we have it automated at GH, I think this particular profile became even less relevant.

HannesWell: I have now split this part into a separate PR so that the other part can be used:
- https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/pull/2570

akurtakov:
Landing this part in.

HannesWell:
> Landing this part in.

To add some context, why this is important now: 
The updateRelease workflow currently suffers the same problems like the workflow for automated version increment, which is why these two repos don't have it yet:
- https://github.com/eclipse-equinox/equinox/pull/678
- https://github.com/eclipse-platform/eclipse.platform.swt/pull/1491

With this being submitted, Equinox and SWT still don't get the version-bumps on first change in the dev-cycle, but at least changing the parent-pom version should be possible now.

HannesWell:
For Equinox and SWT the following PRs could now be created, after renaming the existing `4.35` milestone creating a new one with the same original data and afterwards deleting the new one and restoring the name of the original one:
- https://github.com/eclipse-equinox/equinox/pull/701
- https://github.com/eclipse-platform/eclipse.platform.swt/pull/1614

PR #2396, opened 2024-09-27
Use temuring-jdk17-latest for I-builds

akurtakov:
As openjdk-jdk17-latest is stuck on older version.


akurtakov:
Should help with https://github.com/eclipse-pde/eclipse.pde/issues/1419

akurtakov:
As this would not be verified - merging and starting new I-build.

PR #2398, opened 2024-09-27
[Build] Define cron-trigger for DockerImagesBuild in Job definition

HannesWell:
The triggers defined in a Jenkins-pipeline are only applied to the Job-configuration after the first execution of the pipeline. This has the consequence that after each execution of the seed job the configuration of the 'DockerImagesBuild' is reset to have no triggers and consequently it would require a manual execution of that job to restore the cron-trigger each execution of the seed job.


HannesWell:
This is not verified at all by verification builds so I submit this immediately and re-run the seed job in the [`RelEng` JIPP](https://ci.eclipse.org/releng/).

PR #2400, opened 2024-09-28
Stop dependabot docker for centos 8

akurtakov:
This is gone now and these images are no longer built.


akurtakov:
No verification for dependabot config changes. Merging.

PR #2402, opened 2024-09-28
[Build] Commit only relevant files in version increment workflow

HannesWell:
Do not just commit every changed file in order to avoid accidental commits of resources changed or generated during builds. For version increments only the following files are relevant:
- META-INF/MANIFEST.MF
- feature.xml
- pom.xml

This became relevant for example in https://github.com/eclipse-jdt/eclipse.jdt.debug/pull/516#issuecomment-2377878801.


HannesWell:
This is not tested by the verification build, but I tested it in my test-setup of this workflow.

Submitting.

PR #2403, opened 2024-09-28
[Build] Make staging of relevant files robust against file absence

HannesWell:
Stage the relevant kind of files in isolation and don't fail if that kind of file does not exist at all in the repository.

Otherwise the entire step fails if the repo for example doesn't contain a feature.xml, because the previous 'git add' command then fails with:
`fatal: pathspec '*/feature.xml' did not match any files`

This became relevant for example in https://github.com/eclipse-jdt/eclipse.jdt.debug/pull/516

### DIFF
--- a/.github/workflows/checkVersions.yml
+++ b/.github/workflows/checkVersions.yml
@@ -49,7 +49,10 @@ jobs:
     - name: Commit version increments, if any
       run: |
         set -x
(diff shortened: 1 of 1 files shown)


HannesWell:
Tested with https://github.com/eclipse-jdt/eclipse.jdt.debug/pull/516 and it worked as expected.

PR #2392, opened 2024-09-26
Align Case-Ing to make topic titles consistent

BeckerWdf:
- Platform and PDE "Development User Guide" instead of "development user guide".
- "Getting started" is used in stead of "Getting Started" by the other projects.

PR #2404, opened 2024-09-29
Fix broken verification build

akurtakov:
Caused by
https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/pull/2392


BeckerWdf:
I am away from my keyboard this weekend.
Thanks for fixing it.

PR #2405, opened 2024-09-29
Remove org.hamcrest.core from Eclipse SDK

merks:
- The org.hamcrest.core bundle is just an empty hull to provide metadata that redirects all references to it to the new org.hamcrest bundle that avoids split packages with the org.hamcrest.library bundle by combining all the content into a single bundle.
- Nothing in the SDK should require org.hamcrest.core anymore, so it can be removed.
- A new org.junit bundle from Orbit is needed that requires the org.hamcrest bundle instead of the org.hamcrest.core bundle.

https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/pull/2343

### DIFF
--- a/eclipse.platform.releng.prereqs.sdk/eclipse-sdk-prereqs.target
+++ b/eclipse.platform.releng.prereqs.sdk/eclipse-sdk-prereqs.target
@@ -23,8 +23,10 @@
       <unit id="org.apache.xmlgraphics.source" version="2.9.0.v20230916-1600"/>
 
       <!-- needed because org.eclipse.jdt.feature.group  includes it, but it should include org.hamcrest (direct-from-maven) instead --> 
+      <!--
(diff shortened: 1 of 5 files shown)


merks:
@HannesWell 

I wanted to push this head by using a nightly build of the new org.junit bundle so that we can verify that this approach is working well before inflicting it on the broader ecosystem via milestone builds, which I want to avoid doing until right after M1 completes  EOD Wednesday.

HannesWell:
> I wanted to push this head by using a nightly build of the new org.junit bundle so that we can verify that this approach is working well before inflicting it on the broader ecosystem via milestone builds, which I want to avoid doing until right after M1 completes EOD Wednesday.

I'm perfectly fine with that. Thank you.

HannesWell (approved):

PR #2407, opened 2024-09-30
Eclipse Platform Binary: Register eclipse+command links for macOS

sratz:
With
https://github.com/eclipse-platform/eclipse.platform.ui/issues/1901 the auto-registration of link handlers was removed to avoid breaking of code signatures.

In order to use the 'eclipse+command' URI schemes on macOS, this must instead be pre-registered during .app bundle build time before code signing.

### DIFF
--- a/eclipse.platform.releng.tychoeclipsebuilder/eclipse.platform.repository/platform.product
+++ b/eclipse.platform.releng.tychoeclipsebuilder/eclipse.platform.repository/platform.product
@@ -19,7 +19,11 @@
       location="org.eclipse.platform" />
    <launcher name="eclipse">
       <linux icon="icons/icon.xpm"/>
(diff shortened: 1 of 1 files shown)


laeubi:
Would this not be needed more for EPP builds?

merks:
It would be needed by EPP and by the Installer too, i.e., by every product.

sratz:
Yes, we are on it :)

See https://github.com/eclipse-platform/eclipse.platform.ui/issues/1901#issuecomment-2382924531

sratz:
EPPs: https://github.com/eclipse-packaging/packages/pull/214
Installer: https://github.com/eclipse-oomph/oomph/pull/105

merks:
You're the best! The darned build failed with heap-space errors.  I just started it again and hope for the success...

PR #2410, opened 2024-10-01
Title Case All Over (Part III)

BeckerWdf:


BeckerWdf:
The log says:

```
13:52:11  [ERROR] Cannot resolve dependencies of project org.eclipse.jdt.feature:org.eclipse.jdt:eclipse-feature:3.20.0-SNAPSHOT
13:52:11  [ERROR]  with context {osgi.os=linux, osgi.ws=gtk, org.eclipse.update.install.features=true, org.eclipse.jdt.buildtime=true, osgi.arch=x86_64, org.eclipse.update.install.sources=true}
13:52:11  [ERROR]   Software being installed: org.eclipse.jdt.feature.group 3.20.0.qualifier
13:52:11  [ERROR]   Missing requirement: org.eclipse.jdt.feature.group 3.20.0.qualifier requires 'org.eclipse.equinox.p2.iu; org.eclipse.jdt.annotation [2.3.0,2.3.1)' but it could not be found: See log for details
```

I don't understand why. I think this is unrelated because this PR only changed HTML files.

PR #2413, opened 2024-10-01
Stop generating 2 PRs for maven updates

akurtakov:
In the past we used to have maven updates for platform-parent only as it was not working for /. Fixes allowed to have dependabot run for / but the run for eclipse-platform-parent was not removed which led to 2 PRs being generated for each update there e.g.:
* https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/pull/2412
* https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/pull/2411


akurtakov:
This change can't be verified thus pushing.

PR #2414, opened 2024-10-01
Remove Bgcolor attribute from HTML

BeckerWdf:


akurtakov:
I see you changed opening from BODY to body but not closing tag. Please keep them in sync.

BeckerWdf:
> I see you changed opening from BODY to body but not closing tag. Please keep them in sync.

That was a mass find and replace...

PR #2415, opened 2024-10-02
Update target job to use released version of Tycho and remove draft flag

laeubi:


laeubi:
Verification Jobs wont cover this.

.github/workflows/updateTarget.yml
```diff
@@ -32,8 +32,8 @@ jobs:
         cache: maven
     - name: Update Target Platform
       run: >-
-          mvn -f eclipse.platform.releng.prereqs.sdk 
-          org.eclipse.tycho.extras:tycho-version-bump-plugin:4.0.9-SNAPSHOT:update-target
+          mvn -ntp -f eclipse.platform.releng.prereqs.sdk 
+          org.eclipse.tycho.extras:tycho-version-bump-plugin:4.0.9:update-target
```
HannesWell, line 36: Wouldn't it be better to add `tycho-version-bump-plugin` to the `pluginManagement` section in the eclipse-parent pom and use the `tycho.version` used for all other Tycho plugins?

This would avoid future maintenance of that version.

laeubi: As this is a standalone action I would rather not blindly update it if there are no relevant changes.

PR #2416, opened 2024-10-02
Add workflow_dispatch trigger so we can run target update manually

laeubi:


laeubi:
Not testable by verification

laeubi:
We now have a button to run it manually if required

https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/actions/workflows/updateTarget.yml

grafik

PR #2441, opened 2024-10-08
Add debian stable based image to build swt natives

laeubi:
Currently we build the swt natives with centos-8 but this is now gone,
also centos-9 no longer provides stable releases but only
rolling-streams,

This adds a new container based on debian stable and only adds the
absolute minimum required to build the swt natives.

## Why Debian:
 Debian offers quite conservative dependencies and LTS releases
 so there is a good chance to have a common baseline that works on
 more ahead distributions as well (e.g. GClib /GTK/... versions).

### DIFF
--- a/JenkinsJobs/Builds/DockerImagesBuild.jenkinsfile
+++ b/JenkinsJobs/Builds/DockerImagesBuild.jenkinsfile
@@ -24,6 +24,7 @@ pipeline {
 				script {
 					dir('eclipse.platform.releng.aggregator/cje-production/dockerfiles') {
 						parallel(
+							'debian-swtnative': { buildAndPushImage('debian/swtnativebuild', 'eclipse/platformreleng-debian-swtnativebuild:12') },
 							'centos9-gtk4': { buildAndPushImage('centos-gtk4-mutter/9-gtk4', 'eclipse/platformreleng-centos-gtk4-mutter:9') },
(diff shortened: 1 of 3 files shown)


akurtakov:
Can you build swt gtk4 on it ? I don't think doing any switch that doesn't allow that is worth the effort.

akurtakov:
Please note that changing the image may raise in some other way (glibc/cairo/.....) and introduce breakage (and extra work - e.g. simrel projects no longer build and test with default image centos) while there is no gain from it. I don't mind the change in principle just looking for some benefit when introducing potential extra work. 
My vote is "0" on it :).

laeubi:
BTW oldstable is currently GLIBC 2.31-13

laeubi:
I now moved to debian-stable what supports gtk4 (4.8.3) and uses glibc 2.36-9 with that I can compile the GTK4 binaries when apply this PR:
- https://github.com/eclipse-platform/eclipse.platform.swt/pull/1519

akurtakov:
> I now moved to debian-stable what supports gtk4 (4.8.3) and uses glibc 2.36-9 with that I can compile the GTK4 binaries when apply this PR:
> 
>     * [Make new GTK4 Dialog dynamic loaded for pre GTK 4.10 eclipse.platform.swt#1519](https://github.com/eclipse-platform/eclipse.platform.swt/pull/1519)

FWIW, binaries produced that way will run just fine on distro having GTK 4.10+ e.g. Ubuntu 24.04 so it is a good first step.

laeubi:
Unluckily the push fails, I created:
- https://gitlab.eclipse.org/eclipsefdn/helpdesk/-/issues/5105

PR #2446, opened 2024-10-11
add ScreenshotOnFailure.onFailure(Supplier<Widget> shell)

jukzi:
Screenshots must be taken before dispose
https://github.com/eclipse-platform/eclipse.platform.swt/issues/1518


jukzi:
ignoring unrelated License vetting status check

PR #2450, opened 2024-10-12
[Version check] Make SDK number read robust against additional print-out

HannesWell:
To fix the problem discovered in P2: https://github.com/eclipse-equinox/p2/pull/558#issuecomment-2408530095.


HannesWell:
Testing https://github.com/eclipse-equinox/p2/pull/560 was successful.

Submitting this.

PR #2452, opened 2024-10-14
Don't fail on null shell supplier

iloveeclipse:
See https://github.com/eclipse-platform/eclipse.platform.swt/issues/1518


merks:
Please ignore this license problem.

HannesWell:
Thank you Andrey for the fix.
The number of test failures is now much lower again. For three out of six configs we even have zero failures.
But still Windows is causing a lot of trouble and also JDK-23 seems to have some problems, but that are other issues.

PR #2457, opened 2024-10-15
Update the target platform

merks:
- Update to Orbit m2 for batik/xmlgraphics
- Update to EMF m2.
- Update apache-el/apache-jsp to 9.0.96


merks:
After 2 builds are killed to save memory, the third finally succeeds.  I'm not sure doing 3 times the work saved anything though.

PR #2459, opened 2024-10-15
Updating Product version to 4.34

MohananRahul:


akurtakov:
Using today's I-build (prior to this change) helps are already showing 
image

I believe these files are no longer used at all. Right, @laeubi ?

PR #2477, opened 2024-10-25
[I-Build-Tests] Revert multiple assignments

HannesWell:
The job creation failed with message:
`General error during canonicalization: The sandbox does not currently support multiple assignment`.

Reverting the assignment simplifications hopefully fixes that.

Follow-up of https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/pull/2005.


HannesWell:
Submitting this immediately since the verification build does not test this at all.

PR #2485, opened 2024-10-28
Adjust bundle path to match platform ui change

elsazac:
Refs : https://github.com/eclipse-platform/eclipse.platform.ui/pull/2411

Corresponding changes in  the workbench bundle in pom.xml in releng aggregator.

Tested locally and this is the only change that is required from releng aggregator.


elsazac:
Merging this now; it’s related to PR https://github.com/eclipse-platform/eclipse.platform.ui/pull/2411

PR #2487, opened 2024-10-28
[I-Build-Test] Remove CPU-arch as criterion for testPlatform property

HannesWell:
All usages depending on the 'testPlatform' variable don't distinguish between architectures.
In case it is attempted to run the tests on a unsupported arch, checking the architecture too produces an explicit error message instead of just having a download that fails because the corresponding artifacts are not available. But those that run I-builds locally or configure a new platform are usually well aware if the executing architecture is supported or not.
And testing for the arch just makes it more complicated to run tests on new platforms.

This helps for example for https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/issues/577.

### DIFF
--- a/production/testScripts/runTests2.xml
+++ b/production/testScripts/runTests2.xml
@@ -199,11 +199,6 @@
         <equals
           arg1="${osgi.ws}"
           arg2="win32" />
-        <or>
-          <equals
-            arg1="${osgi.arch}"
-            arg2="x86_64" />
-        </or>
       </and>
     </condition>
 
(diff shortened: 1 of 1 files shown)


HannesWell:
If there are no objection's I'll submit this tomorrow (Tuesday) evening.

akurtakov:
No objection from me. Big thanks for deep diving into our most cumbersome releng parts!

HannesWell:
You're welcome, it's always nice to get a bit of configuration, even if there is still huge amount left.

Let's have this.

PR #2488, opened 2024-10-28
Unify labels for Eclipse products and SWT artifacts at eclipse-downloads

HannesWell:
For the eclipse downloads page at
https://download.eclipse.org/eclipse/downloads/
This unifies the labels for all offered products and artifacts:
- use '64-bit' instead of '64 bit version'
- always list x86_64 artifact first
- unify descriptions of artifacts for other CPU architectures

The current labels can for example be seen at
https://download.eclipse.org/eclipse/downloads/drops4/I20241028-0700/#EclipseSDK

The new labels would be
- `Windows (64-bit)`
- `Windows (64-bit for ARM)`
- `Linux (64-bit)`
- `Linux (64-bit for ARM)`
- `Linux (64-bit for Power PC)`
- `Linux (64-bit for RISC-V)`
- `Mac OSX (64-bit)`
- `Mac OSX (64-bit for ARM)`

Alternatively we could also add the architecture to for the x86_64 artifacts explicitly and label it `(64-bit for x86)`? But I still would list it first since it's the most important arch (at the moment).
What do you think?

### DIFF
--- a/eclipse.platform.releng.tychoeclipsebuilder/eclipse/publishingFiles/testManifest.xml
+++ b/eclipse.platform.releng.tychoeclipsebuilder/eclipse/publishingFiles/testManifest.xml
@@ -10,36 +10,36 @@
 
     <zipType name="Eclipse SDK">
       <platform
-        id="SWAARCH64"
-        name="Windows (ARM 64 bit version)"
-        fileName="eclipse-SDK-${BUILD_ID}-win32-aarch64.zip"></platform>
-      <platform
         id="SWX8664"
-        name="Windows (64 bit version)"
+        name="Windows (x86 64-bit)"
         fileName="eclipse-SDK-${BUILD_ID}-win32-x86_64.zip"></platform>
       <platform
+        id="SWAARCH64"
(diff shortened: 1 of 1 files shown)


laeubi:
64 bit seems quite redundant to me (unless there will arise a 128 bit architecture soon), so maybe just use `<OS> (<ARCH>)`

HannesWell:
> 64 bit seems quite redundant to me (unless there will arise a 128 bit architecture soon), so maybe just use `<OS> (<ARCH>)`

Who knows when 128bit archs will arrive 😄, but the main reason I would like to mention it is that it's clear that it's not 32-bit.

But it can be less prominent by moving it to the end. Furthermore I think `x86` should be mentioned explicitly too. So the updated list is
- `Windows (x86 64-bit)`
- `Windows (ARM 64-bit)`
- `Linux (x86 64-bit)`
- `Linux (ARM 64-bit)`
- `Linux (PowerPC 64-bit)`
- `Linux (RISC-V 64-bit)`
- `Mac OSX (x86 64-bit)`
- `Mac OSX (ARM 64-bit)`

PR #2489, opened 2024-10-28
[I-Build] Add nightly tests for win32.win32.aarch64

HannesWell:
Part of https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/issues/577

This is currently a draft since I first have to figure out in a copy of the existing test jobs which software is currently missing from the machine running the tests.

### DIFF
--- a/jobs/buildConfigurations.json
+++ b/jobs/buildConfigurations.json
@@ -74,6 +74,17 @@
 					"type": "local",
 					"path": "C:\\Program Files\\Eclipse Adoptium\\jdk-21.0.5.11-hotspot"
 				}
+			},
+			{
+				"os": "win32",
+				"ws": "win32",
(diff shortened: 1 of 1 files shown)


HannesWell:
> Should this one be closed now? Or do you think that infra can really survive one more test?

I think this is still relevant. As the win32.aarch64 is currently not used for tests at all, only to compile the SWT and Equinox binaries, I think it won't be overloaded. Furthermore we currently have one I-build test for Windows and I think it would be important to have two, like for the other OS to identify potential infrastructure issues easier.

The problem, why this is not yet pushed is that the current Jenkins setup on the win32.aarch64 machine seems to lead to too long paths (presumably) and I have not yet found out where exactly this is not handled right.
But I still have the intention to complete this.

PR #2490, opened 2024-10-28
[Build] remove windows2012tests authentication token from Jenkins jobs

HannesWell:
@fredg02 or @heurtematte or anybody else, can you tell if `windows2012tests` authentication token is still useful?
I have the impression it is not.

### DIFF
--- a/JenkinsJobs/AutomatedTests/I_unit_mac.groovy
+++ b/JenkinsJobs/AutomatedTests/I_unit_mac.groovy
@@ -16,8 +16,6 @@ for (ARCH in ARCHS){
       stringParam('buildId', null, 'Build Id to test (such as I20240611-1800, N20120716-0800).')
     }
 
-    authenticationToken('windows2012tests')
- 
     definition {
       cps {
         sandbox()
--- a/JenkinsJobs/AutomatedTests/I_unit_win32.groovy
+++ b/JenkinsJobs/AutomatedTests/I_unit_win32.groovy
@@ -11,8 +11,6 @@ for (STREAM in STREAMS){
       stringParam('buildId', null, 'Build Id to test (such as I20240611-1800, N20120716-0800).')
     }
 
-    authenticationToken('windows2012tests')
- 
     definition {
       cps {
         sandbox()
--- a/JenkinsJobs/PerformanceTests/ILR_perf_lin64.groovy
+++ b/JenkinsJobs/PerformanceTests/ILR_perf_lin64.groovy
@@ -49,8 +49,6 @@ jdtuirefactoring
 
     label('performance')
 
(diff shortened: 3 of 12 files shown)


fredg02:
I have no insight if the token is still used to access the build scripts remotely. If [startTests.sh](https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/blob/master/production/testScripts/startTests.sh) is not used anymore, it would be a strong indication, but not necessarily ultimate proof. :slightly_smiling_face:

HannesWell:
> I have no insight if the token is still used to access the build scripts remotely. 

I was hoping that you can maybe tell, if that token is even usable or e.g. expired?

> If [startTests.sh](https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/blob/master/production/testScripts/startTests.sh) is not used anymore, it would be a strong indication, but not necessarily ultimate proof. 🙂

Good point: I don't think so, see https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/pull/2498
So maybe my question from above does not need an answer :)
But if this works, you could probably remove that token.

fredg02:
> I was hoping that you can maybe tell, if that token is even usable or e.g. expired?

There are different kinds of tokens in Jenkins. In this case, the authentication token is not the name of a Jenkins credential that contains the actual token... **it is the actual token**. So `windows2012tests` is the literal token that can be used to trigger this job remotely (e.g. from the `startTest.sh` shell script).

You can see the corresponding field in the Jenkins UI in the screenshot here: https://stackoverflow.com/questions/42817169/jenkins-trigger-builds-remotely-authentication-token-option-missing

HannesWell:
> In this case, the authentication token is not the name of a Jenkins credential that contains the actual token... it is the actual token.

Oh, ok. That's a surprise. Thanks for that clarification.
I'll check if that still works, the link you posted indicates that other options might still mandate additional authentication, rendering the token unusable.
But if that works, I think we should really remove this and should not allow the world to trigger builds on our Jenkins. That should only be possible for committers using the web UI or their personal credentials/tokens.

HannesWell:
I have now made some attempts to launch builds in the releng Jenkins just using the token and it was indeed possible to launch a build by invoking the corresponding URL with that token specified. I didn't manage to pass a valid parameter value (most of the affected builds require at least one parameter) which had the consequence the build failed quickly. Still I was able to launch it (and could probably fix the parameter issue), even if not logged-in to the `releng` Jenkins in the browser.
I think that should not be possible. Only authenticated committers should be able to launch builds explicitly.

At the same time I don't see HUDSON_TOKEN being referenced at any place in this repo or it's sub-module and therefore assume it's unused.

Therefore I submit this now.

PR #2491, opened 2024-10-28
[I-Build-Tests] Clean-up and unify test launch scripts

HannesWell:
- Use curl instead of wget (the former seems to be available on more machines by default, but I'll very this).
- Remove unnecessary whitespace in print-out and just use
- Remove unnecessary print-outs and unnecessary variable references

### DIFF
--- a/JenkinsJobs/AutomatedTests/I_unit_linux.groovy
+++ b/JenkinsJobs/AutomatedTests/I_unit_linux.groovy
@@ -88,6 +88,7 @@ spec:
               JAVA_HOME = tool(type:'jdk', name:'openjdk-jdk''' + JAVA_VERSION + '''-latest')
               ANT_HOME = tool(type:'ant', name:'apache-ant-latest')
               PATH = "${JAVA_HOME}/bin:${ANT_HOME}/bin:${PATH}"
+              ANT_OPTS = "-Djava.io.tmpdir=${WORKSPACE}/tmp -Djava.security.manager=allow"
           }
           steps {
               container ('custom'){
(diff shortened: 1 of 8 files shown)


JenkinsJobs/AutomatedTests/I_unit_linux.groovy
```diff
@@ -131,16 +92,14 @@ spec:
...
                         
                         RAW_DATE_END="$(date +%s )"
-                        
-                        echo -e "\\n\\tRAW Date End: ${RAW_DATE_END} \\n"
+                        echo "RAW Date End: ${RAW_DATE_END}"
                         
                         TOTAL_TIME=$((${RAW_DATE_END} - ${RAW_DATE_START}))
-                        
-                        echo -e "\\n\\tTotal elapsed time: ${TOTAL_TIME} \\n"
+                        echo "Total elapsed time: ${TOTAL_TIME}"
```
HannesWell, line 102: Actually the manual time measurement could just be removed entirely.
Jenkins measures the runtime of each stage and can also print timestamps per line.
So only if one inspects the raw logs the total runtime is not available, but I assume that's a relevant use-case.

akurtakov: IMO remove these measurements. This files are useless outside of Jenkins so duplicating what Jenkins already does makes them only more complicated.

HannesWell:
With the follow-up fix https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/pull/2508 applied, all I-build tests succeeded in the sense that there are no failures related to this set-up change.

Issue #2494, opened 2024-10-29
Linux test results summary broken


iloveeclipse:
See https://download.eclipse.org/eclipse/downloads/drops4/I20241028-1800/

Probably caused by renaming configs in https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/pull/2481

All three test rows for Linux show same result for the Linux ep434I-unit-linux-x86 64-java23_linux.gtk.x86_64 23 build:

![image](https://github.com/user-attachments/assets/fe8989d9-ef67-4dc0-9503-c91a0587d87a)

In reality all three rows can't be same as we have different count of compiler tests per build.

@HannesWell : please check.

HannesWell:
Verified with https://download.eclipse.org/eclipse/downloads/drops4/I20241029-1420/ that this is really fixed.
And the labels now look good again as well.

Thanks for the report.

PR #2497, opened 2024-10-29
[I-Build] Fix test result summary configuration label break-up

HannesWell:
The underscore in 'x86_64' is still part of the first part, the 'job-name' part, of the configuration label.
Therefore the only the first underscore after x86_64 should be considered as delimiter.

Additionally improve dynamic adaption of listing of log files.

Fixes https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/issues/2494 (at least it should).

This will also fix the wrong display of label line breaks. The wrong break-up is also the cause for the wrong numbers because, because simplify speaking, the test-result folder name must start with the 'first line' of the label to be considered a match.
Therefore the jobnames for the linux test-configs are all the same and the first one in the file-system is used for all Linux configs:
grafik

### DIFF
--- a/eclipse.platform.releng.tychoeclipsebuilder/eclipse/publishingFiles/staticDropFiles/utilityFunctions.php
+++ b/eclipse.platform.releng.tychoeclipsebuilder/eclipse/publishingFiles/staticDropFiles/utilityFunctions.php
@@ -276,7 +276,7 @@ be split as
  */
 function computeDisplayConfig($config) {
   $lastUnderscore = strrpos ($config, "_");
-  $firstUnderscore = strpos ($config, "_"); 
+  $firstUnderscore = strpos ($config, "_", strpos($config, "x86_64") + 6);
   $platformLength=$lastUnderscore - $firstUnderscore - 1;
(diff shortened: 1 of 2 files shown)


HannesWell:
The verification build does not test this, so this can be submitted immediately.

Once the corresponding change in [eclipse.platform.releng.buildtools](https://github.com/eclipse-platform/eclipse.platform.releng.buildtools/pull/70) is deployed, I'll start another I-build to verify this change.

HannesWell:
I-build started: https://ci.eclipse.org/releng/job/Builds/job/I-build-4.34/106/

eclipse.platform.releng.tychoeclipsebuilder/eclipse/publishingFiles/staticDropFiles/utilityFunctions.php
```diff
@@ -276,7 +276,7 @@ function fileSizeForDisplay($filename) {
  */
 function computeDisplayConfig($config) {
   $lastUnderscore = strrpos ($config, "_");
-  $firstUnderscore = strpos ($config, "_"); 
+  $firstUnderscore = strpos ($config, "_", strpos($config, "x86_64") + 6);
```
HannesWell, line 279: In general it would probably be better to use a more robust delimiter, e.g. two dashes.
But I couldn't find out where the config strings are assembled and all the locations that expect the current delimiter.

PR #2502, opened 2024-10-31
Update icu4j to 76.1

merks:


merks:
@laeubi 

I'm not sure how to configure the automatic TP updater that major versions of icu4j should be suggested...

laeubi:
We use a maven-versions rule file here:

https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/blob/master/eclipse.platform.releng.prereqs.sdk/update-rules.xml

But major versions are explicitly disabled here:

https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/blob/19abe6dc4cbe9580c12b0a5d7dd68a8b83e4cb46/.github/workflows/updateTarget.yml#L38

so if one wants updates for *some* only we need to globally enable it and add specific rules for others I think that could become a bit cumbersome.

Alternatively one might enhance Tycho to support some kind of delegation model e.g.

`-Dcom.ibm.icu.allowMajorUpdate=true`

would allow major updates for everything under the `com.ibm.icu` group.

`-Dcom.ibm.icu.icu4j.allowMajorUpdate=true`

would allow major updates for everything under the `com.ibm.icu` group with artifact id `icu4j` but this can maybe become brittle if we have many of those exceptions.

PR #2504, opened 2024-11-02
Use environment variables in workflow for automated version increments

HannesWell:
to harden workflow against poisoned pipeline executions (PPE).


HannesWell:
Submitted this immediately since the verification build does not test this.
All callers will pick this up automatically in their next run.

PR #2507, opened 2024-11-02
[I-Build-Tests] Simplify archive name computation and don't guess os, ws and arch

HannesWell:
1. Unify archive name computation and make it less static
2. Don't guess os-ws-arch configuration, with the existing infrastructure os, ws and arch properties are explicitly specified anyways.

Additionally sanitize white-space.

This also helps for https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/issues/577


HannesWell:
The verification build does not test this.

PR #2508, opened 2024-11-03
[I-Build-Test] Fix call of ant-diagnostics on Windows machines

HannesWell:
Calling 'ant -diagnostics' on Windows somehow exists the bat script immediately. Call it in a sub-cmd-shell to workaround this problem until a completely correct solution is found.


HannesWell:
The verification build does not test this, but I have already verified this with a replay of the latest failed execution of the I-build tests for Windows.
Submitting this and regenerate the jobs.

HannesWell:
With this fix applied, the windows I build tests 'succeeded' again.

Btw. if somebody knows how to launch ant to not exit the bat script I'd be happy to learn it.

PR #2510, opened 2024-11-04
How to use junit.jupiter Extensions in OSGi and plain java

jukzi:


eclipse.platform.releng/bundles/org.eclipse.test/src/org/eclipse/test/services/LoggingTestExtension.java
```diff
@@ -0,0 +1,43 @@
+package org.eclipse.test.services;
```
akurtakov, line 1: Missing license header

jukzi:
@Bananeweizen is this snippet basically what we discussed? Or am i missing something?

Bananeweizen:
@jukzi The mechanics look good to me. I haven't used the properties file yet (because we have all properties in the maven invocation for our company projects), therefore I can't judge that file. Also if you just want the timing as such, https://github.com/junit-team/junit5/blob/main/documentation/src/test/java/example/timing/TimingExtension.java would be available.

You also had mentioned a problem with TestSuites, e.g. forgetting to add new tests to existing suites, or re-running suites taking long time? Can you eventually point me to some existing discussion or bug about that? I run all the tests in our company without using any suites, therefore I can't really remember anymore which problem that was...

eclipse.platform.releng/bundles/org.eclipse.test/src/org/eclipse/test/services/LoggingTestExtension.java
```diff
@@ -0,0 +1,56 @@
...
+		getStore(context).put("TIME", n0);
+		System.out.println(BEFORE_TEST_START + context.getDisplayName());
+	}
+
+	@Override
+	public void afterTestExecution(ExtensionContext context) throws Exception {
+		long t1 = System.nanoTime();
+		long t0 = getStore(context).remove("TIME", long.class);
+		String took = " after: " + (t1 - t0) / 1_000_000 + "ms";
+		Throwable t = context.getExecutionException().orElse(null);
```
Bananeweizen, line 43: that's the most simple way to judge the test outcome. alternatively you could also implement https://junit.org/junit5/docs/5.4.1/api/org/junit/jupiter/api/extension/TestWatcher.html, which gets notified on test results. Not sure what fits better for your case.

jukzi:
@Bananeweizen the main point is that somehow "junit.jupiter.extensions.autodetection.enabled=true" has to be set (you did not mention that) otherwise it does not work.

regarding TestSuite see:
https://github.com/eclipse-platform/eclipse.platform.ui/pull/2391 and
https://www.eclipse.org/lists/jdt-dev/threads.html#02401

eclipse.platform.releng/bundles/org.eclipse.test/src/org/eclipse/test/services/LoggingTestExtension.java
jukzi: thanks, fixed

eclipse.platform.releng/bundles/org.eclipse.test/META-INF/MANIFEST.MF
```diff
@@ -11,7 +11,9 @@ Require-Bundle: org.apache.ant,
  org.eclipse.core.runtime,
  org.eclipse.ui.ide.application,
  org.eclipse.equinox.app,
- org.junit
+ org.apache.aries.spifly.dynamic.bundle,
```
HannesWell, line 14: You should not the bundle `org.apache.aries.spifly.dynamic.bundle` by it's name.

If you want to make sure that this bundle properly registers it's provided service at the OSGi service loader mediator is present then you should instead register the extension as provided Java service:
```
Require-Capability: osgi.extender;filter:="(&(osgi.extender=osgi.serviceloader.registrar)
 (version>=1.0.0)(!(version>=2.0.0)))"
Provide-Capability: osgi.serviceloader;osgi.serviceloader=org.junit.jupiter.api.extension.Extension
```
Maybe this also helps with your registration issue, because without these changes the extension is basically invisible in an OSGi runtime. Of course this requires also that the JUnit bundles are configured correspondingly for OSGi runtimes as service consumer.
For reference see
https://aries.apache.org/documentation/modules/spi-fly.html#specconf
https://docs.osgi.org/specification/osgi.cmpn/8.0.0/service.loader.html

HannesWell (changes requested):

eclipse.platform.releng/bundles/org.eclipse.test/META-INF/MANIFEST.MF
HannesWell: The issue you mentioned in 
> the main point is that somehow "junit.jupiter.extensions.autodetection.enabled=true" has to be set (you did not mention that) otherwise it does not work.

> I would like to understand what is changed/improved if i would use the "Capability" thing instead. Especially how that would work in an non-osgi environment like SWT tests.

To summarize it in short: It makes the Java service loader work in OSGi runtimes, where it does not work out of the box due to classloader separation of bundles. Additionally you have to make sure that a provider is auto-started by the app or product.
In non-OSGi environments like SWT tests the java Service-Loader works out of the box because the app has one flat classpath.

In general I'm not aware of any situation where a bundle should require `org.apache.aries.spifly.dynamic.bundle` or one of it's packages directly.

If you point me to your exact use-case I can try to help making it work.

jukzi: The general idea is to surround all junit tests in the I-Build with the same logging and checks. However as far as i have tested it seems to work on junit5 tests only and still most tests are 3 or 4. Thats where i am stuck with this experiment.

laeubi: > In general I'm not aware of any situation where a bundle should require org.apache.aries.spifly.dynamic.bundle or one of it's packages directly.

I think we more want a require capability then here this will require/pull in that automatically.

akurtakov:
Does anyone plan to finish this one or we should close it?

PR #2515, opened 2024-11-05
[I-Build-Tests] Run Linux tests on default 'centos-8' agent

HannesWell:
Save the complexity of a customized kubernetes pod configuration which doesn't seem to be necessary.

@akurtakov or @sravanlakkimsetti do you know any reason that requires a special image?

I'm currently experimenting with this change by replaying the latest I-build test execution for Linux-java17 at
https://ci.eclipse.org/releng/job/AutomatedTests/job/ep434I-unit-linux-x86_64-java17/18/

This should confirm if this a save change.
At least when running only the `UI` suite all tests passed:
https://ci.eclipse.org/releng/job/AutomatedTests/job/ep434I-unit-linux-x86_64-java17/17/

### DIFF
--- a/JenkinsJobs/AutomatedTests/I_unit_linux.groovy
+++ b/JenkinsJobs/AutomatedTests/I_unit_linux.groovy
@@ -24,61 +24,7 @@ pipeline {
     buildDiscarder(logRotator(numToKeepStr:'5'))
   }
   agent {
-    kubernetes {
-      label 'centos-unitpod''' + JAVA_VERSION + ''''
-      defaultContainer 'custom'
-      yaml """
-apiVersion: v1
-kind: Pod
-spec:
-  containers:
-  - name: "jnlp"
-    resources:
-      limits:
-        memory: "2048Mi"
-        cpu: "2000m"
-      requests:
-        memory: "512Mi"
-        cpu: "1000m"
-  - name: "custom"
-    image: "eclipse/platformreleng-centos-gtk3-metacity:8"
-    imagePullPolicy: "Always"
-    resources:
-      limits:
-        memory: "4096Mi"
-        cpu: "1000m"
-      requests:
-        # memory needs to be at least 1024Mi, see https://gitlab.eclipse.org/eclipsefdn/helpdesk/-/issues/2478
-        memory: "1024Mi"
-        cpu: "1000m"
-    securityContext:
-      privileged: false
-    tty: true
-    command:
-    - cat
-    volumeMounts:
-    - mountPath: "/opt/tools"
-      name: "volume-0"
-      readOnly: false
-    workingDir: "/home/jenkins/agent"
-  nodeSelector: {}
-  restartPolicy: "Never"
-  volumes:
-  - name: "volume-0"
-    persistentVolumeClaim:
-      claimName: "tools-claim-jiro-releng"
-      readOnly: true
-  - configMap:
-      name: "known-hosts"
-    name: "volume-1"
-  - emptyDir:
-      medium: ""
-    name: "workspace-volume"
-  - emptyDir:
-      medium: ""
-    name: "volume-3"
-"""
-    }
+    label 'centos-8'
   }
 
   stages {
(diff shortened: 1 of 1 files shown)


HannesWell:
If this change is applied, as far as I can tell, then only the Y-builds seem to be using the `platformreleng-centos-gtk3-metacity` docker-image:
https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/blob/f3dfcfd7697d7f93494051c1a069543ad17207a1/JenkinsJobs/YBuilds/Y_unit_cen64_gtk3_java17.groovy#L47

But I assume if we can use the default `centos-8` image here, we can do the same for Y-builds and the `platformreleng-centos-gtk3-metacity` image can be put to rest.

Of course, instead of `centos-8`, we could also use the new Ubunut based images to be on par with the verification builds?
But I would change that in a follow-up PR.

akurtakov:
I don't remember exactly but it's even possible that it's that way from the time when default image was centos 7 based. Please go ahead and merge.

JenkinsJobs/AutomatedTests/I_unit_linux.groovy
```diff
@@ -24,61 +24,7 @@ pipeline {
...
-    name: "volume-1"
-  - emptyDir:
-      medium: ""
-    name: "workspace-volume"
-  - emptyDir:
-      medium: ""
-    name: "volume-3"
-"""
-    }
+    label 'centos-8'
```
laeubi, line 27:
```suggestion
    label 'ubuntu-latest'
```

akurtakov: Hannes already said he plans to do so in a follow up PR.

laeubi: > Hannes already said he plans to do so in a follow up PR.

Sure but why not do it right now? I think there is a good proof it already works with centos-8 so why not do the same with ubuntu and save one immediate step? e.g if `centos-8` do **not** work there is no way for us to make it work so I don't see any benefits here.

akurtakov: So far what I have found (and is hopefully fixed/changed ) is that Ubuntu used C locale (rather than some UTF8) and dash(rather than bash) and who knows what other difference is lingering around. It is easier for me to spot things if/when they are done incrementally but IMO it's up to Hannes to decide whether to do it in multiple steps or not as in case of issue he will the first one to look into it.
One can not be sure that centos-8 and platformreleng-centos-gtk3-metacity have the same package sets installed and in case of difference comparing package names (with their hierarchies) between distributions is an adventure I'm not eager to do at once.

HannesWell: > > Hannes already said he plans to do so in a follow up PR.
> 
> Sure but why not do it right now? I think there is a good proof it already works with centos-8 so why not do the same with ubuntu and save one immediate step?

Because this step is in the broader sense just a 'simplification of the configuration' and not a change of the underlying Linux distribution. I know in detail it's still another image, but I think the difference is still smaller than when changing the distro. And the configuration change is not only about the image, as you can see the kubernetes config defines volumes, multiple containers and various settings. All that hopefully does not break the build but I would like to have that confirmed first.
But since https://ci.eclipse.org/releng/job/AutomatedTests/job/ep434I-unit-linux-x86_64-java17/18/ concluded with only three test-failrues I'm confident that it will also work well for all other Linux/Java configs.

I'm going to submit this now and will check the result of tonight's I-build and provide a follow-up PR to run I-build tests on ubuntu.

JenkinsJobs/AutomatedTests/I_unit_linux.groovy
```diff
@@ -24,61 +24,7 @@ pipeline {
...
-      yaml """
-apiVersion: v1
-kind: Pod
-spec:
-  containers:
-  - name: "jnlp"
-    resources:
-      limits:
-        memory: "2048Mi"
-        cpu: "2000m"
```
HannesWell, line 39: Now that this is applied, the I-build tests for Linux seem to complete in about 2.5 to 3 hours instead of 4h.
I assume that's because the default image has higher resource limits:
```
      limits:
        memory: "4096Mi"
        cpu: "2000m"
```

PR #2516, opened 2024-11-05
[I-Build-Tests] Inline 'testAll' scripts

HannesWell:
This is currently a draft since I have to test this locally first.

As a follow-up we could consider to inline the `runtests.sh/bat` scripts as well or at least to simplify them to what's still relevant nowadays.
Keeping might be interesting for those that run I-builds locally, but on the long run I would like to simplify the set up to an extend so that we can provide a simple launch config to run the I-build locally for the current platform and with the current config in this repo. The compiled archives would then be either provided locally (if one wants to test changes) or be fetched from a matching I-build. But that's another story.


HannesWell:
I'm now quite confident that this works as expected.
Submitting and will trigger an I-build to verify it.

HannesWell:
Looks like the creation of the `${testedPlatform}_consolelog.txt` log file does not happen, which becomes visible in the `Console Output Logs` section of the `Test Results` overview for an I-build and that the `Summary of Unit Tests Results` and the test status icons show that tests would not yet have completed.

I'm looking into that.

Issue #1912, opened 2024-03-16
The https://download.eclipse.org/eclipse/updates/4.32-I-builds don't have a content.jar/content.xml.xz


merks:
All the builds https://download.eclipse.org/eclipse/updates/4.32-I-builds have only a content.xml

![image](https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/assets/208716/1547cda7-a3a7-46e9-88bc-9f7510e7c157)

The same is also true for the release build:

https://download.eclipse.org/eclipse/updates/4.31/R-4.31-202402290520/

This is one of the most heavily uses sites so that's pretty nasty for the server.

This is the size the files should be:

![image](https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/assets/208716/7c2b070b-a5f1-4a77-ae71-4a44b08dd2ee)

Perhaps this is lingering effect from this ongoing manual processing of the file:

https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/issues/509

MohananRahul:
For R-4.31-202402290520, during the promotion, it appears that the generation of content.jar and content.xml.xz did not take place as anticipated. let me compress the content.

merks:
@MohananRahul 

I suspect that if we don't fix this in the I-builds then this problem will happen again when we publish the release...

merks:
@MohananRahul  @akurtakov 

This continues to be a problem for 4.33-I-builds and like my previous comment, I'm concerned that we better be sure this isn't also a problem for the 4.33 release repository that we will soon publish.

MohananRahul:
@akurtakov  ,removal of [delete](https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/blob/master/eclipse.platform.releng.tychoeclipsebuilder/eclipse.platform.repository/pom.xml#L71)  should suffice I think for IBuilds. 

for 4.33 release , It will be take care during promotion.

akurtakov:
Please do it.

akurtakov:
Should this one be closed?

merks:
It's not fixed yet:

![image](https://github.com/user-attachments/assets/d1296d42-09f9-42f4-a3eb-95de16374e6b)

MohananRahul:
> @MohananRahul what is your plan here?

I'll check it once I'm back. Currently, I'm traveling.

HannesWell:
[...]

PR #2518, opened 2024-11-06
Removing unjar of content.jar

SougandhS:
This fix will remove the extracing of content.jar file and retain existing content.jar and content.xml.xz

Fix https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/issues/1912

Output [local]- 
image

### DIFF
--- a/eclipse.platform.releng.tychoeclipsebuilder/eclipse.platform.repository/pom.xml
+++ b/eclipse.platform.releng.tychoeclipsebuilder/eclipse.platform.repository/pom.xml
@@ -58,27 +58,6 @@
         </configuration>
       </plugin>
       <plugin>
-        <artifactId>maven-antrun-plugin</artifactId>
-        <executions>
-          <execution>
-            <id>unpack-content-jar</id>
-            <phase>package</phase>
-            <configuration>
-              <target>
-                <unjar
-                  src="target/repository/content.jar"
-                  dest="target/repository" />
-                <delete file="target/repository/content.jar" />
-                <delete file="target/repository/content.xml.xz" />
-              </target>
-            </configuration>
-            <goals>
-              <goal>run</goal>
-            </goals>
-          </execution>
-        </executions>
-      </plugin>
-      <plugin>
         <groupId>org.codehaus.mojo</groupId>
         <artifactId>xml-maven-plugin</artifactId>
         <executions>


akurtakov:
I have no memories about this and fully defer to you guys to figure it out.

mickaelistria:
IIRC, the content.xml is tweaked in some further steps, and since p2 preferes xml.xz or jar, if those files are still present and only the content.xml is tweaked, then the tweaks will not be found by consumers (as .xml.xz without trick will be processed instead)

SougandhS:
> IIRC, the content.xml is tweaked in some further steps, and since p2 preferes xml.xz or jar, if those files are still present and only the content.xml is tweaked, then the tweaks will not be found by consumers (as .xml.xz without trick will be processed instead)

do you mean that still we need to retain content.xml with jar and xz ?

merks:
I'm traveling so have no time for reviews.  I assume that nothing (no script) is altering the contents of content.jar after it is produced by p2.    A "normal" simple p2 repository has just these files:

https://download.eclipse.org/releases/2024-09/202409111000/

image

mickaelistria:
The very next mojo execution execution in pom.xml with id `generate-extra-iu` tweaks the content.xml file after it's produced. As long as we use it, we need to clear the .jar and .xml.xz (and optionally recreate them later from the altered content.xml).
There might be better ways to do that, such as using p2 instructions in a p2.inf file. If we can leverage such more standard approach, then we may be more easily able to get rid of some executions.

SougandhS:
> The very next mojo execution execution in pom.xml with id `generate-extra-iu` tweaks the content.xml file after it's produced. As long as we use it, we need to clear the .jar and .xml.xz (and optionally recreate them later from the altered content.xml). There might be better ways to do that, such as using p2 instructions in a p2.inf file. If we can leverage such more standard approach, then we may be more easily able to get rid of some executions.

Shall we then try to extract the content.jar to another target ? and retain .jar and .xz

akurtakov:
IMO you should investigate why the jar has to be extracted and it's content manipulated after that and remove the need for that.

SougandhS:
> IMO you should investigate why the jar has to be extracted and it's content manipulated after that and remove the need for that.

Sure 👍

akurtakov:
I'm closing this one as the real change will have to be quite different.

Issue #1960, opened 2024-04-04
Class file diff in validation builds


laeubi:
The PDE validation build currently reports a classfile difference here:

https://ci.eclipse.org/pde/job/eclipse.pde/job/PR-1217/2/maven-warnings/new/source.b3bd3a11-fb9e-418b-b444-b56138864442/#11109
https://ci.eclipse.org/pde/job/eclipse.pde/job/PR-1217/2/eclipse/new/source.72006889-9cf8-483f-9cc9-ec32c906cc12/#0

Compile logs and artifact comparison can be found here:

https://ci.eclipse.org/pde/job/eclipse.pde/job/PR-1217/2/artifact/e4tools/bundles/org.eclipse.e4.tools.emf.ui/target/

The PR itself does not change that file at all.

akurtakov:
Do you think updating to latest ecj in parent is what's needed ?

laeubi:
Probably yes at laest the Ibuild seem to not report comparator diffs.

PR #2538, opened 2024-11-13
[Build] Use last successful ECJ snapshot in verification builds

HannesWell:
Use the latest ECJ snapshot version from the last successful I-build (without comparator errors) for verification builds, to automatically align with I-builds.
Currently ECJ has to be deployed manually to a JDT snapshot-repository after changes in ECJ that caused comparator-errors:
- https://github.com/eclipse-pde/eclipse.pde/pull/1461#issuecomment-2466780430

This prevents issues like
- https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/issues/1960
- https://github.com/eclipse-pde/eclipse.pde/pull/1467

If there are no other usages of `https://repo.eclipse.org/content/repositories/eclipse-staging`, this should make the JDT Jenkins job to deploy ECJ to it obsolete
- https://ci.eclipse.org/jdt/job/copyAndDeployJDTCompiler/

as well as the target repository and should allow us to delete:
https://repo.eclipse.org/content/repositories/eclipse-staging

@jarthana and @akurtakov this is what we talked about in today's PMC call.
@MohananRahul can you please check if the release process documentation update is correct and fine?

### DIFF
--- a/JenkinsJobs/Builds/I_build.groovy
+++ b/JenkinsJobs/Builds/I_build.groovy
@@ -428,6 +428,11 @@ spec:
           }
 		}
 		stage('Trigger publication to Maven snapshots repo') {
+			when {
+				environment name: 'COMPARATOR_ERRORS_SUBJECT', value: ''
+				// On comparator-erros, skip the deployment of snapshot version to the 'eclipse-snapshots' maven repository to prevent that ECJ snapshot
+				// from being used in verification builds. Similar to how the p2-repository is not added to the I-build composite in that case.
+			}
 			steps {
               container('jnlp') {
 				build job: 'CBIaggregator', parameters: [string(name: 'snapshotOrRelease', value: '-snapshot')], wait: false
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -58,16 +58,6 @@
          - Use the mail template from the promotion build [artifacts](https://ci.eclipse.org/releng/job/eclipse.releng.renameAndPromote/lastSuccessfulBuild/artifact/) in Jenkins to get the download urls.
          - Make sure to mention that the Master branch is now again open for development.
        * For **Milestone builds** return the I-builds to the normal schedule.
-     * **Update ECJ compiler** in the platform build (if it needs to be updated).
-       * To find the new compiler version:
-         - Go to the update site for the release candidate
-         - Click `plugins`
-         - Find `org.eclipse.jdt.core.complier.batch_${ecjversion}.jar`
-       * Edit the [copyAndDeployJDTCompiler](https://ci.eclipse.org/jdt/job/copyAndDeployJDTCompiler) job in Jenkins
-         - Only JDT committers can run the job, but Releng/Platform committers can configure it
-         - Update the default values of the `versionfolder`, `buildid` and `ecjversion` parameters.
-         - Update the build triggers to schedule a build for the current date.
-       * Finally update the `cbi-ecj-version` in [eclipse.platform.releng.aggregator/eclipse-platform-parent/pom.xml](https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/blob/master/eclipse-platform-parent/pom.xml)
      * **After RC1**
        * Leave the I-builds running on the milestone schedule for RC2. 
        * Comment on EMF, ECF and Orbit issues to ask for final release builds.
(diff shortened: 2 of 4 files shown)


MohananRahul:
> @MohananRahul can you please check if the release process documentation update is correct and fine?

Ok , looks good.

akurtakov (commented):
Looks good to me but let's land it after 4.34 is branched to prevent possible last minute surprise.

HannesWell:
> Looks good to me but let's land it after 4.34 is branched to prevent possible last minute surprise.

Acknowledge.

I have now changed this PR to use a version range for `cbi-ecj-version`:
`<cbi-ecj-version>[3.40.0,4)</cbi-ecj-version>`

This way the version does not have to be updated at all as long as ECJ stays at version 4.x.
If we use this, we can simplify the release procedure and just drop the update step.
WDYT?

laeubi:
Maven version ranges can give surprising results, e.g. a version of `4.0.0-alpha1` is _included_ in your range.
It is also bad in terms of reproducibility as older branches will get suddenly newer releases.

HannesWell:
> Maven version ranges can give surprising results, e.g. a version of `4.0.0-alpha1` is _included_ in your range. 

In general for the master branch I assume that we want the latest version anyways, so the upper bound could be empty/infinity.

> It is also bad in terms of reproducibility as older branches will get suddenly newer releases.

OK, for older branches this is indeed probably not wanted. I reverted that part.
Overall this is still a significant simplification. 



> If there are no other usages of `https://repo.eclipse.org/content/repositories/eclipse-staging`, this should make the JDT Jenkins job to deploy ECJ to it obsolete
> 
>     * https://ci.eclipse.org/jdt/job/copyAndDeployJDTCompiler/
> 
> 
> as well as the target repository and should allow us to delete: https://repo.eclipse.org/content/repositories/eclipse-staging

@jarthana or @akurtakov can you tell if there are other users of that repository? If not I suggest we delete that repository and see if there are any unknown users. And if nobody complains or someone cannot be moved to the maven-repo into that the I-builds deploy, we can continue to delete the Jenkins job too.

akurtakov:
In the light of https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/issues/2595 what prevents this one from merging? I'm not troubled by 4.0.0-alpha at all. There is no way to know whether this staging repo is used by anyone else but being "staging" means I have no concerns about changes in it.

HannesWell:
> In the light of #2595 what prevents this one from merging? I'm not troubled by 4.0.0-alpha at all.

Nothing specific, I just had to update the version to the current version of `org.eclipse.jdt.core.compiler.batch`. It's currently 3.40.100 so I assume that it will change to 3.41 at some point in this dev-cycle since @jarthana mentioned it's usually always bump by one minor version?
So yes, using a version range would avoid that. But on the other hand, if we build fixes on maintenance branches we would always use the latest ECJ version and not the one used at the time the branch was forked.
But maybe it would be better anyways to make it part of the release procedure to set the `cbi-ecj-version` property to the fixed version of the about to be released ECJ, when creating the maintanance branche?
Then we could use an open range for the compiler in active development and always use a matching version for maintenance.

> There is no way to know whether this staging repo is used by anyone else but being "staging" means I have no concerns about changes in it.

My suggestion is to remove that repo immediately, after this is submitted, to find out if there are other users (by breaking them and hope thy will complain).

MohananRahul:
> My suggestion is to remove that repo immediately, after this is submitted, to find out if there are other users (by breaking them and hope thy will complain).  
Removal this repo will break old maintenance branch build.

MohananRahul:
> > In the light of #2595 what prevents this one from merging? I'm not troubled by 4.0.0-alpha at all.
> 
> Nothing specific, I just had to update the version to the current version of `org.eclipse.jdt.core.compiler.batch`. It's currently 3.40.100 so I assume that it will change to 3.41 at some point in this dev-cycle since @jarthana mentioned it's usually always bump by one minor version? So yes, using a version range would avoid that. But on the other hand, if we build fixes on maintenance branches we would always use the latest ECJ version and not the one used at the time the branch was forked. But maybe it would be better anyways to make it part of the release procedure to set the `cbi-ecj-version` property to the fixed version of the about to be released ECJ, when creating the maintanance branche? Then we could use an open range for the compiler in active development and always use a matching version for maintenance.
> 
> > There is no way to know whether this staging repo is used by anyone else but being "staging" means I have no concerns about changes in it.
> 
> My suggestion is to remove that repo immediately, after this is submitted, to find out if there are other users (by breaking them and hope thy will complain).

Removal of repo will cause failures in maintenance branch builds right?

HannesWell:
> > My suggestion is to remove that repo immediately, after this is submitted, to find out if there are other users (by breaking them and hope thy will complain).
> 
> Removal of repo will cause failures in maintenance branch builds right?

I checked the version used there and it would indeed break them, because they use a fully-qualified version and at Maven-Central only unqualified versions are used. I adjusted the release procedure to consider that for future maintenance branches.
But for existing ones we either have to keep the repo or update the branches (if needed).
So I guess it's simpler to just keep the repo for some time. But this also means that this, respectivly the adjustments in release procedure will make future maintenance branches more robust since they rely on one staging repo less.

But since there are no objections against this change, I'll submit this now in order to make the latest ECJ also available for verification builds.

HannesWell:
> But since there are no objections against this change, I'll submit this now in order to make the latest ECJ also available for verification builds.

Ok, the Jenkins build of ECJ itself is objecting. Will look into that tomorrow.

HannesWell:
> > But since there are no objections against this change, I'll submit this now in order to make the latest ECJ also available for verification builds.
> 
> Ok, the Jenkins build of ECJ itself is objecting. Will look into that tomorrow.

I think was because `3.40.0-SNAPSHOT` is considered to be smaller than 3.40 in the logic of Maven's version comparator, isn't it?
Because when setting the lower-bound to `3.39` it works.
@laeubi can you confirm that, Maven always uses the latest available version that's within the range?
When running the build locally with `-X` it said that the tycho-compiler-plugin has `org.eclipse.jdt:ecj:jar:3.40.0-SNAPSHOT` on it's classpath even though a version 3.39 of ECJ is also available on Maven-central.

So I think this is now as desired and finally ready for submission.

HannesWell:
Since we still have comparator issues in some verification builds we should make sure that these verification builds have `--update-snapshots` respectively it's short form `-U` specified.

HannesWell:
> Since we still have comparator issues in some verification builds we should make sure that these verification builds have `--update-snapshots` respectively it's short form `-U` specified.

In the shared GH workflow this is set, but only for some `Jenkinsfile`:
https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/blob/fdac5e82dc0a837b9ee2b3eddf3237c35720e1b4/.github/workflows/mavenBuild.yml#L82

PR #2539, opened 2024-11-13
[Build] Unify and clean-up list of declared CBI plugin-repositories

HannesWell:
From the three declared Eclipse CBI plugin repositories only the maven-plugins with groupId 'org.eclipse.cbi.maven.plugins' are obtained. And at least with respect to these mavne-plugins, the repository at 
- https://repo.eclipse.org/content/repositories/cbi

seems to be the combination of the co-located repositories:
- https://repo.eclipse.org/content/repositories/cbi-releases
- https://repo.eclipse.org/content/repositories/cbi-snapshot

Therefore it seems unnecessary to declare all three repositories.

@akurtakov can you tell if my analysis and this change is correct or if not, can you tell who can tell?

### DIFF
--- a/eclipse-platform-parent/pom.xml
+++ b/eclipse-platform-parent/pom.xml
@@ -42,6 +42,8 @@
     <buildTimestamp>${maven.build.timestamp}</buildTimestamp>
     <buildType>I</buildType>
     <buildId>${buildType}${buildTimestamp}</buildId>
+    <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
+
      <!--
       releaseName should match the yearly "release train" name.
       For example, "Mars", "Mars.1", "Luna SR2", etc.
@@ -85,7 +87,6 @@
     <!-- only used when Tycho snapshot repo is enabled in <pluginRepositories> further down -->
     <tycho-snapshot-repo.url>https://repo.eclipse.org/content/repositories/tycho-snapshots/</tycho-snapshot-repo.url>
 
-    <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
     <cbi-jdt-repo.url>https://repo.eclipse.org/content/repositories/eclipse-staging/</cbi-jdt-repo.url>
     <cbi-ecj-version>3.40.0.v20241112-0530</cbi-ecj-version>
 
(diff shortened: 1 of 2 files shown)


akurtakov:
The analysis seem to be correct. 
Please even remove 
```
<releases>
        <enabled>true</enabled>
      </releases>
```
as it's the default and just clutters the pom.

HannesWell:
> The analysis seem to be correct. 

Great. Thanks for the confirmation!

Can you tell where the 'merge' happens? In the distribution management of o.e.cbi I can only see the releases and snapshots repo being defined:
https://github.com/eclipse-cbi/org.eclipse.cbi/blob/5d0a693e90524c337f088f91aa0e9432153c5d39/pom.xml#L65-L81

Or is this some kind of virtual repository?


> Please even remove
> 
> ```
> <releases>
>         <enabled>true</enabled>
>       </releases>
> ```
> 
> as it's the default and just clutters the pom.

Yes, with joy. According to https://maven.apache.org/pom.html#Repositories and the doc proposed by sonar lint, even snapshots=true is the default. So it can be removed too.

Should we submit this then now or also hold this back until the next dev cycle?

akurtakov:
> > The analysis seem to be correct.
> 
> Great. Thanks for the confirmation!
> 
> Can you tell where the 'merge' happens? In the distribution management of o.e.cbi I can only see the releases and snapshots repo being defined: https://github.com/eclipse-cbi/org.eclipse.cbi/blob/5d0a693e90524c337f088f91aa0e9432153c5d39/pom.xml#L65-L81
> 
> Or is this some kind of virtual repository?

I really have no idea. Maybe @fredg02  can answer that.

> 
> > Please even remove
> > ```
> > <releases>
> >         <enabled>true</enabled>
> >       </releases>
> > ```
> > 
> > 
> >     
> >       
> > as it's the default and just clutters the pom.
> 
> Yes, with joy. According to https://maven.apache.org/pom.html#Repositories and the doc proposed by sonar lint, even snapshots=true is the default. So it can be removed too.
> 
> Should we submit this then now or also hold this back until the next dev cycle?

Let's do it at once in the next dev cycle.

HannesWell:
> > Should we submit this then now or also hold this back until the next dev cycle?
> 
> Let's do it at once in the next dev cycle.

Acknowledge.


> It's a group repository and it combines the releases and the snapshots repositories. You can also see all of the group repos here:
> 
> https://repo.eclipse.org/#view-repositories

Thank you for the confirmation. Good to know.

PR #2545, opened 2024-11-18
Update the target platform to the Orbit and EMF release repositories

merks:


merks:
I'm going to ignore the annoying, recurring heap space error.

HannesWell:
> I'm going to ignore the annoying, recurring heap space error.

Since I recently increased the maxRAM for the I-build container we could also create the max-RAM for the JVM a bit.

PR #2558, opened 2024-11-22
Move 4.34-I builds to R4_34_maintenance branch

MohananRahul:
https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/issues/2557


JenkinsJobs/JobDSL.json
```diff
@@ -3,6 +3,6 @@
         "4.34"
     ],
     "Branches": {
-		"4.34": "master"
+		"4.34": "R4_34_maintenance"
```
akurtakov, line 6: As jobs are not recreated from branch and this will become 4.35:master in a next commit I believe this is not really needed.

MohananRahul: Yes not required in maintenance branch

akurtakov (approved):

Issue #2203, opened 2024-08-01
Auto-assign milestones to merged PR's


iloveeclipse:
Follow up on https://github.com/eclipse-jdt/eclipse.jdt.core/discussions/2627

We should automatically assign milestones to merged PR's to simplify our life tracking which release contained which fix etc.

There are various solutions possible, see list below fro what google found:

- https://github.com/eclipse-theia/theia/pull/8631
- https://github.com/DataDog/dd-trace-java/blob/master/.github/workflows/add-milestone-to-pull-requests.yaml
- https://www.jessesquires.com/blog/2022/08/04/automatically-assign-milestones-with-github-actions/

This should be done for all eclipse SDK projects (Platform, JDT, PDE, Equinox).

Note, differently to Platform, JDT has BETA_JAVA23 milestone](https://github.com/eclipse-jdt/eclipse.jdt.core/milestones?direction=asc&sort=due_date&state=open) that should be excluded from the "automated" process.

Beside this I believe all projects should be very similar in what milestones we have, as they are created from a job during release preparation process AFAIK.

@HannesWell, @laeubi : you seem to have some experience in the github actions area, I would really appreciate if you could help here.

HannesWell:
Sounds good and is probably not too difficult to implement (haven't looked into the examples yet).
I can try to help with that next week (I'm away over the weekend).

laeubi:
For Tycho I use `is:pr is:merged no:milestone` and then batch assign milestones or `is:issue no:milestone is:closed`.

> We should automatically assign milestones to merged PR's to simplify our life tracking which release contained which fix etc.

Github also allows to see what has changed when one creates a release and can generate a list or one can generate it form the Tags created in a repository.

Beside that assigning a milestone to a *closed* PR is somewhat useless, a Milestones main purpose is to track what works needs to be done (or is planned) see for example https://github.com/eclipse-tycho/tycho/milestones but from my experience no one is really doing such planning, but your BETA_23 is a good example, if there is work that needs to be done until one can declare "finish" of 23 support then a milestone (e.g. Java 23 Support) would be a good choice to let people know what is left.

iloveeclipse:
> Beside that assigning a milestone to a _closed_ PR is somewhat useless

There are JDT & platform committers that don't think so, me including. Also looking on the links, there are projects like theia which are also doing this, so it seem to be not useless also for other projects.

[...]

PR #2571, opened 2024-11-23
Add manual trigger for 'updateRelease'-workflow and use project-bot

HannesWell:
Adding a manual trigger allows to re-run that job manually with the latest changes applied to the referenced workflow from the releng.aggregator repository.

And since we have a eclipse bot specifically for each project we can use that bot as author of these automated changes instead of the generic GH bot.

### DIFF
--- a/.github/workflows/updateRelease.yml
+++ b/.github/workflows/updateRelease.yml
@@ -2,10 +2,26 @@ name: Update For Next Release
 
 on:
   workflow_call:
+    inputs:
+      nextReleaseVersion:
+        description: 'The version of the release, for example: 4.35'
+        type: string
+        required: true
+      botName:
+        description: The name of the bot that authos the changes
+        type: string
+        default: 'Eclipse Releng Bot'
+      botMail:
+        description: The name of the bot that authos the changes
+        type: string
+        default: 'eclipse-releng-bot@eclipse.org'
+    secrets:
+      githubBotPAT:
+        description: The personal access token (with scope 'public_repo') of the bot to push a required change to a branch.
+
 jobs:
   update:
     runs-on: ubuntu-latest
-    if: contains(github.event.milestone.description, 'Release') 
     steps:
     - uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
       with:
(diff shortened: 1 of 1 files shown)


HannesWell:
The changes necessary in the caller to use this are worked out in
- https://github.com/eclipse-equinox/equinox/pull/702

All other callers have to be adapted in a similar way.

.github/workflows/updateRelease.yml
```diff
@@ -23,20 +40,19 @@ jobs:
...
-        title: Update for release ${{ github.event.milestone.title }}
-        body: A new release milstone was created, please review the changes and merge if appropriate.
+        token: ${{ secrets.githubBotPAT }}
+        commit-message: Update for release ${{ inputs.nextReleaseVersion }}
+        branch: update_R${{ inputs.nextReleaseVersion }}
+        title: Update for release ${{ inputs.nextReleaseVersion }}
+        body: A new release milestone was created, please review the changes and merge if appropriate.
+        author: ${{ inputs.botName }} <${{ inputs.botMail }}>
         delete-branch: true
-        milestone: ${{ github.event.milestone.number }}
```
laeubi, line 37: Now it no longer adds the proper milestone link to the PR ...

HannesWell: Yes. I thought this could be covered by https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/issues/2203.

But we can probably also just keep this. If the workflow is triggered manually the number will hopefully be just null and the milestone not assigned automatically.
I'll try it out.

HannesWell: Just keeping it works as expected, so I reverted the removal.

.github/workflows/updateRelease.yml
```diff
@@ -23,20 +40,20 @@ jobs:
...
+          org.eclipse.tycho:tycho-versions-plugin:set-version -DnewVersion=${{ inputs.nextReleaseVersion }}.0-SNAPSHOT
+          org.eclipse.tycho:tycho-versions-plugin:set-parent-version -DnewParentVersion=${{ inputs.nextReleaseVersion }}.0-SNAPSHOT
+    - name: Create Pull Request for Release ${{ inputs.nextReleaseVersion }}
       uses: peter-evans/create-pull-request@5e914681df9dc83aa4e4905692ca88beb2f9e91f # v7.0.5
       with:
-        commit-message: Update for release ${{ github.event.milestone.title }}
-        branch: update_R${{ github.event.milestone.title }}
-        title: Update for release ${{ github.event.milestone.title }}
-        body: A new release milstone was created, please review the changes and merge if appropriate.
+        token: ${{ secrets.githubBotPAT }}
```
HannesWell, line 49: In fact we could also just use the default Github token issued for the workflow as usual.
The only difference would be that the PR would be created by the GH bot instead of the project-specific eclipse bot.
The resulting commit would be the same both would work, but I think it's a bit better to use the eclipse-bot accounts for that. And since we have all the PAT setup and in use already I don't see a drawback in using it again (besides a few more lines of configuration).

laeubi: Maybe just use the github one as a default

HannesWell: Applied that, although in general I think the goal should be to call it with equivalent parameters from everywhere.

HannesWell:
With https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/pull/3078 this is obsolete and can be discarded.

PR #2573, opened 2024-11-24
Define a single property for the releaseVersion and use in property

laeubi:
Actually the both properties 'releaseNumberSDK' and 'releaseNumberPlatform' are always the same and even if the comment mentions they might be different at some time it is better to unify them until then.

Because of this there is now a new 'releaseVersion' defined as a single property.


akurtakov:
One less thing to edit and possibly mismatch is definetely the right thing to do. Merging.

Issue #2564, opened 2024-11-22
Create 4.35-I-builds repo


MohananRahul:

MohananRahul:
Done Via https://ci.eclipse.org/releng/job/Releng/job/newStreamRepos/14/

Repo created :
https://download.eclipse.org/eclipse/updates/4.35-I-builds/
https://download.eclipse.org/eclipse/updates/4.35-Y-builds/
https://download.eclipse.org/eclipse/updates/4.35-P-builds/

PR #2574, opened 2024-11-24
Use placeholder to define the integration build repository

laeubi:
Currently we use fixed strings that regularly needs to be updated, but
the contents could actually be computed automatically.

This replace fixed strings with using properties making it needless to
change them all the time. (this also includes the changes from https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/pull/2573)

This would need just one little adjustment in the process, the **next stream** I-Build repository must point to the **previous stream** until the first I-Build happens.

So what do you think? @merks can you tell how much effort it would be regarding the inclusion of the previous stream? I'm not sure how the exact process is with the ibuild repos.


merks (approved):
Seems fine.  But I don't have in-depth knowledge about the details.

merks:
> So what do you think? @merks can you tell how much effort it would be regarding the inclusion of the previous stream? I'm not sure how the exact process is with the ibuild repos.

Sorry, I don't really understand the question?

eclipse-platform-parent/pom.xml
```diff
@@ -77,9 +78,9 @@
       'eclipiserun-repo' repository, such as for computing .api-descriptions and
       generating API Tools reports.
     -->
-    <eclipserun-repo>https://download.eclipse.org/eclipse/updates/4.34-I-builds/</eclipserun-repo>
+    <eclipserun-repo>https://download.eclipse.org/eclipse/updates/${releaseVersion}-${buildType}-builds/</eclipserun-repo>
```
akurtakov, line 81: Should releaseVersion be 4.35 (not 4.35.0) for this to work? There shouldn't be https://download.eclipse.org/eclipse/updates/4.35.0-I-builds/

akurtakov (commented):
Overall looks good, just the question about releaseVersion  containing micro version .

eclipse-platform-parent/pom.xml
merks: Yes, good point.

laeubi:
> Sorry, I don't really understand the question?

Currently we have https://download.eclipse.org/eclipse/updates/4.34-I-builds and the ibuilds seem to push their content there but:

1. at some point (maybe at GA?!?) the content vanishes
2. there is some magic to mark a build unstable
3. ... maybe more ...
4. At some point in time, there will be https://download.eclipse.org/eclipse/updates/4.35-I-builds populated and we change that

So what would be required is that unless "point 4" has happened **there already is** https://download.eclipse.org/eclipse/updates/4.35-I-builds but simply points to https://download.eclipse.org/eclipse/updates/4.34-I-builds

So we can use the "new" URL already even though it has not yet the content (and just is an alias).

> Should releaseVersion be 4.35 (not 4.35.0) for this to work? 

Yes that's because I'm using an outdated version here, I'll rebase this on master.

merks:
I see.    The empty state of https://download.eclipse.org/eclipse/updates/4.35-I-builds is indeed rather inconvenient.  Unfortunately I have no knowledge of how these composites are maintain/managed.  😞

HannesWell (approved):
To me this looks good as well, during the release process it should just be made sure that upon creation of the new I-build repo should get the linked to the latest RC as mentioned already.
At least the repository already exists a.t.m. but is empty:
https://download.eclipse.org/eclipse/updates/4.35-I-builds/

Their creation is tracked in https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/issues/2564

> I see. The empty state of https://download.eclipse.org/eclipse/updates/4.35-I-builds is indeed rather inconvenient. Unfortunately I have no knowledge of how these composites are maintain/managed. 😞

https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/issues/2564 mentions the Jenkins job https://ci.eclipse.org/releng/job/Releng/job/newStreamRepos/.

I think in general all `eclipse` (-SDK) repos are managed through the Jenkins jobs in https://ci.eclipse.org/releng/job/Releng/, which are defined in
https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/tree/5bddc6c2de3374aed50318b550b71222f2963f15/JenkinsJobs/Releng

From the definition of the `newStreamRepos` job I think after the https://download.eclipse.org/eclipse/updates/template_repo/ has been copied we should 'simply' do the same as the I-build does when it adds a new I-build sub-repo, just linking the latest RC. Or the new stream repos should already be created before RC2 and adding an RC should be part of the RC2+x process in order to always use the latest RC.

HannesWell:
The commit message does not match the content anymore.

Issue #2550, opened 2024-11-20
Preparation work for 4.35(2025-03) and open master for development


MohananRahul:
This preparation work involves the following tasks. For previous issue please refer to eclipse-platform/eclipse.platform.releng.aggregator#2274

- [x] eclipse-platform/eclipse.platform.releng.aggregator#2554
- [x] eclipse-platform/eclipse.platform.releng.aggregator#2556
- [x] eclipse-platform/eclipse.platform.releng.aggregator#2557
- [x] eclipse-platform/eclipse.platform.releng.aggregator#2559
- [x] eclipse-platform/eclipse.platform.releng.aggregator#2562
- [x] eclipse-platform/eclipse.platform#1634
- [x] #2564
- [x] #2567
- [x] #2568
- [x] Move previous version to 4.34RC2 across build scripts
- [x] #2577
- [x] #2580
- [x] eclipse-platform/eclipse.platform.releng.aggregator#2555
- [x] #2578
- [x] Create new I-build job for 4.35 release
- [x] #2586
- [x] #2588
- [x] Version bumps for 4.35 stream  Clean in https://download.eclipse.org/eclipse/downloads/drops4/I20241126-0110/buildlogs/reporeports/reports/versionChecks.html
- [x] #2607
- [x] Update build calendar for platform 4.35 release
- [x] Deploy ecj compiler from 4.34 GA and use it for 4.35 M1 build
- [x] Update generic repos I-builds, Y-builds, P-builds to point to 4.35 repos

HannesWell:
> * Move previous version to 4.34RC2 across build scripts

@MohananRahul or @akurtakov can you tell if it would be possible to perform this step before creating all the version updated?
I.e. could we set the api- and comparator-baseline repo property `previous-release.baseline` to the current RC2-repo before the version updates are triggered, IIRC by creating the milestones? Then the workflows that perform the POM parent version updates could also immediately perform the version updates necessary in the bundle/feature projects because the project is changed the first time in a dev-cycle. See also https://github.com/eclipse-pde/eclipse.pde/pull/1489#issuecomment-2494993698

Maybe this should also be included in https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/pull/2565?
[...]

PR #2576, opened 2024-11-24
Run prepareRelease workflow manually and set 'previous-release.baseline'

HannesWell:
As described in https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/issues/2550#issuecomment-2495044382 ff running the `prepareRelease` workflow and submitting the automatically created changes in the `eclipse-platform-parent/pom.xml` before the milestones are created and thus the `updateRelease` workflow is triggered in all SDK repos, will ensure that in the parent-version update PR created in the other repos the project versions can already be bumped by the automatic version bump workflow. With that all release preparation changes necessary in the other repos are created automatically and one just has to review the PRs and submit it.

### DIFF
--- a/.github/workflows/prepareRelease.yml
+++ b/.github/workflows/prepareRelease.yml
@@ -1,7 +1,16 @@
 name: Prepare Next Release
 on:
-  milestone:
-    types: [created]
+  workflow_dispatch:
+    inputs:
+      nextReleaseName:
+        description: 'Next release name, e.g.: 2025-03'
+        required: true
+      nextReleaseVersion:
+        description: 'Next release name, e.g.: 4.35'
+        required: true
+      baselineRepository:
+        description: 'Previous release baseline repository, e.g.: 4.34-I-builds/I20241120-1800'
+        required: true
 
 permissions:
   contents: read
@@ -9,7 +18,6 @@ permissions:
 jobs:
   prepare:
     runs-on: ubuntu-latest
-    if: contains(github.event.milestone.description, 'Release')
     permissions:
       pull-requests: write
       contents: write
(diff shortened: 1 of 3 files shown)


.github/workflows/prepareRelease.yml
```diff
@@ -1,15 +1,23 @@
 name: Prepare Next Release
 on:
-  milestone:
-    types: [created]
```
laeubi, line 4: I don't see how running the workflow manually helps in any way, the whole purpose of this is to automate the process, e.g. exactly this manually run was performed before (but not with an action).

If one want some "timed" process, then better the process to create the milestones should be adjusted, for example one can create first the milestones in the aggregator, then this workflow runs and this can then trigger the creation of the milestones in other repositories.

laeubi (changes requested):

.github/workflows/prepareRelease.yml
```diff
@@ -1,15 +1,23 @@
...
+  workflow_dispatch:
+    inputs:
+      nextReleaseName:
+        description: 'Next release name, e.g.: 2025-03'
+        required: true
+      nextReleaseVersion:
+        description: 'Next release name, e.g.: 4.35'
+        required: true
+      baselineRepository:
+        description: 'Previous release baseline repository, e.g.: 4.34-I-builds/I20241120-1800'
```
laeubi, line 12: With

- https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/pull/2574

and 

- https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/pull/2575

we can save the update altogether by simply using placeholders.
-

HannesWell: I don't see how this would work in the peride after RC1 and GA, the one we are currently in.

It used to be `https://download.eclipse.org/eclipse/updates/4.33/R-4.33-202409030240/`
now it is `https://download.eclipse.org/eclipse/updates/4.34-I-builds/I20241120-1800/`
and it will be `https://download.eclipse.org/eclipse/updates/4.33/R-4.33-2024XXXXXXXX`.

Besides the transition period between RC2 and GA, #2575 misses the date component.

HannesWell:
> If one wants, it would even work to create the (release) milestone manually in this repository using Github UI first.

In general it still would be nice to have the release calendar (not only the milestone) a better accessible format/location than currently the google calendar. Then it would be easier to derive the milestone dates in the release preparation Jenkins pipeline automatically. Maybe we can export the calendar to a file into this git repo and read it in the release preparation workflow and also the freeze check (which currently relies on an internal API).
But we can discuss this separately later.

PR #2560, opened 2024-11-22
Remove unused apifreeze releng scripts

akurtakov:
Not used in years and just complicate the scripts and stream opening procedures.
Fixes https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/issues/2555 .

PR #2582, opened 2024-11-25
Remove useless step from newReleasePrep.sh

akurtakov:
Since https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/pull/2560 the cleanup aproved api list task is no longer needed (actually it wasn't for years).


akurtakov:
Merging as this script is only used to create the main task for opening new stream so it won't be used in the next ~3 months.

PR #2584, opened 2024-11-25
Do not call removed dofreezereport task

akurtakov:
It has been removed with
https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/pull/2560 as it was not configured to do anything for years.


akurtakov:
Merge as verification builds doesn't execute this script at all.

PR #2593, opened 2024-11-26
Touch o.e.help feature

akurtakov:


akurtakov:
https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/pull/2592 shows that successful verification build doesn't mean I-build will succeed too thus merging and triggering I-build.

PR #2596, opened 2024-11-26
Do not report deprecations for 3rd party libs

akurtakov:
Looking at
https://download.eclipse.org/eclipse/downloads/drops4/I20241126-0600/apitools/deprecation/apideprecation.html shows deprecations from Apache Commons IO which are irrelevant for the project as API stability for 3rd party libs can not be guaranteed thus add commons-io to the excluded list and pass the list to apitooling.apideprecation ant task.


akurtakov:
The code changed is executed only by I-builds thus merging without waiting.

PR #2597, opened 2024-11-26
Remove links to non existing javadoc logs

akurtakov:
Javadoc failures break the build for some time and no separate log files are kept for later inspection.


akurtakov:
Code used only by I-build generated sites thus merging without waiting.

PR #2599, opened 2024-11-27
Remove remnants from git.eclipse.org

akurtakov:
No longer used so these sed calls just make scripts more complex.


akurtakov:
I-build used only thus merging directly.

PR #2600, opened 2024-11-27
Update to latest ecj snapshot version

akurtakov:


akurtakov:
Oops, merged the maven gpg passprase by mistake with this one but it wouldn't have been verified without running I-build so I would have to merge it again without waiting. Starting new I-build to verify.

PR #2603, opened 2024-11-28
Fix undo / redo icons according to standard

BeckerWdf:
Contributes to:
https://github.com/eclipse-platform/eclipse.platform.images/issues/113


BeckerWdf:
failure: "Caused by: java.lang.OutOfMemoryError: Java heap space" is unrelated

PR #2608, opened 2024-11-29
Update previous release version to 4.34 GA across build scripts

MohananRahul:
Fix https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/issues/2607

### DIFF
--- a/cje-production/P-build/buildproperties.txt
+++ b/cje-production/P-build/buildproperties.txt
@@ -42,17 +42,17 @@ LOCAL_REPO="localMavenRepo"
 BASEBUILDER_TAG="4.34"
 API_PREV_REF_LABEL="4.31"
 
-PREVIOUS_RELEASE_VER="4.33"
-PREVIOUS_RELEASE_REPO_ID="4.33"
+PREVIOUS_RELEASE_VER="4.34"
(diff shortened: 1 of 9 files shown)


MohananRahul:
Merging this PR as verification builds will not verify.

HannesWell:
Isn't this applied too early? 4.34 is not yet released.
As far as I can tell, this currently breaks the I-build (even though it currently seems to be more severely broken at a infrastructure level).

MohananRahul:
> Isn't this applied too early? 4.34 is not yet released. As far as I can tell, this currently breaks the I-build (even though it currently seems to be more severely broken at a infrastructure level).

The GA has already been promoted to SimRel, and it’s available—it's just not yet visible.

HannesWell:
> > Isn't this applied too early? 4.34 is not yet released. As far as I can tell, this currently breaks the I-build (even though it currently seems to be more severely broken at a infrastructure level).
> 
> The GA has already been promoted to SimRel, and it’s available—it's just not yet visible.

Indeed. The actual problem was that one property was not updated accordingly in this change. Did it now in
- https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/pull/2612

Issue #2610, opened 2024-11-30
Completly broken I-build


HannesWell:
Since Thursday night the I-build is completely broken and fails to check-out the git repository:
https://ci.eclipse.org/releng/job/Builds/job/I-build-4.35/16/console
```
 + git clone -b master git@github.com:eclipse-platform/eclipse.platform.releng.aggregator.git
 Cloning into 'eclipse.platform.releng.aggregator'...
 Could not create directory '/home/jenkins/.ssh'.
 Host key verification failed.
 fatal: Could not read from remote repository.
 
 Please make sure you have the correct access rights
 and the repository exists.
```

This looks like an infrastructure problem and the ssh-agent does not seem available:
https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/blob/0e4fb784ed9c401cf861fd0ee03f5fbe0f00b725/JenkinsJobs/Builds/I_build.groovy#L95

HannesWell:
Gitlab issue already exists:
- https://gitlab.eclipse.org/eclipsefdn/helpdesk/-/issues/5366

PR #2611, opened 2024-11-30
[Build] Run I-/Y-/P-builds on ubuntu-2404

HannesWell:
and run all other infrastructure-work jobs on the 'basic' image.

See https://www.eclipse.org/lists/cross-project-issues-dev/msg19990.html

Fixes https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/issues/2610


HannesWell:
The verification build does not test this and to unblock the I-build I submit this now immediately.
If the agents need further refinement we can create a follow-up.

PR #2615, opened 2024-12-01
[Build] Remove dormant win32 smoke test

HannesWell:
The smoke-test 'ep-smoke-test-win32' is not executed and we have a dedicated I-build-test that runs on the same machine. Therefore this smoke-test is completely obsolete.


HannesWell:
The current instantiation of that job has no recorded executions and has to be deleted manually once this is submitted:
- https://ci.eclipse.org/releng/job/SmokeTests/job/ep-smoke-test-win32/

PR #2522, opened 2024-11-06
[I-Build-Tests] Run Linux tests on 'ubuntu-2404' image agent

HannesWell:
Follow-up for https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/pull/2515

I have already replayed the current linux-java17 configuration to use the ubunut image for the smaller `ui` test-suite:
https://ci.eclipse.org/releng/job/AutomatedTests/job/ep434I-unit-linux-x86_64-java17/21/

Once that has passed I'll create a follow-up run with the complete `all` suite to see if all tests pass, then I think this change is save enough to apply it to all other configs as well.

PR #2616, opened 2024-12-01
[Build] Remove duplicating Ubuntu smoke-tests

HannesWell:
Running smoke-tests on Ubuntu is not necessary anymore since I-build tests are already executed on Ubuntu now.

See https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/pull/2522


akurtakov:
Makes perfect sense. At some point it would be nice if someone has the time to monitor RHEL/Fedora smoke tests runs but that should be left for whenever such a person appears.

PR #2617, opened 2024-12-01
[Build] Define complete configuration of StartSmokeTests job via Job-DSL

HannesWell:
This way the complete configuration of the 'Start-smoke-tests' job is under version control.
The only change is that this moves the 'Start-smoke-tests' job from the root of the Jenkins instance into the 'SmokeTests' folder, where all dedicated smoke-tests are located too.
This simplifies and cleans-up the structure of the Releng-Jenkins.


HannesWell:
The cron trigger from the Start-smoke-tests job is now removed again since it's always triggered by I- and Y-builds and if I understood the specification `15 4 16 2 4` correctly, it runs only on the 16th of Feburary and only if it's the fourth day of the week. So round about every seven years?


Furthermore I added another commit to define JDK-URLs within the `StartSmokeTests` Jenkinsfile, which has the advantage that changes are immediately visible upon the next execution without the need to run the seed-job again and without the one-job-delay that happens if parameters change.

Furthermore only the latest Temurin JDK builds of the corresponding (feature) release in all smoke tests. This reduces the maintenance effort respectively ensures that old JDK builds are automatically replaced.

HannesWell:
For the record and for reference, since I'm no going to delete the old, manually maintained job configuration (i.e.https://ci.eclipse.org/releng/job/Start-smoke-tests/), this is the status overview of it's last executions:

grafik

PR #2609, opened 2024-11-30
[Build] Simplify bootstrapping of I-build tests and fetch configuration files directly from git

HannesWell:
Instead of fetching the test-configuration files from eclipse.downloads and push it there in each I-build, simply get the test-configuration files directly from the eclipse.platform.releng.aggregator git repository, using partial-clones for improved performance:
- https://git-scm.com/docs/partial-clone
- https://github.blog/open-source/git/get-up-to-speed-with-partial-clone-and-shallow-clone/

This avoids the need to pack and upload the test-configuration zip-file and also simplifies fetching the test-configuration for other branches, like maintenance branches or to test changes.

This also removes from the getEBuilder.xml ant script the possibility to download the test-configuration zip file and to use a local zip.
Furthermore this removes from the I-/Y-builds the creation and publication of the test-configuration zip-file.

As a consequence remove the definitions of the properties `EBUILDER_HASH` and `downloadURL` that are not necessary anymore to be passed to the `getEBuilder` ant-script. And remove the `eclipseStream` property that now can be computed by the 'getEBuilder' ant-script. 
Also rename the folder containing the `getEBuilder.xml` from to just `bootstrap` since Hudson is not used anymore.

PR #2618, opened 2024-12-01
[Build] Create consolelog.txt file in test execution directory again

HannesWell:
This fixes the regression that the the <testedPlatform>_consolelog.txt is created at the wrong location.
It was missed that the <testedPlatform>_consolelog.txt file was created in the executionDir (because the scripted inlined ran in that directory) in https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/pull/2516

See https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/pull/2516#issuecomment-2509795054

### DIFF
--- a/production/testScripts/runTests2.xml
+++ b/production/testScripts/runTests2.xml
@@ -816,13 +816,13 @@
     depends="init"
     unless="env.TESTING_TEST_XML">
 
-    <mkdir dir="results/consolelogs"/>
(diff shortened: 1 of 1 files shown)


HannesWell:
Verified this by replaying one of the I-build tests and redirecting the configuration to this branch with the help of https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/pull/2609.

The verification build does not test this.

PR #2619, opened 2024-12-01
[Build] Fix reference to secManager parameter in smoke test jobs

HannesWell:
Maybe this helps to fix some of the smoke-test failures.


HannesWell:

Issue #2625, opened 2024-12-03
Unify Y-build and I-build configurations and their tests


HannesWell:
Recently I have reworked the definition of I-build tests to reduce duplication and to make them more unified and simpler.
This has the effect that the definition of Y-build tests has now significantly diverged from the I-build tests and I intended to apply the same simplifications for Y-build tests as well.

And while looking into all the definitions of I-builds and Y-builds I noticed that they have a lot in common and wonder respectively think we should try to unify their definitions and try to define both pipelines from one source, while considering the points where they are different.
But this questions what exactly the differences between I-builds and Y-builds are? What I have found so far is

- Y-builds built another branch than the master, at least for jdt.core the development branch of ECJ supporting the next Java release
- Y-build tests have one linux setup that uses the latest EA JDK instead of the latest release
- The results of Y-builds are published to another website than I-builds
- Y-builds publish their build p2-repo e.g. to
https://download.eclipse.org/eclipse/updates/4.35-Y-builds/ 
instead of https://download.eclipse.org/eclipse/updates/4.35-I-builds/
- Y-builds don't publish to a Maven snapshot repository (I assume)

Together with https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/issues/1950 this should to further simplify the overall releng.
Of course a unification would have to happen in multiple steps.

An alternative solution would be to just develop ECJ's support for upcoming java versions on the master branch and stop using another branch. AFAICT this has already been discussed respectively is currently discussed but will not be implemented soon.

The questions I have to the JDT-team:
- Are you fine with the proposed general goal?
- Who of you is interested to discuss the details that come along the path of making this happen?

@jarthana, @stephan-herrmann or @iloveeclipse are you interested in discussing this or can you tell who is interested?

## Community

- [x] I understand suggesting an enhancement doesn't mandate anyone to implement it. Other contributors may consider this suggestion, or not, at their own convenience. The most efficient way to get it fixed is that I implement it myself and contribute it back as a good quality patch to the project.

stephan-herrmann:
I am interested in seeing successful Y-builds with useful test result, ideally with no big gaps around the times of release, or infra changes. If unification helps to this end, then surely I fully agree with the goals, and your description looks correct to me.

I don't, however, have much insight into the inner workings of I-builds nor Y-builds, so I don't know in which detail discussion I could help. I'm already looking into P-builds, which is about all I can contribute in terms of releng builds.

[...]

PR #2626, opened 2024-12-04
[Build] Unify Y-build test job definitions into one file per OS

HannesWell:
A first part of https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/issues/2625


HannesWell:
I have verified with my local Jenkins instance that the job configurations generated from the updated groovy-files is effectively equal to the existing configuration. And since the verification build does not test this, this can be submitted immediately.

PR #2630, opened 2024-12-05
Use webtools 3.36

akurtakov:


akurtakov:
Not used outside of I-build thus merging.

PR #2481, opened 2024-10-26
[I-Build] Unify test job names to schema os-arch-javaVersion

HannesWell:
and further simplify variable interpolation.
This simplifies later additions of new os-arch combinations.

I expect that this will change the labels and URLs of test reports accordingly, for example in
https://download.eclipse.org/eclipse/downloads/drops4/I20241025-1840/

But I assume that's not a problem, is it?
@akurtakov or @iloveeclipse or anybody else do you have any objections or do you anything that could be missing?

I tested locally that the changes in how to apply the variables in the job configs leads to the desired results, so from my side this is ready.

PR #2633, opened 2024-12-05
[Build] Unify Y-build test job names to schema os-arch-javaVersion

HannesWell:
This simplifies later additions of new os-arch combinations.

This is similar to https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/pull/2481, just for Y-build tests.


HannesWell:

PR #2636, opened 2024-12-05
[Build] Unify all smoke test job configurations to be Jenkins-pipelines

HannesWell:
and simplify the definition of kubernetes-agents by inheriting the base configuration from the 'basic' image.


HannesWell:

PR #2638, opened 2024-12-06
[Build] Simplify setup and execution script for smoke tests

HannesWell:
Remove unnecessary print-outs and simplify the download and installation of the executing JDK.
Also use a downsized 'basic' agent to trigger the executions on specific test-agents. It's not necessary to occupy for 4GiB of precious memory while just waiting for the test-agents to complete.


HannesWell:

PR #2639, opened 2024-12-06
[Build] Use Jenkins matrix to define setup combinations in smoke tests

HannesWell:
This makes changes, e.g. using other java versions or adding a new platform simpler.
This also has the effect that ppc64le is now also tested with Java-21 and 23.


HannesWell:
It's so nice that changes to a Jenkinsfile are in effect immediately upon the next run and don't need the execution of the seed job.

PR #2643, opened 2024-12-06
[Build] Further simplify and unify I/Y-build test job configurations

HannesWell:
- Simplify custom JDK download in I-builds on Linux with Java-23 by using the same approach like for Y-builds
- Remove unnecessary '-Dargs=all' property definition for Windows and unnecessary environment variable resets on Mac.


HannesWell:
The verification build does not test this, therefore build failures are unrelated.

PR #2644, opened 2024-12-06
[Build] Double number of kept smoke test runs

HannesWell:
Smoke tests are launched by I-builds and Y-builds and their results are currently not stored elsewhere. Therefore we should keep 10 old runs instead of just five.


HannesWell:
Submitting this immediately since the verification build does not test this anyways.

PR #2645, opened 2024-12-07
[Build] Deploy RelEng scripts in master branch build

HannesWell:
This ensures that releng scripts are always updated on the download-server as soon as they change in the git repository. Furthermore this avoids the need for the separate Jenkins job to deploy the scripts, thus making
https://ci.eclipse.org/releng/job/Publish_shell_scripts_for_releng_jobs/ obsolete (can be deleted once this is submitted).

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -18,7 +18,7 @@ pipeline {
 				sh 'git submodule foreach "git fetch origin master; git checkout FETCH_HEAD"'
 			}
 		}
-		stage('Deploy eclipse-platform-parent pom and eclipse-sdk target') {
+		stage('Deploy parent-pom and SDK-target') {
 			when {
 				anyOf {
 					branch 'master'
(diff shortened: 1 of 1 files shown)


HannesWell:
Verified that this works by replaying the verification build and modified it to deploy to:
- https://download.eclipse.org/eclipse/relengScripts2-tries/

Will delete that entire folder again, together with the https://ci.eclipse.org/releng/job/Publish_shell_scripts_for_releng_jobs/ Jenkins job, once this is submitted.

HannesWell:
I have now added a 'staging' mechanism to minimize downtime, but I have to test this better.

HannesWell:
With the latest adjustments and after some more testing I think this is now ready.
The result can be seen in my testing area at:
- https://download.eclipse.org/eclipse/relengScripts2-tries/relengScripts/

In general I would be nice to just start the deployment if something has changed, e.g. by uploading the commit-id of the current commit too and then running git diff before uploading, but I assume that's not worth the effort?

akurtakov:
Please get this one in and checks can be following commits.

HannesWell:
> Please get this one in and checks can be following commits.

Yes. I just had it almost completely running quickly after my last message. So we have it with check now already. :)

Submitting now, as the verification build does not test this.

To bootstrap the `state` file this change I will have to `Replay` the master build with the modification that the git diff is skipped and the deployment happens unconditionally so that the `state` file is created initially.

HannesWell:
After unanticipated build failures on the initial run on master I figured out that the disposal location `download.eclipse.org/eclipse/relengScripts-disposal` was created by the `genie.releng` and can therefore not be deleted by the `genie.platform`:
```
 drwxr-sr-x   8 genie.platform eclipse.platform  4096 Dec 12 16:24 relengScripts/
 drwxr-sr-x   7 genie.releng   eclipse.platform  4096 Dec  8 12:30 relengScripts-disposal/
```

Deleting `download.eclipse.org/eclipse/relengScripts-disposal` from the releng JIPP (that uses the `genie.releng`) solved the problem now and the scripts could be deployed completely successful.
I'm going to delete the `Publish_shell_scripts_for_releng_jobs` job from the RelEng Jenkins now. It's now obsolete.

PR #2649, opened 2024-12-11
Configure NotThreadSafe javadoc tag

akurtakov:
Prereq for https://github.com/eclipse-equinox/equinox/pull/716


akurtakov:
Merging it in even if build fails for other reasons as going even down this rabbit hole means nothing will be completed.

PR #2637, opened 2024-12-05
[Build] Unify definition of all smoke tests into StartSmokeTests job

HannesWell:
Define the different smoke-test Jenkins agents in a scripted sub-part of the `StartSmokeTests` job and unify the multiple definitions of the executed shell-script into one.

PR #2655, opened 2024-12-12
[Build] Define all smoke-test parameters in job definition again

HannesWell:
Defining them only in the pipeline has the effect that the jobs forgets it's parameter each time the job-configuration is recreated by the seed-job and thus the subsequent first execution of the smoke-test fails because it has 'forgotten' it's parameters and has to re-discover them on the first pipeline execution again.

Follow-up on
https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/pull/2637


HannesWell:
Submitting as the verification build does not test this.

Issue #2654, opened 2024-12-12
Update Lucene to version 10


akurtakov:
New major Lucene version is out with claimed significant speed ups https://www.elastic.co/search-labs/blog/apache-lucene-10-release-highlights

akurtakov:
As per https://lists.apache.org/thread/4bhnkkvvodxxgrpj4yqm5yrgj0ppc59r version 9.12.0 shipped with 4.34 is the last minor 9.x version with only security releases aka 9.12.x to be done from 9.x stream. It's time to start preparing the update to 10.x as that would require a lot of changes in releng due to Java 21 requirement.

merks:
FYI, one strange thing I get is this failure:
>[ERROR]   Missing requirement: org.apache.lucene.core 10.0.0.v20241213-0927 requires 'java.package; jdk.incubator.vector 0.0.0' but it could not be found

I can make that dependency optional, on the assumption that it really is optional.  It's apparent used by this package:
```
org.apache.lucene.internal.vectorization jdk.incubator.vector
```
I can't find much detail about that.  Looking at https://lucene.apache.org/core/corenews.html I see it mentioned here:

![image](https://github.com/user-attachments/assets/61320b57-7db4-447e-8949-769d1a2ab3a6)

Given one needs to take explicit action to enable incubator modules and given they are documented to be definitely removed in the future, assuming it's optional seems reasonably safe...

laeubi:
> Given one needs to take explicit action to enable incubator modules and given they are documented to be definitely removed in the future, assuming it's optional seems reasonably safe...

Its worth a try, I just want to mention that with Multi_Release jars one can target specific JDK releases to use such features in a safe way and have it still being mandatory. But I haven't checked the details.

merks:
In my Oomph environment I end up resolving a target platform using the orbit aggregation nightly so ended up with a combination of lucene 9 and 10 and then ended up with a mylyn wiring disaster trying to launch an IDE.  It's a bit worrisome. 😱

ruspl-afed:
> mylyn wiring disaster

what should I know?

merks:
This is how the launch fails:
```
!ENTRY org.eclipse.mylyn.tasks.index.core 4 0 2024-12-14 17:12:08.719
!MESSAGE FrameworkEvent ERROR
!STACK 0
org.osgi.framework.BundleException: Could not resolve module: org.eclipse.mylyn.tasks.index.core [392]
[...]

PR #2666, opened 2024-12-16
Update Jenkins Build folder and I_unit_linux to use Java 21

akurtakov:
Needed by
https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/issues/2654


akurtakov:
Will not be verified by anything else but I-bulid thus merging.

Issue #2653, opened 2024-12-12
Longer CI-BUild and I-Build statistic history


jukzi:
Currently only 10 builds are kept. For logfile etc that is typically sufficient, but it would be helpfull to have longer statistics like Test Result Trend and buildTimeTrend
for exmple https://ci.eclipse.org/jdt/job/eclipse.jdt.core-Github/job/master/buildTimeTrend

![image](https://github.com/user-attachments/assets/9587275b-1b45-44a0-a168-739a5fad6898)
![image](https://github.com/user-attachments/assets/082963dc-f2f9-41d0-8925-29236f984056)

Especially after a vacation it's not visible what changed and if there had been silent regression in build time. A buildTimeTrend over the last quarter would be helpfull.
I don't know how to configure that, or if a special plugin is needed. https://plugins.jenkins.io/build-time-blame/ ?

akurtakov:
@fredg02 Can you help here?

fredg02:
* Nitpick: I see 15 builds. :slightly_smiling_face: 
* How can this be configured:  see https://github.com/eclipse-jdt/eclipse.jdt.core/blob/master/Jenkinsfile#L4
* We generally advise doing two things:
  * keep the number of builds reasonable
  * don't keep build artifacts around for more than 1 or 2 builds

   So I'd recommend setting this to something like:
   ```
   buildDiscarder(logRotator(numToKeepStr: '50', artifactNumToKeepStr: '2'))
   ```

---

>  A buildTimeTrend over the last quarter would be helpfull.

Taking https://ci.eclipse.org/jdt/job/eclipse.jdt.core-Github/job/master/ as a good example, here are the current stats:
* on the master branch, 15 builds are kept. Each build uses between 80-100 MB. Total size of the master branch job: 1.4 GB.
* between December 4 and today, the builds have been executed every day (except December 7) between 1 and 4 times
* So 15 builds would span a history between almost 4 days (with 4 builds a day) and a maximum of 15 days (with 1 builds per day, not counting days without builds)
* Let's scale this up to a quarter:
  * 90 days of 1 build per day => 90 builds with ~100 MB  => 9 GB or
  * 90 days of 4 builds per day => 360 builds with ~100 MB => 36 GB
  * I'd assume that the truth is somewhere in the middle, let's say 180 builds => 18 GB
[...]

PR #2668, opened 2024-12-16
Jenkins: longer i-Build history without artifacts #2653

jukzi:
to get better statistics

https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/issues/2653


jukzi:
cc @fredg02

PR #2669, opened 2024-12-16
Remove out of date compatibility note

jonahgraham:
Many years ago Activites weren't support in e4. That has long since been resolved and when the "not currently in 4.1" note was removed from the main help page in commit 5f66a3ea1ea9e71c32ce4fa63b5148530165bf16 the same update was not made to the table of contents.


akurtakov:
Build failed with OOM. As this is a docs change only merging without rerunning test. 
Thanks Jonah!

PR #2670, opened 2024-12-16
Adjust jenkins scripts for Java 21

akurtakov:
* Do not start Linux Java 17 tests
* Rename Windows tests to Java 21 to match version used
* Publish correct test results


akurtakov:
Scripts used only by I-builds thus merging directly.

Issue #2623, opened 2024-12-03
Changes needed to work with Java 24


merks:
The latest early access version of Java 24, i.e, 24+26 fails to start when `-Djava.security.manager=allow` is specified.  I.e., eclipse quietly fails to start; via the console we see why/how it fails to start:
```
$eclipse/eclipse.exe -consoleLog 2>&1 | tee log
CompileCommand: exclude org/eclipse/jdt/internal/core/dom/rewrite/ASTRewriteAnalyzer.getExtendedRange bool exclude = true
WARNING: Using incubator modules: jdk.incubator.vector
Error occurred during initialization of VM
java.lang.Error: A command line option has attempted to allow or enable the Security Manager. Enabling a Security Manager is not supported.
        at java.lang.System.initPhase3(java.base@24-beta/System.java:1948)
```
That option is specified in the eclipse.ini and is used in quite a few places (in the Platform SDK setup):

![image](https://github.com/user-attachments/assets/01f0e2bd-b1de-4ac6-a611-c0f154a10683)

Even if we were to remove this option from the product, existing installations would retain the option in the eclipse.ini when updating.

Starting with `-Djava.security.manager=disallow` allows the IDE to start, though it's not so pretty!
```
$eclipse/eclipse.exe -consoleLog -vmargs -Djava.security.manager=disallow 2>&1 | tee log
CompileCommand: exclude org/eclipse/jdt/internal/core/dom/rewrite/ASTRewriteAnalyzer.getExtendedRange bool exclude = true
WARNING: Using incubator modules: jdk.incubator.vector
WARNING: A terminally deprecated method in sun.misc.Unsafe has been called
WARNING: sun.misc.Unsafe::staticFieldOffset has been called by org.eclipse.osgi.internal.url.MultiplexingFactory (file:/D:/Users/test38/.p2/pool/plugins/org.eclipse.osgi_3.22.100.v20241127-0924.jar)
WARNING: Please consider reporting this to the maintainers of class org.eclipse.osgi.internal.url.MultiplexingFactory
WARNING: sun.misc.Unsafe::staticFieldOffset will be removed in a future release
Dez. 03, 2024 9:31:20 AM org.apache.aries.spifly.BaseActivator log
INFO: Registered provider org.slf4j.simple.SimpleServiceProvider of service org.slf4j.spi.SLF4JServiceProvider in bundle slf4j.simple
WARNING: A restricted method in java.lang.System has been called
[...]

PR #2659, opened 2024-12-14
[Build] Disallow Java security-manager in all I-/Y-/Smoke-tests

HannesWell:
Part of https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/issues/2623

@merks do you think we should explicitly disallow the security-manager or just remove the argument to explicitly enable it?
Since Java-18 the default value of the `java.security.manager` property is  `disallow` as part of [JEP-411](https://openjdk.org/jeps/411):
- https://bugs.openjdk.org/browse/JDK-8270380

In order to make the builds a bit simpler I would be in favor to just remove the `java.security.manager` argument. With that only the Java-17 tests would run with enabled security-manager and all others would have it disabled. And Java-17 will probably not last long any-ways.

Btw. I have replayed a I-build test with explicit disallowed security-manager in order to test in advance if everything falls apart or not (you can see it in the diff):
- https://ci.eclipse.org/releng/job/AutomatedTests/job/ep435I-unit-linux-x86_64-java21/25/

PR #2673, opened 2024-12-17
Remove -Djava.security.manager=allow from the products

merks:
- Define an unconfigure touchpoint to remove -Djava.security.manager=allow from the eclipse.ini which is done to ensure that updates will leave installations in a state that can run with Java 24.
- Do the above for both products and ensure that both products specify -Dosgi.requiredJavaVersion=21

https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/issues/2623

### DIFF
--- a/eclipse.platform.releng.tychoeclipsebuilder/eclipse.platform.repository/platform.p2.inf
+++ b/eclipse.platform.releng.tychoeclipsebuilder/eclipse.platform.repository/platform.p2.inf
@@ -1,5 +1,6 @@
 instructions.configure=\
-  mkdir(path:${installFolder}/dropins);
+  mkdir(path:${installFolder}/dropins);\
+  removeJvmArg(jvmArg:-Djava.security.manager=allow);
 
 # Restrict range so we are not an automatic update for 3.x.
 update.id = org.eclipse.platform.ide
--- a/eclipse.platform.releng.tychoeclipsebuilder/eclipse.platform.repository/platform.product
+++ b/eclipse.platform.releng.tychoeclipsebuilder/eclipse.platform.repository/platform.product
@@ -9,7 +9,7 @@
    <launcherArgs>
       <programArgs>--launcher.defaultAction openFile --launcher.appendVmargs
       </programArgs>
-      <vmArgs>-Dosgi.requiredJavaVersion=17 -Dosgi.dataAreaRequiresExplicitInit=true -Dorg.eclipse.swt.graphics.Resource.reportNonDisposed=true --add-modules=ALL-SYSTEM -Dorg.slf4j.simpleLogger.defaultLogLevel=off -Djava.security.manager=allow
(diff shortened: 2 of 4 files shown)


merks:
This approach is an alternative to just removing the `-Djava.security.manager=allow` from the products which would leave existing updated installations forever with that option in their eclipse.ini, eventually failing to start with Java 24.  It's also an alternative to setting `-Djava.security.manager=disallow` in the product which might cause problems immediately (though ones that will definitely happen with Java 24 eventually), and that being said, the default for Java 21 is `dissallow` so setting it to that explicitly would maybe be fine too.  I'm not sure how others  @HannesWell @laeubi @akurtakov @jonahgraham  @iloveeclipse feel about this?

An `unconfigure` is used because the existing product IUs' p2 metadata looks like this:
```
<instruction key="configure">
...;addJvmArg(jvmArg:-Djava.security.manager=allow);...
</instruction>
<instruction key="unconfigure">
...;removeJvmArg(jvmArg:-Djava.security.manager=allow);...
</instruction>
```

akurtakov (approved):
It looks like the correct thing to do (removing setting to default value that we know will break soon )

merks:
I'll leave this open for 24 hours for anyone else to comment.

laeubi (approved):
I'm fine with that even though I probably would not have came up with such an sophisticated approach :-)

merks:
Thanks @laeubi  for pointing out that there were two separate `instructions.configure` properties!

merks:
Darn, now there are two commits.  I will force push again...

merks:
@akurtakov 

An earlier build passed all the tests but now the license Java 21 issue appears to be back.  Do you have idea of why that might be?

akurtakov:
I rerun it (just in case :) and it failed again. Looking into it.

HannesWell (approved):
Thank you for working out the details.
Removing existing `java.security.manager=allow` is probably the best we can do, because not everyone can consider changed defaults:
https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/pull/2659#issuecomment-2543862374

jonahgraham (approved):
I'll pick up the same changes for the EPP products when I do the following for 2025-03 M1 https://github.com/eclipse-packaging/packages/issues/251

- Synchronize the following - Remember to check branch, these links are to master, but around RC2 master may be setup for next release already
-  Synchronize any changes to [platform.product](https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/blob/master/eclipse.platform.releng.tychoeclipsebuilder/eclipse.platform.repository/platform.product) into all the epp.product files.
-  Synchronize any changes to [platform.p2.inf](https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/blob/master/eclipse.platform.releng.tychoeclipsebuilder/eclipse.platform.repository/platform.p2.inf) into all the *.product/p2.inf files.

merks:
Thanks everyone for the feedback!  💞

PR #1941, opened 2024-03-26
Add java security properties for surefire

laeubi:

PR #2677, opened 2024-12-17
Revert "Add java security properties for surefire"

HannesWell:
Now that all tests can handle a disabled SecurityManger it doesn't have to be enabled by default anymore.

This reverts https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/pull/1941


akurtakov (approved):
It's best to figure possible issues now.

Issue #2642, opened 2024-12-06
Fix unstable Centos-9 and OpenSUSE leap smoke tests


HannesWell:
At the moment the smoke tests running on CentOS-9 and Open SUSE leap are unstable and often fail or time out:
https://ci.eclipse.org/releng/job/SmokeTests/job/Start-smoke-tests/

This issue is an umbrella to investigate and fix this.
The first thing to notice is that when running for example the smaller `teamcore` test-suite all smoke tests on all platforms succeeded.

Therefore I assume the failure is related to the UI, since by default smoke tests run the `ui` suite.

## Community

- [x] I understand reporting an issue to this OSS project does not mandate anyone to fix it. Other contributors may consider the issue, or not, at their own convenience. The most efficient way to get it fixed is that I fix it myself and contribute it back as a good quality patch to the project.

HannesWell:
A pattern that I see regularly on CentOS-9 and OpenSUSE machines is:
```
Executing command: "pkill" "Xvnc" 
exit
sh: line 31: pkill: command not found
```

Therefore I wonder if installing procps could solve the failures/timeouts? As suggested in:
- https://www.thegeekdiary.com/pkill-command-not-found/

Can anyone with more Linux experience assess this?

akurtakov:
It will probably remove the warning/error from the log but I doubt it will actually improve the test results (it's still good to see less false positives so please do it) as I don't see pkill used in our scripts at all.

HannesWell:
> It will probably remove the warning/error from the log but I doubt it will actually improve the test results (it's still good to see less false positives so please do it) as I don't see pkill used in our scripts at all.

In the latest [smoke-test executions](https://ci.eclipse.org/releng/job/SmokeTests/job/Start-smoke-tests/), i.e. since https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/pull/2678 is available the CentOS and OpenSuse tests seem to be much more stable. The OpenSuse tests seem to have stabilized before a bit, but CentOS seem to succeeded only since that change.

Initially my theory is that without pkill the xvnc session somehow survived between the multiple builds on the same platform with different Java-versions. But since each build runs in a docker-container each run should actually be fully independent.
Or can xvnc somehow leak through the host machine?

Besides the good news about OpenSuse and CentOS we have also bad news that the linux.riscv tests seem to be unstable too. maybe something is missing on that machine too.
Will look into them later.

[...]

PR #2678, opened 2024-12-17
Install git and `procps` package in all docker-images

HannesWell:
Install git on all docker-images, which is a prerequisite for
- https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/pull/2609

Install `procps` package in all docker-images for smoke-tests. The 'procps'-package contains the pkill command which is currently
missing according to the smoke-tests logs. Part of
- https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/issues/2642

### DIFF
--- a/cje-production/dockerfiles/centos-gtk4-mutter/9-gtk4/Dockerfile
+++ b/cje-production/dockerfiles/centos-gtk4-mutter/9-gtk4/Dockerfile
@@ -17,6 +17,7 @@ RUN dnf -y update && dnf -y install \
       tigervnc \
       mutter \
       xorg-x11-server-utils \
+      procps-ng \
       mesa-libGL \
       xorg-x11-fonts-misc \
       xorg-x11-fonts-75dpi \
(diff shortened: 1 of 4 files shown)


cje-production/dockerfiles/centos-gtk4-mutter/9-gtk4/Dockerfile
```diff
@@ -17,6 +17,7 @@ RUN dnf -y update && dnf -y install \
       tigervnc \
       mutter \
       xorg-x11-server-utils \
+      procps \
```
HannesWell, line 20: For CentOS there is also `procps-ng` available. Can anyone tell if `procps-ng` or `procps` is better to get the `pkill` command?

For open-suse only `procps` seems available.

akurtakov: On my fedora system pkill comes from procps-ng thus I assume that's the preferred one on Red Hat family of distros.

HannesWell: Thanks for checking. I changed this PR to install `procps-ng` in CentOS and `procps` into the Suse and Debian docker image.

PR #2680, opened 2024-12-18
Switch Mac I-build tests to Java 21

akurtakov:
Fixes https://github.com/eclipse-platform/eclipse.platform/issues/1665


akurtakov:
Used only by I-builds thus merging without waiting.

PR #2684, opened 2024-12-18
[Build] Simplify deployment of RelEng scripts in master branch build

HannesWell:
I have already verified that the result is the same.


HannesWell:
The verification build does not test this, thus merging it immediately.

PR #2667, opened 2024-12-16
Run windows I-build tests with Java 21

akurtakov:

PR #2685, opened 2024-12-19
[Build] Adapt initial testConfigs.php to move of test-builds to Java-21

HannesWell:
Follow-up on
- https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/pull/2667
- https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/pull/2666
- https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/pull/2670
- https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/pull/2680

### DIFF
--- a/eclipse.platform.releng.tychoeclipsebuilder/eclipse/publishingFiles/staticDropFiles/testConfigs.php
+++ b/eclipse.platform.releng.tychoeclipsebuilder/eclipse/publishingFiles/staticDropFiles/testConfigs.php
@@ -10,9 +10,8 @@
 include("buildproperties.php");
 
 $expectedTestConfigs = array();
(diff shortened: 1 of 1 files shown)


akurtakov:
Looking at https://download.eclipse.org/eclipse/downloads/drops4/I20241218-1800/ and previous I question what are these expected configs needed for even.

HannesWell:
> Looking at https://download.eclipse.org/eclipse/downloads/drops4/I20241218-1800/ and previous I question what are these expected configs needed for even.

Initially I assumed the expected configs are necessary to display the complete list of test-configurations that are executed before the tests have completed. But after looking more into the details, I think these files are actually irrelevant since they are already re-generated respectively overwritten during each I-build before tests are even kicked off. Therefore I started an attempt to remove them in:
- https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/pull/2699

PR #2689, opened 2024-12-19
[Build] Define max heap size for Maven only via environment variables

HannesWell:
This avoids confusion which value is actually used and allows to reduce the memory demand of the container used for I-builds.
Defining it in the corresponding pipeline allows fine-tuning for the circumstances of the corresponding build and it's settings.

I'll have to verify in an experiment that we can really reduce the memory-limit of the container used for I-builds.

### DIFF
--- a/.mvn/jvm.config
+++ b/.mvn/jvm.config
@@ -1,2 +1,1 @@
 -XX:+PrintFlagsFinal
--Xmx3G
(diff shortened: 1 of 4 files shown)


HannesWell:
I have now done some testing and it looks like an XMX size of 4Gi is sufficient for the I-build and also allows to reduce the overall size of the executing kubernetes pod.

PR #2691, opened 2024-12-19
[Y-Build] Move Y-build and its tests on Java-21 and above

HannesWell:
Synchronizes the Y-build with all the changes recently applied to the I-build.


HannesWell:
The verification build doesn't test this, thus submitting immediately.

PR #2693, opened 2024-12-19
[Build] Unify declarations of list of executed I-/Y-build tests

HannesWell:
This reduces the number of places to adapt in case the I-/Y-build tests change and reduces the difference between I-builds and Y-builds. Furthermore it reduces duplication introduced to handle the small difference between I- and Y-builds.

I have to run a few more tests first, before I can submit this with enough confidence. Therefore this is currently a draft.


akurtakov:
This one looks quite promissing.

HannesWell:
With https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/pull/2699 being successful this can become simpler since now only one list of test-configs has to be maintained for the result website.
But at the same time this has to be enhanced to be able to pass the list of (expected) test-configs down to the `Collect Result` Jenkins job. Currently I think this can be solved by recording the list e.g. into the `buildproperties.properties/shsource` file.

HannesWell:
> But at the same time this has to be enhanced to be able to pass the list of (expected) test-configs down to the `Collect Result` Jenkins job. Currently I think this can be solved by recording the list e.g. into the `buildproperties.properties/shsource` file.

This should work now. But the verification build does not test this, therefore I submit this immediately and start an I-build to verify it works as expected.

HannesWell:
The initial creation of the list of tested configurations looks good. 
Awaiting the updates after the first tests have completed.

HannesWell:
On the other hand, updating the results after a successful test failed because the  `TESTS_CONFIGURATIONS_EXPECTED` property somehow didn't made it into the generated build properties.shsource/properties files. 
I'll investigate it.

PR #2688, opened 2024-12-19
[Build] Unify tool-installation and PGP-settings in I-/Y-builds

HannesWell:
Part of https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/issues/2625

PR #2696, opened 2024-12-20
[Build] Stop explicit usage of the 'jnlp' container in I-/Y-builds

HannesWell:
The 'jnlp' container in a Kubernetes pod is the one used by default and it's therefore not necessary to specify it explicitly.

Furthermore using it in combination with a top-level declarative Maven-tool breaks the usage of SSH-agents. For more details see: https://gitlab.eclipse.org/eclipsefdn/helpdesk/-/issues/4171

Fixes the problem mentioned in https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/pull/2688/files#r1893297729.

### DIFF
--- a/JenkinsJobs/Builds/I_build.groovy
+++ b/JenkinsJobs/Builds/I_build.groovy
@@ -84,14 +84,11 @@ spec:
   stages {
       stage('Clean Workspace'){
           steps {
-              container('jnlp') {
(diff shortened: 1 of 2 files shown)


HannesWell:
The verification build doesn't test this, thus submitting immediately.

PR #2697, opened 2024-12-20
[Build] Unify performance-test jobs into one Jenkins matrix pipeline

HannesWell:
This reduces the number of jobs from four to one and therefore cleans-up the `RelEng` Jenkins instance and the Jenkins-Jobs defined in this repository further.

Performance tests are not executed anymore as per
- https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/discussions/828
- https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/discussions/2264

Nevertheless I wasn't sure that they should be fully deleted as this also allows respectively would mean we should delete all the other infrastructure in test scripts etc. Therefore I have cleaned up at least the job definitions as far as possible.

While testing the new job I noticed two things, that I would leave as an exercise in case performance tests are re-activated one day.

1. The `performance` agent [zyt5z-centos83](https://ci.eclipse.org/releng/computer/zyt5z%2Dcentos83/) cannot run the latest I-build anymore, very likely because it's has CentOS-8 installed. So that would probably have to be updated or performance tests should run on another machine.
2. I wasn't able to understand or find any difference between the `I-linux.x86_64` and `ILR-linux.x86_64` job respectivly between their baseline counterparts. Their job definitions looked very much the same. So having both is probably obsolete.

For both points I have left a `FIXME`/`TODO` comment in the pipeline.


HannesWell:
The verification build does not test this and I already tested this change, as far as possible, in a job I created manually in the `RelEng` Jenkins. Submitting.

PR #2698, opened 2024-12-22
[Build] Fix condition of maven-snapshot publication step

HannesWell:
Current the 'COMPARATOR_ERRORS_SUBJECT' environment variable always contains a line-break, even if otherwise 'empty' (indicating no comparator-error). Trimming the value in the condition removes any leading/trailing white-space.

Follow-up on:
- https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/pull/2538

@merks, @akurtakov IIRC we had issues due to not deployed Maven-snapshots (of ECJ)? This was probably the cause because since the changed mentioned above snapshots were not published anymore.

### DIFF
--- a/JenkinsJobs/Builds/I_build.groovy
+++ b/JenkinsJobs/Builds/I_build.groovy
@@ -368,7 +368,7 @@ spec:
 		}
 		stage('Trigger publication to Maven snapshots repo') {
 			when {
-				environment name: 'COMPARATOR_ERRORS_SUBJECT', value: ''
+				expression { env.COMPARATOR_ERRORS_SUBJECT.trim().isEmpty() }
 				// On comparator-erros, skip the deployment of snapshot version to the 'eclipse-snapshots' maven repository to prevent that ECJ snapshot
 				// from being used in verification builds. Similar to how the p2-repository is not added to the I-build composite in that case.
 			}


HannesWell:
The verification build does not test this, submitting immediately.

merks:
The publishing stopped happening because someone configured the cron timing such that this job no longer started:

https://ci.eclipse.org/releng/view/Publish%20to%20Maven/job/CBIaggregator/

HannesWell:
Yes, but actually each I-build triggers a publication of Maven-snapshots for two years already.
So I wonder if the cron-trigger was 'disabled' on purpose?

Actually looking at the Job-Definition in this repo there is no trigger defined at all:
https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/blob/5a7acfe47ed0d6ad9407f81c1b4842c02d369b53/JenkinsJobs/Releng/CBIaggregator.groovy
With results to: https://ci.eclipse.org/releng/job/Releng/job/CBIaggregator/

On the other hand that's not actually the job called by the I-Build... That's :
- https://ci.eclipse.org/releng/view/Publish%20to%20Maven/job/CBIaggregator/

(which is also the job linked in the `Publish to Maven` view respectivly the one you linked):
- https://ci.eclipse.org/releng/view/Publish%20to%20Maven/job/CBIaggregator/

It looks like something should be cleaned up here too... I'll look into that.

PR #2699, opened 2024-12-22
[Build] Remove testConfigs.php file that is generated in I-/Y-builds

HannesWell:
The 'testConfigs.php' is regenerated respectively immediately overwritten during an I-/Y-build in the 'Gather Eclipse Parts' stage:
https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/blob/26fe77d38f5426fbde6221db71d97cd47db480f5/JenkinsJobs/Builds/I_build.groovy#L238
Which calls via the 'gatherEclipseParts.sh' script the 'publish' target in the 'org.eclipse.platform.releng.tychoeclipsebuilder/eclipse/helper.xml' ANT-script:
https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/blob/26fe77d38f5426fbde6221db71d97cd47db480f5/cje-production/mbscripts/mb300_gatherEclipseParts.sh#L242-L265
Within that the 'generateIndex' target of the 'cje-production/scripts/publish.xml' ANT-script is called:
https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/blob/26fe77d38f5426fbde6221db71d97cd47db480f5/eclipse.platform.releng.tychoeclipsebuilder/eclipse/helper.xml#L125
Eventually this calls
org.eclipse.releng.generators.TestResultsGenerator.getTestsConfig(), which regenerates the 'testConfigs.php':
https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/blob/26fe77d38f5426fbde6221db71d97cd47db480f5/cje-production/scripts/publish.xml#L20-L25
https://github.com/eclipse-platform/eclipse.platform.releng.buildtools/blob/3d2e9c43409f4f36f2dd9b32c6875059d114d166/bundles/org.eclipse.build.tools/src/org/eclipse/releng/generators/TestResultsGenerator.java#L1320-L1344

And since 'testConfigs.php' is generated from the already existing list of ran test-configurations there is no need to maintain it manually again.


HannesWell:
Verification builds don't test this.
Therefore I'm submitting this immediately and start an I-build to verify everything still works as expected.

HannesWell:
Looks like this experiment was successful and the table of test configurations in the Summary of Unit Tests Results section still works as expected, even when no I-build test has completed yet.

PR #2704, opened 2024-12-27
[Build] Unify declaration of build-properties for I-, Y- and P-builds

HannesWell:
Instead of duplicating the `buildproperties.txt` _template_ for each build type, keep only variables with common values in the general `buildproperties.txt` and define differing variables in the Jenkins pipeline and add it dynamically to the generated/derived `buildproperties.shsource/properties`.

Remove unused `TESTED_BUILD_TYPE` variable.
Update `API_PREV_REF_LABEL` for Y- and P-build to be in sync with the value defined for I-builds.

Part of https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/issues/2625.


HannesWell:
The verification build does not test this, thus I'm submitting this immediately.

@MohananRahul I have also updated the RelEng documentation for the creation of new Y-/P-builds. Please let me know if anything is unclear or missing.

PR #2708, opened 2024-12-29
Switch javadoc source to 21

akurtakov:
Now that there is Java 21 content this would allow such code to be accepted.
Additionally, maven-javadoc-plugin uses that to decide which Java version to link javadoc to aka with this change String will link to Java 21 and not Java 17.


akurtakov:
Merging and starting new I-build to ensure no breakage.

PR #2711, opened 2024-12-30
[Build] Simplify and clean-up I/Y-build Jenkins pipeline

HannesWell:
- Declare ANT as top-level tool instead of multiple `withAnt()` steps
- Use dir-step and leverage `CJE_ROOT` environment variable to change the working directory
- Use `readProperties()` step to read build-properties and trim them immediately after reading
- Remove duplicated environment variable exports
- Replace duplicated `BUILD_VERSION` env-variable by `RELEASE_VER`
- Remove unused environment variables `POM_UPDATES_SUBJECT` and `POM_UPDATES_BODY`, which are left-overs of Bug 578179 respectively commit 6a93348887705263333e7748a8c3f5f3cce3d82a
- Send plain-text mails instead of HTML content to simply the creation of the mail body.
- Simplify references to variables in Groovy string-templates.
- Remove unnecessary initial `cleanWs()` step. The build runs inside a kubernetes-container that is always empty.

### DIFF
--- a/JenkinsJobs/Builds/build.jenkinsfile
+++ b/JenkinsJobs/Builds/build.jenkinsfile
@@ -58,26 +58,22 @@ spec:
 """
     }
   }
-  tools {
-      jdk 'temurin-jdk21-latest'
-      maven 'apache-maven-latest'
-  }
-  environment {
+	tools {
+		jdk 'temurin-jdk21-latest'
+		maven 'apache-maven-latest'
+		ant 'apache-ant-latest'
+	}
+	environment {
 		BUILD_TYPE = "${BUILD.type}"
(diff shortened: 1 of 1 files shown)


HannesWell:
The verification build does not test this, but I have started an I-build where I applied the proposed changes temporarily using Jenkins `Replay`:
https://ci.eclipse.org/releng/job/Builds/job/I-build-4.35/69/

HannesWell:
> The verification build does not test this, but I have started an I-build where I applied the proposed changes temporarily using Jenkins `Replay`: https://ci.eclipse.org/releng/job/Builds/job/I-build-4.35/69/

That build was successful, submitting.

PR #2709, opened 2024-12-29
[Build] Unify I-build and Y-build into one common Jenkinsfile

HannesWell:
Contributes to
- https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/issues/1950
- https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/issues/2625

PR #2713, opened 2024-12-30
[Build] Use default declarative checkout for I/Y-builds

HannesWell:
and restore ability to define I/Y-builds jobs for other branches, only through the job definition.
This ability was lost for I-builds in https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/pull/2709. For maintenance branches the `build.jenkinsfile` would have to be modified directly, like it is done for Y- and P-builds.

Furthermore use the error() step instead of throwing an 'IllegalStateException' because the latter doesn't always seem to be permitted.


HannesWell:
The verification build does not test this, submitting immediately.

PR #2714, opened 2024-12-30
[Build] Fix Y-Build test result collection

HannesWell:
Don't use xvnc, which is not available on the 'basic' agent and also not necessary (for a headless build task) [1].
Use The unified  'cje-production/scripts/publish.xml' ANT script. The script customized for Y-builds is not necessary anymore since [2].
Follow-up for
- https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/pull/2611 [1]
- https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/pull/2693 [2]


HannesWell:
The verification build doesn't test this, but I have already verified that this works by applying the changes manually to the job definition, thus submitting.

PR #2718, opened 2025-01-02
[Build] Update, clean-up and use job for Maven publication

HannesWell:
Currently the publication of the Maven artifacts of Eclipse happens in five jobs manually/directly defined in the 'Releng' Jenkins instance:
1. CBIaggregator
2. RepositoryValidator
3. PublishPlatformToMaven, PublishJDTtoMaven and PublishPDEToMaven

Corresponding job-definitions for these jobs already in this SCM (except for the repository-validation), but the jobs generated from them are not used.

This change replaces the existing (unused) multiple job-definitions by a single Jenkins pipeline that contains all steps and performs the final publication for all covered projects in a single matrix. The new pipeline is updated to incorporate all the changes made in the meantime to the manually maintained publication jobs.

The I-build job is changed to use the new publication job.

With respect to the existing, manually maintained publication job, the new job is adjusted in the following aspects:

- Rename job to just 'PublishToMaven'. The CBI part by itself is irrelevant and a tool in the publication process
- Remove documentation about the jobs involved and scripts called from the job description as it's an implementation detail that can be looked up in the pipeline if of interest.
- Don't require the 'projects-storage.eclipse.org-bot-ssh' SSH agent for the aggregation step.
- Only run publication on explicit invocation (from I-builds). Remove the timer-trigger: Without changes a (re-)publication is not necessary.

With this change following Jenkins jobs can be deleted:
- manually maintained (in the JIPP root):
  - https://ci.eclipse.org/releng/job/CBIaggregator/
  - https://ci.eclipse.org/releng/job/RepositoryValidator/
  - https://ci.eclipse.org/releng/job/PublishPDEToMaven/
  - https://ci.eclipse.org/releng/job/PublishJDTtoMaven/
  - https://ci.eclipse.org/releng/job/PublishPDEtoMaven/
- with definition in this git-repo (unused and located in the `Releng` folder):
  - https://ci.eclipse.org/releng/job/Releng/job/CBIaggregator/
  - https://ci.eclipse.org/releng/job/Releng/job/PublishPlatformToMaven/
  - https://ci.eclipse.org/releng/job/Releng/job/PublishPDEToMaven/
  - https://ci.eclipse.org/releng/job/Releng/job/PublishJDTtoMaven/

And one new jobs will be added:
- https://ci.eclipse.org/releng/job/Releng/job/PublishToMaven/

PR #2719, opened 2025-01-02
[Build] Fix invocation of 'PublishToMaven' job from I-build job

HannesWell:
Restore the previous (correct) definition of the 'snapshotOrRelease' parameter.

Follow-up on
- https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/pull/2718


HannesWell:
Not checked by verification build, submitting immediately.

PR #2724, opened 2025-01-04
[Build] Simplify verification of Maven artifacts to publish

HannesWell:
- Make scripts easier to understand for Java-programms that aren't UNIX-shell/bash experts.
- Separate the eclipse.platform.releng.aggregator repository clone from the working directory into a separate folder.

I'm also contemplating to move the validation into the matrix to avoid defining the list of projects twice. Performance-wise this doesn't make a difference but further simplifies the configuration a bit.
The only downside is that, if the validation only fails for one project (e.g. jdt) then the other projects would likely still be published. On the other hand if there is a validation error it it's probably systematic and the validation probably fails for all projects.

### DIFF
--- a/JenkinsJobs/Releng/publishToMaven.jenkinsfile
+++ b/JenkinsJobs/Releng/publishToMaven.jenkinsfile
@@ -4,10 +4,11 @@ pipeline {
 		timestamps()
 		timeout(time: 120, unit: 'MINUTES')
 		buildDiscarder(logRotator(numToKeepStr:'10'))
+		checkoutToSubdirectory('git-repo')
 	}
 	agent {
 		label 'basic'
-	}	
+	}
 	tools {
 		jdk 'temurin-jdk21-latest'
 		maven 'apache-maven-latest'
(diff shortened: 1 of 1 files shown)


JenkinsJobs/Releng/publishToMaven.jenkinsfile
```diff
@@ -20,50 +21,43 @@ pipeline {
...
 			steps {
-				dir('repo-validation') { // Do the work in a clean folder without a pom.xml
-					sh '''#!/bin/bash
-						workingDir=$(pwd)
-						pushd "${REPO}"
-						# Find all the version folders for pde, jdt, and platform.
-						# Filter out all the feature, test, and product IUs that are not published.
-						# And transform each path to a Maven artifact coordinate groupId:artifactId:version.
-						find org/eclipse/pde org/eclipse/jdt/ org/eclipse/platform/ -regextype posix-egrep -regex '.*/[0-9]+\\.[0-9]+[^/]*' \\
-							| grep -vE "\\.feature\\.group|\\.feature\\.jar|\\.executable|\\.test|\\.platform\\.ide|\\.platform\\.sdk|_root|\\.id/|\\.sdk\\.ide/" \\
```
HannesWell, line 41: @merks can you tell if filtering the list of artifacts is still necessary here?
To me it looks like the repo assembled by the CBI-aggregator already only contains the artifacts to publish in the
- `org/eclipse/platform`
- `org/eclipse/jdt`
- `org/eclipse/pde`

sub-folders. When performing the actual publication, all artifacts in these folders are published too and nothing is filtered.

merks: Yes I’m quite sure the filtering is necessary to remove tests, features, and products as the comment suggests.  If course I’m reading on an iPhone so maybe context is missing.

HannesWell: Ok, my confusion comes from the fact that currently tests, features, etc. are already removed in the `CBIaggregator.sh` script, e.g. in
https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/blob/d163455499ab4f07ac03949df02dfe387f1fc081/eclipse.platform.releng/publish-to-maven-central/CBIaggregator.sh#L143

Therefore I didn't found any test, feature etc. in the aggregated repo, but if just running the CBI-aggregator they are still there.
So at some point, filtering is necessary.

I wonder if it would be possible to define exclusions already in the CBI-aggregator model as described in (assuming it's still possible):
https://wiki.eclipse.org/CBI/aggregator/manual#Exclusion_Rule

But I assume that has drawbacks as well?

merks: Probably nothing was added that wasn't needed at the time, but maybe some things have changed...

With a large enough list of the exclusions, a list that might need to grow as the content changes, one could get the aggregation to exclude them earlier in the process.   Or the aggregator could be enhanced to implement the pattern there.

HannesWell: > With a large enough list of the exclusions, a list that might need to grow as the content changes, one could get the aggregation to exclude them earlier in the process. Or the aggregator could be enhanced to implement the pattern there.

Yes. It looks like the CBI-Aggregator already has the concept of Exclusion rules, but I failed to apply them through the UI:
- https://wiki.eclipse.org/CBI/aggregator/manual#Exclusion_Rule

But according to that doc, their use is limited to one IU. It would be of course convenient to exclude e.g. features and products by their type etc.

But in general I wonder if the CBI aggregator is the right place to put effort into or if the time would be invested better in testing and if necessary enhancing the poms generated by Tycho pomless. IIRC that's achievable with:
- https://tycho.eclipseprojects.io/doc/4.0.10/tycho-packaging-plugin/update-consumer-pom-mojo.html

But for this I keep the filtering as it is. Thanks for your help.

merks: Anyone can invest effort wherever they like.  One tends to find out later that it's a more complex problem than meets the eye.  E.g., mapping dependencies onto dependency coordinates that really exist in maven central, mapping package imports onto the coordinates that actually provide that package, blah, blah, blah.   While one person might say, if it isn't broken don't fix it, the other might says, lets re-engineer that and make it better.  😜

HannesWell: Sure, I already assumed it wouldn't be trivial. But I have to admit that haven't looked into how well this is already working in Tycho. I was just thinking about where, ideally, the journey should go. :)

Issue #2725, opened 2025-01-04
Remove org.eclipse.ecf.ssl fragment from the Eclipse IDE


scottslewis:
As per the investigation described in comments on this issue:

https://github.com/eclipse/ecf/issues/140

the org.eclipse.ecf.ssl fragment should be removed from the Eclipse IDE.

merks:
You mention just the fragment `org.eclipse.ecf.ssl` but there's there's also the `org.eclipse.ecf.core.ssl.feature` which includes that fragment so presumably that must go too, right?  With the only plugin that it includes removed it would be pointless  That feature in turn is included by `org.eclipse.ecf.filetransfer.ssl.feature` so presumably with the previous bundle and feature removed, the `org.eclipse.ecf.filetransfer.ssl.feature` would include only the single fragment `org.eclipse.ecf.provider.filetransfer.ssl`.  Is that being removed as well?  That bundle has just a single class org.eclipse.ecf.internal.provider.filetransfer.ssl.ECFURLConnectionModifier. In my test environment I tried running without the class ECFURLConnectionModifier and that worked fine.  Even when I tried running with that class present, ECFURLConnectionModifier.setSocketFactoryForConnection(URLConnection) was never called.

Given that the only ECF-specific subclass of javax.net.ssl.SSLSocketFactory that I see is org.eclipse.ecf.internal.ssl.ECFSSLSocketFactory the ECFURLConnectionModifier class is maybe worse than useless.

![image](https://github.com/user-attachments/assets/4282180e-22a7-45d9-84e5-73f91b26edbf)

A connection modifier appears to be used only from URLFileSystemBrowser.runRequest() but that class is used only when there is no more specific provider present for the protocol.  For http and https different implementation classes are used without using the modifier.

![image](https://github.com/user-attachments/assets/bfa96796-cfac-4623-b4f7-a2b404712d9e)

So I assume it is safe to assume that we should be removing references to all of the following:

- org.eclipse.ecf.ssl
- org.eclipse.ecf.provider.filetransfer.ssl
- org.eclipse.ecf.core.ssl.feature
- org.eclipse.ecf.filetransfer.ssl.feature

scottslewis:
Even when I tried running with that class present, ECFURLConnectionModifier.setSocketFactoryForConnection(URLConnection) was never called.
> 

[...]

PR #2726, opened 2025-01-05
Remove references to all of ECF's ssl bundles and features

merks:
https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/issues/2725


merks:
It looks like the license check is running against older binaries:

> Missing requirement: org.eclipse.equinox.p2.core.feature.feature.group 1.7.500.qualifier requires 'org.eclipse.equinox.p2.iu; org.eclipse.ecf.core.ssl.feature.feature.group [1.1.0,2.0.0)' but it could not be found

Given I've only removed dependencies here, we can safely ignore the license check...

PR #2727, opened 2025-01-05
[Build] Use Maven-deamon to speed up Maven artifact publication

HannesWell:
Using Maven-deamon reduces the runtime of the repository validation from ~14min to ~1.5min and the publication steps for the different projects from ~23min respectively ~5min to ~9min respectively ~2min.
Overall the runtime is reduced from ~42min to ~16min.

### DIFF
--- a/JenkinsJobs/Releng/publishToMaven.jenkinsfile
+++ b/JenkinsJobs/Releng/publishToMaven.jenkinsfile
@@ -15,6 +15,10 @@ pipeline {
 	}
 	environment {
 		REPO = "${WORKSPACE}/repo"
+		PATH = "${installMavenDaemon('1.0.2')}/bin:${PATH}"
+		// Folder ~/.m2 is not writable for builds, ensure mvnd metadata are written within the workspace.
+		// prevent jline warning about inability to create a system terminal and increase keep-alive timeouts to increase stability in concurrent usage
+		MVND = "mvnd -Dmvnd.daemonStorage=${WORKSPACE}/tools/mvnd -Dorg.jline.terminal.type=dumb -Dmvnd.keepAlive=1000 -Dmvnd.maxLostKeepAlive=100"
 	}
 	// parameters declared in the definition of the invoking job
 	stages {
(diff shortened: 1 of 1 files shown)


JenkinsJobs/Releng/publishToMaven.jenkinsfile
```diff
@@ -15,6 +15,9 @@ pipeline {
 	}
 	environment {
 		REPO = "${WORKSPACE}/repo"
+		PATH = "${installMavenDaemon('1.0.2')}/bin:${PATH}"
+		MVND_DAEMON_STORAGE = "${WORKSPACE}/tools/mvnd" // Folder ~/.m2 is not writable for builds, ensure mvnd metadata are written within the workspace.
+		TERM = 'dumb' // Prevent jline warning about inability to create a system terminal.
```
HannesWell, line 20: This is necessary due to
- https://github.com/apache/maven-mvnd/issues/1028

merks:
We really need a new I-build to fix the license checking...

laeubi:
Is this really mvn deamon or because we now store the m2 repository somewhere else?!? Seems a bit odd as the deamnon would be killed after the agent is stopped?

HannesWell:
> Is this really mvn deamon or because we now store the m2 repository somewhere else?!? Seems a bit odd as the deamnon would be killed after the agent is stopped?

I don't expect the maven-repo to be stored somewhere else. The gain is still explainable since we start maven twice (once for verification and once for the actual deployment) for each artifact to publish. And since we publish 296 artifacts it means maven is launched almost 600 times. And for all those invocations the daemon can be reused, even if it's killed after the build.

laeubi:
> The gain is still explainable since we start maven twice (once for verification and once for the actual deployment) for each artifact to publish.

ouch...

HannesWell:
> > The gain is still explainable since we start maven twice (once for verification and once for the actual deployment) for each artifact to publish.
> 
> ouch...

Yep. At least we save 10min of total build time for each second of runtime reduction of each build. And it looks like we save some seconds per build with this.

After increasing the keep-alive timeout a bit using the maven-daemon works now stable as desired. I assume that's necessary due to the parallel execution of the matrix on a single-core CPU, which I think makes some threads sometimes sleep for some time.

With that this is ready for submission to me.

PR #2729, opened 2025-01-05
[Build] Inline publishProject.sh and simplify Maven publication stage

HannesWell:
Simplify by
- Skipping the copying of the repository content
- Failing on first error in any command executed
- Enable printing of commands to the build console instead of capturing it in log files.


HannesWell:
The verification build doesn't test this but I have already tested this in a separate Jenkins job, thus submitting immediately.

PR #2613, opened 2024-11-30
Publish 4.34 to Maven central 

elsazac:
Publish to Maven central 4.34

Update SDK4Mvn.aggr and properties.sh file

Refs: https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/issues/2606

cc/ @MohananRahul

PR #2730, opened 2025-01-05
[Build] Clean-up and update maven deployment configuration

HannesWell:
Move the Maven plugin configuration to the `eclipse-platform-parent/pom.xml` and update the configured plugins.


eclipse-platform-parent/pom.xml
```diff
@@ -1130,6 +1130,45 @@
...
+              <groupId>org.sonatype.plugins</groupId>
+              <artifactId>nexus-staging-maven-plugin</artifactId>
+              <version>1.7.0</version>
+              <extensions>true</extensions>
+              <configuration>
+                <!-- The Base URL of Nexus instance where we want to stage -->
+                <nexusUrl>https://oss.sonatype.org/service/local/staging/deploy/maven2/</nexusUrl>
+                <!-- The server "id" element from settings to use authentication from -->
+                <serverId>ossrh</serverId>
+              </configuration>
```
HannesWell, line 1166: I wonder if should set the `skipStagingRepositoryClose` property here as described in
https://github.com/sonatype/nexus-maven-plugins/blob/f5ba90d339064c4981716a98e4afa6a76bc3ff2e/staging/maven-plugin/README.md#plugin-flags

But I also wonder in general why it has worked until now without it being set while each file project is deployed in an individual Maven build.
From https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/pull/2718#issuecomment-2568711862, @elsazac can you tell me where I can find the staging repository in case I ran a release publication to a staging repo as a test?

eclipse-platform-parent/pom.xml
```diff
@@ -1130,6 +1130,45 @@
             </plugins>
         </build>
     </profile>
+    <profile>
+      <id>maven-deploy</id>
+      <distributionManagement>
+        <repository>
+          <id>ossrh</id>
+          <url>https://oss.sonatype.org/service/local/staging/deploy/maven2/</url>
+        </repository>
```
HannesWell, line 1139: I also wonder if we have to specify the distribution repo here again, while we also have to specify it for the 
`nexus-staging-maven-plugin` below:
https://github.com/sonatype/nexus-maven-plugins/tree/main/staging/maven-plugin#configuring-the-plugin

That doc says the `serverId` is only used for authentication.

eclipse-platform-parent/pom.xml
elsazac: When I publish the artefacts for a release, I first run the CBIaggregator Job, which in turn triggers the PDE, JDT, and Platform. After these jobs successfully complete, the generated repositories can be seen under the `Staging Repositories` tab in the `Build Promotion section` in [Maven central repository](https://oss.sonatype.org/).

Screenshot 2025-01-06 at 1 29 06 PM

I’m trying to get the full context of the change and reviewing the details to understand it better. I’m not sure if this is exactly what you were asking me to clarify, so please let me know if there’s anything else you’d like me to address.

merks: The problem as I understand is is that what you see at https://oss.sonatype.org/#stagingProfiles depends on your credentials/access, so when I log in there I can see EMF's staging  content.

image

But we/I don't see the jdt/pde/platform repositories.

We believe that some -release builds have been done that have populated those 3 repositories but we can't see that and can't help manage that.  Probably best that content be dropped, but it seems maybe only you can do that.

---

Note that all the separate jobs have been combined into this single pipeline job:

https://ci.eclipse.org/releng/view/Publish%20to%20Maven/job/Releng/job/PublishToMaven/

All the other jobs will be deleted...

elsazac: > The problem as I understand is is that what you see at https://oss.sonatype.org/#stagingProfiles depends on your credentials/access, so when I log in there I can see EMF's staging content.

> But we/I don't see the jdt/pde/platform repositories.

Even for me, as I have access to the [Maven Central site](https://oss.sonatype.org/), I can see the three repositories in the staging area only, once the Jenkins jobs are successfully completed running with `-release` parameter.

Screenshot 2025-01-06 at 1 57 49 PM

HannesWell: Thank you Elsa for your help. That's interesting because as explained in https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/pull/2718#issuecomment-2567723904, I actually expected that (accidentally) some staging repositories were created due to the old configuration of the parameter default in combination with the cron trigger.
E.g. https://ci.eclipse.org/releng/job/CBIaggregator/1281/ ran with parameter `-release` and triggered https://ci.eclipse.org/releng/job/RepositoryValidator/1020/, which eventually triggered
- https://ci.eclipse.org/releng/job/PublishJDTtoMaven/1114/
- https://ci.eclipse.org/releng/job/PublishPDEToMaven/1109/
- https://ci.eclipse.org/releng/job/PublishPlatformToMaven/1149/

and that happened multiple times.
And the changes you are applying for releases to the `SDK4Mvn.aggr`, e.g. in https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/pull/2613, as far as I understand them only change the source.

Therefore I still would expect that the Jenkins job executions listed above should have created a staging repo.

But maybe I have to run the job with `-release` option and more print-out to see where the artifacts are actually going to.

HannesWell: > Even for me, as I have access to the [Maven Central site](https://oss.sonatype.org/), I can see the three repositories in the staging area only, once the Jenkins jobs are successfully completed running with `-release` parameter.

@elsazac can you please check it again. For testing I ran the testing job I use again with the `-release` parameter.

eclipse-platform-parent/pom.xml
HannesWell: Maybe the [gpg:sign-and-deploy-file
](https://maven.apache.org/plugins/maven-gpg-plugin/sign-and-deploy-file-mojo.html) goal doesn't use the usual means of the maven-deploy-plugin and therefore the interception by the `nexus-staging-maven-plugin` doesn't have any effect?

eclipse-platform-parent/pom.xml
elsazac: Now I can see the three repositories in the staging area - PDE JDT and Platform

HannesWell: Ok, great. Thank you! This means that this works. 🎉
Can you please **discard or drop** then so that they are **not!!!!** released, as described in 
https://central.sonatype.org/publish/release/#locate-and-examine-your-staging-repository

Or can you tell if it is possible to inspect the content without being logged in? I see in in the release process of other Maven projects that they provide the staging repository publicly to test release-candidates.

elsazac: Usually I do close all the repositories and then release. while releasing, I make sure to check the `Automatically Drop` checkbox;

In this case, Do you want me to just only drop ?  and not close or release it?

Screenshot 2025-01-06 at 5 54 43 PM

laeubi: Drop is the correct thing here if it was only for testing purpose.

HannesWell: > In this case, Do you want me to just only drop ? and **not close or release** it?

Yes please just **drop** and **do not close or release**!

But before doing it, it would be nice if you could check if these staging repos can be accessed publicly or can be made publicly accessible?

elsazac: I don't see a built-in option as such to make staging repositories in Maven Central publicly available.

These are the only options that's seen available for any work related to the staging repositories.

Screenshot 2025-01-06 at 6 13 33 PM

Is there anything else you’d like me to check before I proceed with the drop?

HannesWell:
After some more testing I'm now convinced that configuring `nexus-staging-maven-plugin` is not necessary because it's effectively unused as the `gpg:sign-and-deploy-file` performs the deployment directly. For the same reason specifying the `ossrh`-repository in the `distributionManagement` is not necessary.

Additionally it's also not necessary to specify the id of the gpg-key used for signing as there is only one key in the keyring anyways and it's therefore selected by default. I have simplified the configuration to remove that unnecessary explicitness as well.

All my extensive testing with these changes was successful and therefore I consider this ready for submission.

Thanks to everyone who helped with this.

eclipse-platform-parent/pom.xml
HannesWell: > Is there anything else you’d like me to check before I proceed with the drop?

No that's fine. Thank you for your help.
Please proceed with dropping these staging repos.

I have a few more changes in this area in the pipeline and once they are all done, I plan to run another release build for testing on a clean repo. It would be great if you could drop/discard that then as well, but I'll ping you when this is necessary.

PR #2732, opened 2025-01-07
[Build] Remove unused creation of missing sources in Maven publication

HannesWell:
Since https://github.com/eclipse-platform/eclipse.platform.releng/issues/126 respectively https://github.com/eclipse-platform/eclipse.platform.releng/pull/167 no missing source artifacts are created anymore so the associated part of the CBIaggregator.sh script can be cleaned-up.

@merks @akurtakov or @mickaelistria or do you think that this could be useful again one day?


akurtakov:
IMO Maven publication should not generate sources and if a missing source case arise it should be handled in build scripts and not in publishing.

merks:
Maybe this was used for ecj in the past...

I would hope we'd never need this again...

laeubi:
> IMO Maven publication should not generate sources and if a missing source case arise it should be handled in build scripts and not in publishing.

Just wanted to mention that sources are a requirement for maven-central publication, and the recommendation is to generate a fake artifact for the cases where no source can be given.

merks:
Yes, it's a requirement, unless you are Apache and then you can dump your open source binaries without your open source source.  😱

PR #2735, opened 2025-01-08
[Build] Simplify exclusion of artifacts during Maven publication and inline CBIaggregator script

HannesWell:
This PR enhances the workflow for publication to Maven in the following two steps:
1. Simplify exclusion of artifacts during Maven publication
  Instead of activly deleting not desired artifacts and collecting the resulting garbage by deleting all then unused dependencies, just skip the validation and deployment of undesired artifacts from the folders of the published projects/groups.
2. Inline CBIaggregator script into publishToMaven Jenkinsfile
 Inlining the script allows to integrate its commands tighter with the pipeline and to reuse the existing facilities to download required tools.
 Additionally simplify the configuration of the tool-setup by deriving the versions of the eclipse-tools from the `PREVIOUS_RELEASE` properties declared in the `buildproperties.properties` file of the latest I-build.

This is currently a draft since a few things have to be tested, but the general form is already ready.


HannesWell:
After a lot of testing and more refinements this is now ready for submission.

The verification build doesn't test this, therefore I submit this immediately.

PR #2736, opened 2025-01-09
Exclude commons-beanutils from deprecation report

akurtakov:
We should not report 3rd party deprecations as there is no API promise for them.
Fixes commons-beanutils appearing in the report after https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/commit/d9a1570180e013c7b4fe208a9ccc035cd27cf3e6


akurtakov:
Report generated only by I-builds thus no point in waiting for verifications.

PR #2737, opened 2025-01-09
[Build] Adapt to fixed credential-id of releng PGP-key passphrase

HannesWell:
The typo in the credential-id was discovered in https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/pull/2731#discussion_r1906164328


HannesWell:

Btw. the changed credential-id caused the latest Y-build failure but should be fixed with this.

PR #2742, opened 2025-01-12
[Build] Launch EnrichPom.java as source program in Maven publication

HannesWell:
This leverages the ability to launch multi-file source-code programs add in Java-22 via JEP-458:
- https://openjdk.org/jeps/458

Technically it's not a problem to use JDK-23 or 24 (as soon as it's released) in this internal part of the build, yet I still hesitate with this because I worry a bit that it might be forgotten to update the JDK tool as soon as the non-LTS releases are out of maintenance since we usually don't use them anywhere else and it's therefore not on our TODO-list.

### DIFF
--- a/JenkinsJobs/Releng/publishToMaven.jenkinsfile
+++ b/JenkinsJobs/Releng/publishToMaven.jenkinsfile
@@ -10,7 +10,7 @@ pipeline {
 		label 'basic'
 	}
 	tools {
-		jdk 'temurin-jdk21-latest'
+		jdk 'temurin-jdk23-latest'
 		maven 'apache-maven-latest'
 	}
 	environment {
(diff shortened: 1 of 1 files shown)


HannesWell:
> I still hesitate with this because I worry a bit that it might be forgotten to update the JDK tool as soon as the non-LTS releases are out of maintenance since we usually don't use them anywhere else and it's therefore not on our TODO-list.

Alternatively we could simply leave this PR open until JDK-25 is released in autumn of this year. Since this is just an internal RelEng usage we don't have to follow any rules regarding usage of new JDK versions.

akurtakov (approved):
There are a bunch of jobs that are still on 17 or even 11 even though 21 is required by Eclipse now (aka "forgotten")  and IMO we will always have such so having one on Java 23 doesn't out weight the improved readability of the job itself.

HannesWell:
> There are a bunch of jobs that are still on 17 or even 11 even though 21 is required by Eclipse now (aka "forgotten") and IMO we will always have such so having one on Java 23 doesn't out weight the improved readability of the job itself.

That's right. Ok, lets have this now and I try to remember this when JDK 24 and 24 are released.

PR #2750, opened 2025-01-15
[Build] Fix signing of the artifact-checksums list on download page

HannesWell:
The checksum signature is gone since:
https://archive.eclipse.org/eclipse/downloads/drops4/R-4.16-202006040540/

This is caused by a failure in the signing process because recent versions of GPG ask for the passphrase via a dialog by default. In order obtain it through a environment variable one has disable pinentry using: --pinentry-mode loopback

- Remove unnecessary specification of '--yes' option.
- Remove attempt to read 'bashUtilities.shsource' as it seems to have been gone in general.
- specify the 'client' through a environment variable in the caller


HannesWell:
The verification build doesn't test this. Submitting this immediately so that it can be picked up by tonight's I-build.

PR #2753, opened 2025-01-16
[I/Y/P-Build] Various minos clean-ups of I/Y/P-Builds

HannesWell:
1. Remove unnecessary 'unset' of empty environment variables
2. Remove use of obsolete SSH-agent for git.eclipse.org
3. Unify and fix artifact archiving and make it more specific
  The archiving step in case of build-failures resulted in no artifacts being archived.


HannesWell:
The verification build doesn't test this and therefore this is submitted immediately.

PR #2754, opened 2025-01-16
[I/Y-Build] Unify index.template.php site for I and Y builds

HannesWell:
The existing differences are not relevant and probably only due to omitted synchronization.


HannesWell:
The verification build doesn't test this and therefore I'll submit this immediately so that the next I- and Y-build can verify this.

PR #2755, opened 2025-01-16
[Build] Fix and unify invocations of 'update-download-index' job

HannesWell:
Update or fix all invocations of a 'update-download-index' job to invoke `Releng/updateIndex`, because its definition is under version-control. The other previously referenced jobs are either manually defined/maintained (with equal definition) or don't exist.

This fixes the index-update for the `Mark (Un)Stable` and `Make Visible` job and allows to delete the manually defined `eclipse.releng.updateIndex` job:
- https://ci.eclipse.org/releng/job/eclipse.releng.updateIndex/


HannesWell:
The verification build doesn't test this. Submitting this immediately so that the next I/Y-build can verify this.

PR #2758, opened 2025-01-18
Add a shared workflow to check dependency version ranges

laeubi:
This adds a new shared workflow to check for inconsistencies in version ranges based on byte code analysis.


laeubi:
Build failure is unrelated and I tested this with my p2 fork where it works as expected.

PR #2760, opened 2025-01-19
Automatically check licenses and request review for target updates

laeubi:
If there are updates to the target contents there is some chance we need a review.

This now adds a job that when a PR is created or updated calls the license check workflow together with a review request.


HannesWell (commented):
Calling the `maven-license-check-action` will trigger a review of all not vetted artifacts, that looks good.
But you wont get a comment with the pretty-printed results added to the PR created, which I think is especially useful if some artifacts are not vetted and one wants to check the requests status.

With the current dash-licenses workflow, this would require the platform-bot to add the usual `/request-license-review`. This seems a bit cumbersome on first sight, but eventually it's just another event-trigger, like adding a label or alike.

laeubi:
Yes this will be "silent" at the moment, but my idea is that if one sometimes later rerun the check it is already vetted (so it is actually good), maybe one can make it add a label to the PR... we are all getting already to much mails I think.

HannesWell:
> Yes this will be "silent" at the moment, but my idea is that if one sometimes later rerun the check it is already vetted (so it is actually good),

Yes, it's definitely better than before.

> maybe one can make it add a label to the PR (e.g `vetted` or `requires vetting`...) as we are all getting already to much mails I think.

Yes there are way to many mails, but luckily my mail-program sorts the mails into threads and deleting a two or three mail thread is not really a difference for me 😅. Nevertheless running the license-workflow if a label is present is probably best implemented by generalizing the case for 'automated requests for dependabot PRs' (which unfortunately doesn't work currently due to the permissions model):
https://github.com/eclipse-dash/dash-licenses/blob/9b1616198579426e36cc81f8fde2fffcc65d6103/.github/workflows/mavenLicenseCheck.yml#L61-L66

laeubi:
By label I mean, that this job can add a label to the PR when the check was successful, so one can "see" it... but thinking further now one better would want to rerun the license check check if vetting was performed... but this will require a lot more work I fear.

HannesWell:
> By label I mean, that this job can add a label to the PR when the check was successful, so one can "see" it... but thinking further now one better would want to rerun the license check check if vetting was performed... but this will require a lot more work I fear.

That would be nice, if possible I would even add a comment to ping the the person responsible for the PR, but as you said that's much more work as it requires a connection from the EF gitlab to the corresponding GH PR. But first of all it would require https://github.com/eclipse-dash/dash-licenses/issues/184, because currently there is no direct link from a Gitlab IP issue to the creating PR.

akurtakov:
Is there a reason to not push this one now? Silent request is still better than manual one IMO.

laeubi:
From my side we can try it out ...

laeubi:
@akurtakov but We would need a new parameter and a condition check here as it is now a shared workflow.

PR #2763, opened 2025-01-20
[Build] Unify marking and update-index jobs and convert to pipeline

HannesWell:
Furthermore derive the issue label from the 'issueURLissueNum and remove the 'issueNum' parameter.
Update the JDK to perform the composite-repository modification to Java-21.


HannesWell:
After testing this in a separate job at the Releng Jenkins instance, this is now now ready for submission.
The verification build doesn't test this and therefore I submit this immediately.

akurtakov:
Thanks Hannes!!! What you have done  for streamlining the build process is huge!

HannesWell:
Thanks and you are welcome 😃

For the record, because the type of the affected Jenkins jobs is changed from `job` to `pipelineJob` I had to delete the existing jobs so that the Seed-job could regenerate them. The main effect is that the build-number is reset to start again at zero for the following jobs (but this is not really a problem):
- https://ci.eclipse.org/releng/job/Releng/job/updateIndex/
- https://ci.eclipse.org/releng/job/Builds/job/markUnstable/
- https://ci.eclipse.org/releng/job/Builds/job/markStable/

PR #2765, opened 2025-01-20
[I/Y-Build] Convert comparator-error message in mail to plain-text too

HannesWell:
And simplify the creation of the comparator-error message in the result mail send to the mailing-list.
Previously the message was still html-code.

Follow-up on
- https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/pull/2711


HannesWell:

PR #2767, opened 2025-01-21
Use java 21 for running the dependency checks

laeubi:


laeubi:
Build will not test this...

PR #2768, opened 2025-01-21
[P-Build] Avoid modified copies of build scripts for P-builds

HannesWell:
Contributes to:
- https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/issues/2625


HannesWell:
The verification build doesn't test this. Submitting this immediately so tonight's I-build can verify this.

PR #2769, opened 2025-01-21
[I/Y-build] Set 'pipefail' option instead of checking PIPESTATUS

HannesWell:


HannesWell:
The verification build doesn't test this, thus submitting this immediately that tonight's I-build can fully verify this.

PR #2771, opened 2025-01-22
Skip PR creation if no report was produced

laeubi:
Currently the job fails for bundles that do not generate a report, as this is not an error, we skip the PR creation if the step did not created any report.


laeubi:
There is no good way than try this out other than submit, so CI build can not verify it as well.

PR #2778, opened 2025-01-24
Exclude Apache Httpclient 4 from deprecation reports

akurtakov:
Cleans up the report which is today
(https://download.eclipse.org/eclipse/downloads/drops4/I20250123-1800/apitools/deprecation/apideprecation.html ) mostly deprecations from this project.
Report should contain only bundles developed by the project.


akurtakov:
Report generated only by I-builds so merging directly.

PR #2781, opened 2025-01-25
Add a shared workflow to automatically perform clean code actions

laeubi:


laeubi:
Verification build will not test this

PR #2782, opened 2025-01-26
Add support for supply extra maven args

laeubi:
That way it should be possible to use it for equinox by using:

`extra-maven-args: '-f bundles'`

then at laest it will already work for the bundles (that don't seem to need the binary repo), also useful for other use cases (e.g. enable extra checks or similar)

### DIFF
--- a/.github/workflows/checkVersions.yml
+++ b/.github/workflows/checkVersions.yml
@@ -11,6 +11,11 @@ on:
         description: The name of the bot that adds the necessary version increment changes
         type: string
         required: true
+      extra-maven-args:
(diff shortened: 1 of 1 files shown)


HannesWell (approved):
This would indeed help for Equinox a little bit.
But it would still miss e.g. features. IIRC one cannot specify more than one --file argument?
However this would allow us to e.g. craft a special profile in Equinox that just excludes the binaries or even fetches them differently. So this is helpful in any ways.

laeubi:
Yes it is then more a matter of creativity :-D

PR #2783, opened 2025-01-26
[Version check] Add optional command to setup the build environment

HannesWell:


HannesWell:
Tested in my private repo and the results looked good.

PR #2785, opened 2025-01-26
Make Update target a shared workflow and move configuration into pom

laeubi:
The workflow to update the target file is quite convenient and proven to work well.

This now moves the configuration into the pom and makes the workflow a shared one so it can be used in other projects easily.


laeubi:
Verification won't test this

PR #2787, opened 2025-01-27
Modify ECF setup to eliminate API errors

merks:
- Remove the ssl text modification because the ssl projects are no longer imported.


merks:
Does not affect the build.  Merging immediately.

PR #2790, opened 2025-01-27
[P-build] Various clean-ups of Jenkins pipeline job definition

HannesWell:
Apply the clean-ups recently applied to I/Y-builds also to the P-build job definition.


HannesWell:
The verification build doesn't test this. Submitting it immediately.

PR #2792, opened 2025-01-28
Enhance cleanup profile to include private final fields and deprecation

laeubi:
This enables the cleanup of private fields that can be made final and replacing of deprecated method calls.


laeubi:
Verification won't test this.

PR #2795, opened 2025-01-30
Install more jvms for the cleancode action

laeubi:
Currently we only have a Java 21 VM for the cleancode action but JDT works best with a real JVM (especially for those without a release option like Java 8), so this one install some more so they are available at the toolchains and can be discovered by Tycho to setup a JDT JRE in the cleanups.


laeubi:
Wont be tested by the verifications

PR #2798, opened 2025-01-30
[Build] Improve labels and order of build- and test-log links

HannesWell:


HannesWell:
This is the state before this change
grafik

And the two marked bullet-points are changed to
- `Build and Release Engineering Logs`
- `Test Console Logs`

PR #2799, opened 2025-01-30
[Build] Unify references to I-Build composite repository

HannesWell:


HannesWell:
Finally I found the reason for the build-failure: P2 redefines the `releaseVersion` property with the value `4.35.0`, which is different to the value from the parent `4.35`:
https://github.com/eclipse-equinox/p2/blob/4de9a3308adafa4066627b6a747e0902b426b8dc/pom.xml#L29

And then of course the URL with the different value points into nowhere. So we either have to use different variable name or make sure that the value is the same to make this work.

HannesWell:
With https://github.com/eclipse-equinox/p2/pull/686 applied this finally works as intended.

PR #2825, opened 2025-02-19
[I/Y-build] Change mail subject if build is skipped to lower-case

HannesWell:
@laeubi FYI


HannesWell:
The verification build doesn't test this, submitting.

PR #2791, opened 2025-01-27
[P-build] Remove all P-build files, scripts and instructions

HannesWell:
The JDT project manages P-builds by itself since
- https://github.com/eclipse-jdt/eclipse.jdt.core/issues/3244

Therefore all files and content related to P-builds can be removed from this repository.

PR #2835, opened 2025-02-21
[P-Build] Remove now obsolete 'AddToPComposite' Jenkins job definition

HannesWell:
Now that P-builds are managed by JDT itself and the P-build job doesn't reside in the RelEng-Jenkins instance anymore the job to add a build to the P-composite repository is obsolete.

Leftover of
- https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/pull/2791


HannesWell:
The verification build does not test this and it's a uncritical RelEng clean-up.
Thus submitting this immediately and during the freeze-period.

PR #2876, opened 2025-03-04
Switch sdk repo to 4.35 as there is no 4.36 yet

akurtakov:
Should allow to unbreak
https://github.com/eclipse-platform/eclipse.platform.swt/pull/1876


akurtakov:
Push without waiting for verification to speed up unbreaking the build.

PR #2881, opened 2025-03-05
Enable 'Convert control statement bodies to block' cleanup

laeubi:


laeubi:
Build will not test this.

PR #2897, opened 2025-03-07
Split releaseYear in its own property in parent

akurtakov:
Needed in order to reduce hard coded years (e.g. in generated javadoc).

### DIFF
--- a/eclipse-platform-parent/pom.xml
+++ b/eclipse-platform-parent/pom.xml
@@ -44,6 +44,8 @@
     <buildId>${buildType}${buildTimestamp}</buildId>
     <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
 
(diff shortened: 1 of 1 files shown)


eclipse-platform-parent/pom.xml
```diff
@@ -44,14 +44,16 @@
     <buildId>${buildType}${buildTimestamp}</buildId>
     <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
 
+	<!-- releaseYear is used in copyright statements, etc.-->
+    <releaseYear>2025</releaseYear>
```
laeubi, line 48: Have you tested that this is correctly updated using the release prepare workflow?

akurtakov: I haven't even thought about that . What/how/where to test ?

PR #2898, opened 2025-03-07
Test if we can use -pl !:infocenter-web to avoid building that module

merks:


merks:
We've just increased the heap instead.

PR #2899, opened 2025-03-07
Set copyright year to release year in javadoc

akurtakov:
Using https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/pull/2897 it will remove the need to update it (which has been forgotten during the whole 2024 - it's still 2023 in our 2025-03 release).


akurtakov:
Build OOMs, merging as it works fine locally.

PR #2902, opened 2025-03-07
Add pty4j to the target platform

laeubi:
Required for 
- https://github.com/eclipse-platform/eclipse.platform/pull/1762

### DIFF
--- a/eclipse.platform.releng.prereqs.sdk/eclipse-sdk-prereqs.target
+++ b/eclipse.platform.releng.prereqs.sdk/eclipse-sdk-prereqs.target
@@ -864,6 +864,22 @@
 			  </dependency>
 		  </dependencies>
 	  </location>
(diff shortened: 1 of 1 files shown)


eclipse.platform.releng.prereqs.sdk/eclipse-sdk-prereqs.target
```diff
@@ -864,6 +864,22 @@
 			  </dependency>
 		  </dependencies>
 	  </location>
+	  <location includeDependencyDepth="none" includeDependencyScopes="compile" includeSource="true" label="Jetbrains" missingManifest="generate" type="Maven">
```
merks, line 867: Does the manifest need to be generated?

laeubi: Yes, sadly they use gradle otherwise I would "just" suggest to add manifest headers.

laeubi:
I will close this as we can use CDT native as a good alternative here once

- https://github.com/eclipse-cdt/cdt/pull/1109

is merged!

PR #2903, opened 2025-03-07
Ensure prepare release works with split year/month

akurtakov:


akurtakov:
@laeubi I've verified locally it does the proper thing but I would appreciate your review here.

laeubi (approved):
This looks good to me.

Issue #2889, opened 2025-03-06
Publish to Maven central


MohananRahul:

elsazac:
@MohananRahul 
Could you update the final release repository details here?

MohananRahul:
Eclipse downloads:
https://download.eclipse.org/eclipse/downloads/drops4/R-4.35-202502280140/

New and Noteworthy:
https://www.eclipse.org/eclipse/news/4.35/

Update existing (non-production) installs:
https://download.eclipse.org/eclipse/updates/4.35/

Specific repository good for building against:
https://download.eclipse.org/eclipse/updates/4.35/R-4.35-202502280140/

Equinox specific downloads:
https://download.eclipse.org/equinox/drops/R-4.35-202502280140/

elsazac:
@merks 

I ran the publish to maven job in jenkins, it passed without fail https://ci.eclipse.org/releng/view/Publish%20to%20Maven/job/Releng/job/PublishToMaven/82/ , but i am unable to close the repos from the staging area;

The three repos's fail to close with the below error :

JDT :
```
typeId	javadoc-staging
failureMessage	Missing: no javadoc jar found in folder '/org/eclipse/jdt/org.eclipse.jdt.jeview/1.6.0'
failureMessage	Missing: no javadoc jar found in folder '/org/eclipse/jdt/org.eclipse.jdt.bcoview/1.3.0'
failureMessage	Missing: no javadoc jar found in folder '/org/eclipse/jdt/org.eclipse.jdt.apt.ui/3.8.600'
failureMessage	Missing: no javadoc jar found in folder '/org/eclipse/jdt/org.eclipse.jdt.ui.unittest.junit/1.2.0'
failureMessage	Missing: no javadoc jar found in folder '/org/eclipse/jdt/org.eclipse.jdt.launching/3.23.200'
failureMessage	Missing: no javadoc jar found in folder '/org/eclipse/jdt/ecj/3.41.0'
failureMessage	Missing: no javadoc jar found in folder '/org/eclipse/jdt/org.eclipse.jdt.debug.ui/3.14.0'
failureMessage	Missing: no javadoc jar found in folder '/org/eclipse/jdt/org.eclipse.jdt.junit.core/3.13.500'
failureMessage	Missing: no javadoc jar found in folder '/org/eclipse/jdt/org.eclipse.jdt.astview/1.7.0'
[...]

PR #2904, opened 2025-03-08
Remove guard preventing all javadoc.jar files from being created

merks:
https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/issues/2889


merks:
The build does not test this.

PR #2905, opened 2025-03-08
Increase MAVEN_OPTS = '-Xmx3300m' to avoid out-of-memory building war

merks:
- The infocenter war is building in parallel typically with infocenter product assembly and this randomly runs out of heap space, so with a little more heap available we should avoid this problem.


merks:
Certainly the memory is being used and we see even this in the log:

image

There's more memory available when building the war:

image

PR #2912, opened 2025-03-10
Fix checkCompositesValidity

akurtakov:
Has been failing since "June 12, 2024" but went unnoticed as it was sending mails to Sravan only.
Problem was it tried to run with Java 11 but now Java 21 and during the whole period Java 17 was needed.


akurtakov:
Merge directly as it would not be verified.

PR #2922, opened 2025-03-14
Update icu4j to 77.1.

merks:


merks:
Ignore heap space war error.

PR #2925, opened 2025-03-16
Enable "Use pattern matching for instanceof" in the cleanup profile

laeubi:


laeubi:
Will not be checked by the CI builds

PR #2932, opened 2025-03-18
Supply testManifest file name directly

akurtakov:
Should fix annoying and big "Releng: Missing testManifest.xml Files" section on downloads page.


akurtakov:
Verification will not test it. Merging.

PR #2948, opened 2025-03-27
Add SWT SVG fragment to SDK tests feature

HeikoKlare:
The SVG fragment needs to be available in SDK tests product as SWT tests require that fragment and the SDK does not currently not include it.


HeikoKlare:
We had quite some discussion on this topic in:
- https://github.com/eclipse-platform/eclipse.platform.swt/issues/1944
- https://github.com/eclipse-platform/eclipse.platform.ui/pull/2862
- https://github.com/eclipse-platform/eclipse.platform.ui/pull/2593
- ...

This seems to be the attempt that least people complained about.

So I will merge this to test if it finally fixes https://github.com/eclipse-platform/eclipse.platform.swt/issues/1944

PR #2951, opened 2025-03-28
Reduce noise in the build output

akurtakov:
These extra outputs clutter the build output and make it harder to find and analyze problems.
Such options should be enable temporarily when issue is analyzed and removed after that.

### DIFF
--- a/.mvn/jvm.config
+++ b/.mvn/jvm.config
@@ -1,1 +1,1 @@
--XX:+PrintFlagsFinal
+


akurtakov:
Looking at failing master build now - most of the output is due to this flag https://ci.eclipse.org/platform/job/eclipse.platform.releng.aggregator/job/master/3345/consoleFull .

akurtakov:
Merge in order to make it easier to analyze logs.

Issue #2950, opened 2025-03-28
4.36 I-Build: I20250327-1800 - BUILD FAILED


iloveeclipse:
https://ci.eclipse.org/releng/job/Builds/job/I-build-4.36/31/console

```
23:07:28  [INFO] Scanning for projects...
23:07:37  [ERROR] [ERROR] Some problems were encountered while processing the POMs:
23:07:37  [FATAL] Non-resolvable parent POM for org.eclipse.platform:org.eclipse.equinox.security.ui:1.4.500-SNAPSHOT: The following artifacts could not be resolved: org.eclipse.platform:bundles:pom:4.36.0-SNAPSHOT (absent): Could not find artifact org.eclipse.platform:bundles:pom:4.36.0-SNAPSHOT and 'parent.relativePath' points at wrong local POM @ 
23:07:37  [FATAL] Non-resolvable parent POM for org.eclipse.platform:org.eclipse.equinox.bidi.tests:1.4.300-SNAPSHOT: The following artifacts could not be resolved: org.eclipse.platform:bundles:pom:4.36.0-SNAPSHOT (absent): Could not find artifact org.eclipse.platform:bundles:pom:4.36.0-SNAPSHOT and 'parent.relativePath' points at wrong local POM @ 
23:07:37  [FATAL] Non-resolvable parent POM for org.eclipse.platform:org.eclipse.equinox.cm.test:1.2.0-SNAPSHOT: The following artifacts could not be resolved: org.eclipse.platform:bundles:pom:4.36.0-SNAPSHOT (absent): Could not find artifact org.eclipse.platform:bundles:pom:4.36.0-SNAPSHOT and 'parent.relativePath' points at wrong local POM @ 
```

iloveeclipse:
@laeubi : I don't see any change in pom files neither in equinox nor in the platform or aggregator responsible for the build failure, could it be some tycho snapshot issue?

akurtakov:
https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/pull/2952 points that it's not Tycho regression.

merks:
Looking at what has changed in all the repositories in the last day, these are the only things that look suspiciously related:

https://github.com/eclipse-platform/eclipse.platform.swt/commit/91e2c7b9acafaee59f133aa8e5b9a690eaabd88e
[...]

PR #2952, opened 2025-03-28
Revert to released Tycho 4.0.12

akurtakov:
Experiment to check whether it's tycho regression for https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/issues/2950


akurtakov:
Close as this doesn't seem to influence the build.

PR #2971, opened 2025-04-04
Add more bundles for which to not show deprecations

akurtakov:
Based on content of
https://download.eclipse.org/eclipse/downloads/drops4/I20250403-1800/apitools/deprecation/apideprecation.html , these jetty bundles are not something the project can claim API stability thus excluding from the list.


akurtakov:
This change affects only reports generated by I-builds thus merging without waiting.

PR #2973, opened 2025-04-04
Remove the SWT website from the SDK configuration

merks:
Set the wwd wrapLineLength to zero to avoid formatting changes to avoid formatting changing the entire file.


merks:
Has no build impact.

Issue #2968, opened 2025-04-03
Drop performance tests releng scripts


akurtakov:
There are no reports generated since before 4.33 (https://download.eclipse.org/eclipse/downloads/drops4/R-4.33-202409030240/performance/performance.php), performance jobs are no longer executing (https://ci.eclipse.org/releng/job/PerformanceTests/) and publishing is dependent on php scripts(to be removed end of year). 
Unless, someone says and steps up to make it all usable again I'll start dropping the various scripts next week to make releng more manageable.

akurtakov:
I'm starting to remove performance tests from releng now.

HannesWell:
> Unless, someone says and steps up to make it all usable again

At least the result collection should be fixable in the same way as for the regular result collection as you did in https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/pull/2911.

The performance tests are integrated  relatively deeply into the current test scripts. If you search for `perf` in the corresponding folders, there should be multiple hits. 
There are even labels for it in the infrastructure:
https://github.com/search?q=repo%3Aeclipse-cbi%2Fjiro%20path%3A%2F%5Einstances%5C%2Feclipse%5C.platform%5C.releng%5C%2F%2F%20performance&type=code

So if this is removed I don't think it will come back soon.
But on the other hand reliable and stable results also need corresponding reliable and stable infrastructure and with the current way of dynamic agents, shared resources and containers etc. I don't see that anymore. We would need a dedicated machine that does nothing else, at least when the tests are executed.
It's a bit sad to see such 'framework' be gone, but I also don't see how this will be used in the foreseeable future.
Sorry for the late reply I oversaw this issue in the flood of mails I get and

akurtakov:
I plan to get rid or the perf tests from releng scripts entirely. IMO the whole way they are thrown in is disfunctional now. If/when performance testing is to be taken seriously (I mean results being actually looked and acted upon) we will need way better(and modern) infrastructure and even representing the results. Please let me know if you disagree with the approach I'm taking.

PR #2978, opened 2025-04-08
Drop perf tests from ant and php files

akurtakov:
Reduce the spaghetti there so running and reporting regular tests is streamlined.
Part of https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/issues/2968 .


akurtakov:
This would be verified only by I-build thus merging and starting new I-build.

HannesWell:
Thanks for cleaning this up.
But at the same time this changed many space indented xml-elements to be tab indented. I think this is not ideal as we now have a mixture of styles in the same file and have lost the git-history for all these lines that didn't have an actual change.

akurtakov:
Sorry about that - tbh I'm more concerned about mixture of styles than git history as these files should eventually be gone IMO.

PR #2984, opened 2025-04-10
Stop quering gtk version using rpm

akurtakov:
Tests no longer run on rpm distribution so this effectively removes "rpm: command not found" and empty "GTK Version: " output from tests consolelog (e.g.
https://download.eclipse.org/eclipse/downloads/drops4/I20250409-1800/testresults/consolelogs/ep436I-unit-linux-x86_64-java24_linux.gtk.x86_64_24_consolelog.txt ).
As nobody probably knew about this output I am not looking into converting it to dpkg command but remove the output altogether. It is simpler to run the command (commands as one may need also glib, webkitgtk, glibc, pango, etc. versions) in dedicated Jenkins job to figure out whatever info is needed.

### DIFF
--- a/production/testScripts/configuration/sdk.tests/testScripts/runtests.sh
+++ b/production/testScripts/configuration/sdk.tests/testScripts/runtests.sh
@@ -27,12 +27,9 @@ then
   echo "ERROR: No JVM define, or the defined one was found to not be executable"
   exit 1
 fi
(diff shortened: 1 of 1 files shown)


akurtakov:
This is used only by I-builds thus pushing directly without waiting for verification.

PR #2992, opened 2025-04-19
[Build] Always run Eclipse at the storage-server with Java-21

HannesWell:
Adding a new I-build to the I-build composite seems to not work at the moment and running Eclipse to perform the addition with Java-17 although the Eclipse-SDK/Platform products now require Java-21 could be a cause.

A Java-21 JDK was installed via
- https://gitlab.eclipse.org/eclipsefdn/helpdesk/-/issues/5780

I'm not entirely sure the addition started to fail with the latest I-build, but after running the [Mark Stable](https://ci.eclipse.org/releng/job/Builds/job/markStable/) job, modified to use Java-21, it worked. Running it explicitly previously with Java-17 didn't add the build to the composite.
Maybe this was also just a caching issue at the storage-server, but still this change should be applied.

### DIFF
--- a/JenkinsJobs/Builds/markBuild.groovy
+++ b/JenkinsJobs/Builds/markBuild.groovy
@@ -69,7 +69,7 @@ pipeline {
 						
 						#triggering ant runner
 						baseBuilderDir=${workspace}/eclipse
-						javaCMD=/opt/public/common/java/openjdk/jdk-17_x64-latest/bin/java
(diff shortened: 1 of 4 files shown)


HannesWell:
Regardless of this change, we should rework all changed locations, sooner rather than later, to run Eclipse locally in Jenkins instead of directly at the storage server.
I'll have a look at this.

HannesWell:
The verification build does not test this, thus submitting immediately.

Issue #2959, opened 2025-03-31
Add support for FreeBSD


chirontt:
Update: since there's no objection, I'm now submitting PRs to various Eclipse repositories for this issue. Here's the central TODO list to keep track of various tasks done and/or under way:
* [ ] add support for FreeBSD to `eclipse.jdt.debug` submodule
  - [ ] eclipse-jdt/eclipse.jdt.debug#676
* [ ] add support for FreeBSD to `eclipse.jdt.ui` submodule
  - [ ] eclipse-jdt/eclipse.jdt.ui#2170
* [ ] add support for FreeBSD to `eclipse.pde` submodule
  - [ ] eclipse-pde/eclipse.pde#1726
* [ ] add support for FreeBSD to `eclipse.platform` submodule
  - [ ] eclipse-platform/eclipse.platform#1831
  - [ ] compile the file system natives for FreeBSD. If cross-compiling from Linux is needed, use the [freebsd-cross](https://github.com/chirontt/eclipse.platform.releng.aggregator/tree/master/cje-production/dockerfiles/alpine/freebsd-cross) Docker container as a suitable cross-compile tool - details are in its `README` file for [x86_64](https://github.com/chirontt/eclipse.platform.releng.aggregator/blob/master/cje-production/dockerfiles/alpine/freebsd-cross/README.md#freebsdx86_64-cross-build-for-eclipse-plaforms-file-system-natives) and/or for [aarch64](https://github.com/chirontt/eclipse.platform.releng.aggregator/blob/master/cje-production/dockerfiles/alpine/freebsd-cross/README.md#freebsdaarch64-cross-build-for-eclipse-plaforms-file-system-natives).
* [ ] add support for FreeBSD to `eclipse.platform.swt` submodule
  - [ ] eclipse-platform/eclipse.platform.swt#2044
[...]

PR #2994, opened 2025-04-22
Add FreeBSD platforms' build.

chirontt:
- part of #2959
- add environment triplets for FreeBSD `aarch64` and/or `x86_64`.

### DIFF
--- a/eclipse.platform.releng/publish-to-maven-central/SDK4Mvn.aggr
+++ b/eclipse.platform.releng/publish-to-maven-central/SDK4Mvn.aggr
@@ -12,6 +12,8 @@
   <configurations operatingSystem="linux" windowSystem="gtk" architecture="ppc64le"/>
   <configurations operatingSystem="linux" windowSystem="gtk" architecture="x86_64"/>
   <configurations operatingSystem="linux" windowSystem="gtk" architecture="riscv64"/>
+  <configurations operatingSystem="freebsd" windowSystem="gtk" architecture="aarch64"/>
+  <configurations operatingSystem="freebsd" windowSystem="gtk" architecture="x86_64"/>
   <configurations operatingSystem="macosx" windowSystem="cocoa" architecture="x86_64"/>
   <configurations operatingSystem="macosx" windowSystem="cocoa" architecture="aarch64"/>
   <configurations architecture="x86_64"/>
--- a/cje-production/dockerfiles/alpine/freebsd-cross/Dockerfile
+++ b/cje-production/dockerfiles/alpine/freebsd-cross/Dockerfile
@@ -1,0 +1,137 @@
+# syntax=docker/dockerfile:1
+
+ARG BASE_IMAGE_NAME="alpine"
+ARG BASE_IMAGE_TAG="3.19"
+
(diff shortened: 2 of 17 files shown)


eclipse.platform.releng/publish-to-maven-central/SDK4Mvn.aggr
```diff
@@ -12,6 +12,8 @@
   <configurations operatingSystem="linux" windowSystem="gtk" architecture="ppc64le"/>
   <configurations operatingSystem="linux" windowSystem="gtk" architecture="x86_64"/>
   <configurations operatingSystem="linux" windowSystem="gtk" architecture="riscv64"/>
+  <configurations operatingSystem="freebsd" windowSystem="gtk" architecture="aarch64"/>
+  <configurations operatingSystem="freebsd" windowSystem="gtk" architecture="x86_64"/>
```
merks, line 16: FYI, this won't just work because there needs to be an enum added

image

https://github.com/eclipse-cbi/p2repo-aggregator/blob/2280b81c35c3aeb4808fa01e65cce9efb246058b/plugins/org.eclipse.cbi.p2repo.aggregator/model/Aggregator.ecore#L398-L406

But that's a trivial task...

merks (commented):
Which JDK works on freebsd?

https://adoptium.net/temurin/releases/

Without that there will be no JustJ JREs...

eclipse.platform.releng/publish-to-maven-central/SDK4Mvn.aggr
chirontt: Honestly I'm not sure what the `SDK4Mvn.aggr` file does, so your knowledge and help about it is appreciated, thanks.

Or I can just remove those 2 lines from the file and leave it intact and move on...

chirontt:
> Which JDK works on freebsd?
> 
> https://adoptium.net/temurin/releases/

Not by Temurin. The FreeBSD team ports the OpenJDK releases and make them available as pre-built binaries to their `pkg` package manager for their FreeBSD end users, but they don't release them as tarballs anywhere on the net, as far as I know.

> Without that there will be no JustJ JREs...

No, unfortunately. But the Eclipse packages can still be built for FreeBSD without embedded JREs.

eclipse.platform.releng/publish-to-maven-central/SDK4Mvn.aggr
merks: It would eventually be needed if we want to publish those fragments to maven central.  Best this be done as a separate commit after everything else is actually done.

chirontt:
I've built Eclipse packages for FreeBSD without problems (and without any embedded JREs) in my [fork](https://github.com/chirontt/eclipse-packages) of the eclipse-packages project, with only some trivial changes.

If you're talking about the Eclipse installer app and its JustJ requirement, to produce an Eclipse package on demand, then it may not work for FreeBSD. But the installer app does allow the user to pick a local JDK/JRE installation for it:

image

So I see no problem with it, once the installer is ported to run in FreeBSD. BTW, where is this installer's source code?

PR #2997, opened 2025-04-26
[Build|GH] Add quick-fix to avoid need for EF-infrastrucutre

HannesWell:
In order to mitigate the current outage of the EF-hosted storage, which will probably continue to block builds for at least a few days (https://www.eclipse.org/lists/eclipse.org-committers/msg01493.html) this implements a quick-fix to avoid the need for said EF infrastructure for the shared GitHub workflows. This is achieved by the following actions:
- Obtain the eclipse-parent pom.xml and the sdk.target project from the GH repository and install it within the build into the local Maven cache
- Revert Tycho to use it's latest release 4.0.12, which is available from Maven-Central

Especially the latter could cause problems for builds that rely on the new features only available in Tycho 4.0.13-SNAPSHOT. Therefore this is currently designed as opt-in feature.  But looking at https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/pull/2870#issue-2892484622, only SWT currently needs the snapshot but doesn't use the shared workflow.
After some more testing we could also enable this by default.

### DIFF
--- a/.github/workflows/checkVersions.yml
+++ b/.github/workflows/checkVersions.yml
@@ -26,6 +26,11 @@ on:
         type: string
         required: false
         default: '.'
+      without-EF-infra:
+        description: 'Whether to run this build without the need for EF-infrastructure (which is currently unavailable).'
+        type: boolean
+        required: false
+        default: true
 
 permissions: {} # all none
 
@@ -51,6 +56,12 @@ jobs:
       with:
         maven-version: 3.9.9
 
+    - name: Install Maven parent locally
+      uses: eclipse-platform/eclipse.platform.releng.aggregator/.github/actions/install-maven-parent-action@master
+      if : inputs.without-EF-infra
+      with:
+        tycho-version: 4.0.12
+
     - name: Additional setup
       if: inputs.extra-setup-command
       run: |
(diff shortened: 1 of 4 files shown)


.github/workflows/mavenBuild.yml
```diff
@@ -68,6 +73,13 @@ jobs:
       uses: stCarolas/setup-maven@d6af6abeda15e98926a57b5aa970a96bb37f97d1 # v5
       with:
         maven-version: ${{ inputs.mavenVersion }}
+
+    - name: Install Maven parent locally
+      uses: HannesWell/eclipse.platform.releng.aggregator/.github/actions/install-maven-parent-action@gh-workflow-quickfix
```
HannesWell, line 78: This and the other two calls have to be replaced by the final location before this can be submitted.
```suggestion
      uses: eclipse-platform/eclipse.platform.releng.aggregator/.github/actions/install-maven-parent-action@master
```

.github/workflows/checkVersions.yml
```diff
@@ -26,6 +26,11 @@ on:
         type: string
         required: false
         default: '.'
+      without-EF-infra:
+        description: 'Weather to run this build without the need for EF-infrastructure (which is currently unavailable).'
```
HeikoKlare, line 30:
```suggestion
        description: 'Whether to run this build without the need for EF-infrastructure (which is currently unavailable).'
```

HeikoKlare (approved):
Great workaround, thank you! 🚀 

The change looks good and seems to work (as you show at the platform.ui repo). I only found a nitpicky typo, which you might fix in all three workflows when exchanging the organization name for the used action for final submission.

.github/workflows/mavenBuild.yml
HannesWell: Done.

.github/workflows/checkVersions.yml
HannesWell: Thanks! Fixed it.

Issue #1733, opened 2024-01-17
Evaluate the quality-monitor github action


laeubi:
See:
- https://github.com/uhafner/quality-monitor

This is similar to what Jenkins already offers but gives a much more integrated/nicer feedback and might even replace the Junit result panel we currently have.

akurtakov:
It's worth trying, maybe in PDE first?

jukzi:
please not. we have enough trouble with the jenkins plugins.

merks:
@jukzi

I don't think this is nice commentary.  The title here is to evaluate something.  Surely its reasonable to evaluate something. 

I see both you and @laeubi doing so very much work to make things better.   Sure, sometimes there are growing pains, sometimes there are differences of opinion, and sometimes some things appear worse than before.  For example, the warning elimination effort is great, but already there have been two regressions as a result, one of which cost me the better part of a day to fix.   But I don't say, "eliminate warnings, please don't!", because I respect that you are doing very good things in general with the best of intentions.  Of course none of us is perfect, so mistakes can and do happen to the best of us.  

I think it's important to remain respectful and to recognize the good work that all the other people are doing.  It's not as if @laeubi  and I always see eye-to-eye, but I greatly respect his huge contribution to releng improvements, often with complaints as his only thank you.  It would be good to be just a little nicer.

laeubi:
Yes I'll probably pick that up later.

PR #2998, opened 2025-04-27
[Build|GH] Unify Maven arguments and replace use of coactions/setup-xvfb

HannesWell:
In the shared GH workflows use long option names to be more readable.
Also don't fail the build on test-failures because the the additional test-result check fails in case of test-failures. This simplifies the distinction between 'hard' failures, like compile-errors and test-failures.

Similar to
- https://github.com/eclipse-platform/eclipse.platform.swt/pull/1492
- https://github.com/eclipse-equinox/equinox/pull/677

### DIFF
--- a/.github/workflows/mavenBuild.yml
+++ b/.github/workflows/mavenBuild.yml
@@ -91,9 +91,8 @@ jobs:
       uses: coactions/setup-xvfb@6b00cf1889f4e1d5a48635647013c0508128ee1a
       with:
        run: >- 
-        mvn --batch-mode -V -U -e
+        mvn --batch-mode --no-transfer-progress --show-version --update-snapshots --errors
         ${{ inputs.without-EF-infra && '-Dtycho.version=4.0.12' || '' }}
-        -ntp
         -Dcompare-version-with-baselines.skip=false
         -Pbree-libs
         -Papi-check
--- a/.github/workflows/mavenBuild.yml
+++ b/.github/workflows/mavenBuild.yml
@@ -47,6 +47,10 @@ jobs:
           - { name: Linux,   os: ubuntu-latest  }
           - { name: Windows, os: windows-latest }
           - { name: MacOS,   os: macos-13   }
+    defaults:
+      run: # Run always on bash, because powershell (the Windows default) interprets dots in arguments differently
+        shell: bash
+
     name: Verify ${{ matrix.config.name }}
     steps:
     - name: Checkout
@@ -88,9 +92,8 @@ jobs:
         target: .github/matcher
     - run: echo "::add-matcher::.github/matcher/${{ steps.api-tools-matcher.outputs.filename }}"
     - name: Build with Maven
-      uses: coactions/setup-xvfb@6b00cf1889f4e1d5a48635647013c0508128ee1a
-      with:
-       run: >- 
+      run: >-
+        ${{ runner.os == 'Linux' && 'xvfb-run' || '' }}
         mvn --batch-mode --no-transfer-progress --show-version --update-snapshots --errors
         ${{ inputs.without-EF-infra && '-Dtycho.version=4.0.12' || '' }}
         -Dcompare-version-with-baselines.skip=false
--- a/.github/workflows/mavenBuild.yml
+++ b/.github/workflows/mavenBuild.yml
@@ -85,12 +85,10 @@ jobs:
         tycho-version: 4.0.12
 
     - name: Download the API Tools matcher
-      uses: suisei-cn/actions-download-file@818d6b7dc8fe73f2f924b6241f2b1134ca1377d9 # v1.6.0
-      id: api-tools-matcher
-      with:
-        url: "https://raw.githubusercontent.com/eclipse-platform/eclipse.platform.releng.aggregator/master/.github/matcher/api-tools.json"
-        target: .github/matcher
-    - run: echo "::add-matcher::.github/matcher/${{ steps.api-tools-matcher.outputs.filename }}"
+      run: |
+        mkdir .github/matcher
+        curl -o .github/matcher/api-tools.json https://raw.githubusercontent.com/eclipse-platform/eclipse.platform.releng.aggregator/master/.github/matcher/api-tools.json
+        echo "::add-matcher::.github/matcher/api-tools.json"
     - name: Build with Maven
       run: >-
         ${{ runner.os == 'Linux' && 'xvfb-run' || '' }}
--- a/.github/workflows/codeQLworkflow.yml
+++ b/.github/workflows/codeQLworkflow.yml
@@ -90,9 +90,8 @@ jobs:
         tycho-version: 4.0.12
 
     - name: Build with Maven
-      uses: coactions/setup-xvfb@6b00cf1889f4e1d5a48635647013c0508128ee1a
-      with:
-       run: >- 
+      run: >-
+        ${{ runner.os == 'Linux' && 'xvfb-run' || '' }}
         mvn --batch-mode --no-transfer-progress --show-version --update-snapshots --errors
         ${{ inputs.without-EF-infra && '-Dtycho.version=4.0.12' || '' }}
         -Dcompare-version-with-baselines.skip=true
--- a/.github/workflows/codeQLworkflow.yml
+++ b/.github/workflows/codeQLworkflow.yml
@@ -93,9 +93,8 @@ jobs:
       uses: coactions/setup-xvfb@6b00cf1889f4e1d5a48635647013c0508128ee1a
       with:
        run: >- 
-        mvn --batch-mode -V -U
+        mvn --batch-mode --no-transfer-progress --show-version --update-snapshots --errors
         ${{ inputs.without-EF-infra && '-Dtycho.version=4.0.12' || '' }}
-        -ntp
         -Dcompare-version-with-baselines.skip=true
         -Pbree-libs
         -DskipTests=true


.github/workflows/mavenBuild.yml
```diff
@@ -88,16 +92,14 @@ jobs:
...
+      run: >-
+        ${{ runner.os == 'Linux' && 'xvfb-run' || '' }}
+        mvn --batch-mode --no-transfer-progress --show-version --update-snapshots --errors
         ${{ inputs.without-EF-infra && '-Dtycho.version=4.0.12' || '' }}
-        -ntp
         -Dcompare-version-with-baselines.skip=false
         -Pbree-libs
         -Papi-check
-        --fail-at-end
+        --fail-at-end -Dmaven.test.failure.ignore=true
```
laeubi, line 102: I don't think such generic properties should be set in a shared action. If this is wanted (and projects might decide differently) its better to either pass as an extra option in the calling workflow or put into `maven.config`.

laeubi (changes requested):

.github/workflows/mavenBuild.yml
HannesWell: It's a generic option yes, but I assume that each project actually would want that or at least would not object.
What do you think speaks against this?

laeubi: > What do you think speaks against this?

1. It is bad practice , as it for example will build downstream projects that then fail with more tests because they rely on correct behavior. And still the "bad" module will be marked as `SUCCESS`.
2. "what projects actually want" is changing / different and you can't override the option if given on the commandline
3. I even remember we have used that option in the past and the people complained that the check shows green even it was failing tests
4. If the additional check fails to run (for whatever reason) then you even won't get a feedback at all

Actually we should even think about removing fail at end... If something is wrong we should report a failure as soon as possible and not waste resources for things that *might* still work.

laeubi:
> This simplifies the distinction between 'hard' failures, like compile-errors and test-failures.

To be honest I expect that under normal conditions no one checks in code that do not compile and we should not make it "comfortable" or "easier" ... if one wants its easy enough to even run a maven build locally.

.github/workflows/mavenBuild.yml
HeikoKlare: I have never though about this option, but when doing so now, I also see some benefits of it:
- People would be better notified about any kind of failure not related to tests. I have seen merges despite different kinds of failures as they were "hidden" by the fact that there was a test failure which was just considered to be the only cause for the workflow to fail.
- The results would be obviously "cleaner" as a test failure only makes one check fail, so it removes the manual effort take a look at multiple check results in case of a failure.

But I also see the drawback that "success" for an execution with test failure is a bit strange and probably there should an option for consuming projects to opt-out.

> 4\. If the additional check fails to run (for whatever reason) then you even won't get a feedback at all

This is quite important point. Even if the additional check does not fail, it takes some time (seconds) until its done and shows up. That check is not shown as "pending" but just appears several seconds after the builds finished. I have seen people waiting actively for the builds to finish to directly merge the change, so this could result in PRs being merged right after all checks become green even though few seconds later a failure pops up.
Thus, I think we should somehow ensure that either the test results check shows up as "pending" or that the builds wait for the test results to show up before finishing (but then a build would only be marked as completed when all of them are).

> Actually we should even think about removing fail at end... If something is wrong we should report a failure as soon as possible and not waste resources for things that already are proven to fail.

This is something we cannot do as we have sporadically failing tests and people are (unfortunately) used to merging PRs with such a failing referring to the issue reporting that failure. So this could easily lead to test failure that would occur after such a known one to stay undiscovered.

HeikoKlare (commented):
I really like the usage of long names to ease understanding of the commands.
I just agree that the option to igore test failures is a bit controversional and may be something to extract into a separate PR with according discussion.

HannesWell:
> I just agree that the option to igore test failures is a bit controversional and may be something to extract into a separate PR with according discussion.

Ok, then let's move that to
- https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/pull/3001

HeikoKlare (approved):

.github/workflows/mavenBuild.yml
```diff
@@ -47,6 +47,10 @@ jobs:
           - { name: Linux,   os: ubuntu-latest  }
           - { name: Windows, os: windows-latest }
           - { name: MacOS,   os: macos-13   }
+    defaults:
+      run: # Run on cmd on Windows because powershell (the default) interprets dots in arguments differently
+        shell: ${{ runner.os == 'Windows' && 'cmd' || 'bash' }}
```
HannesWell, line 52: Actually we could just use bash on all OS, even on Windows:
`"When specifying a bash shell on Windows, the bash shell included with Git for Windows is used."` -
https://docs.github.com/en/actions/writing-workflows/workflow-syntax-for-github-actions#defaultsrunshell

From my experience using the git-bash on Windows works fine for the usual things. At least the current state works fine with it. This would also help if more complicated scripts are created, where the commands are not the same on Windows' CMD and a UNIX bash.

What do you think?

HannesWell (commented):
Besides moving the ignore-test-failures option out, I have no fixed a misconfiguration in the shell selection, but this could also be made the default (see below).

Furthermore I have simplified the Download of the API Tools matcher to just use local tools (i.e. curl).

.github/workflows/mavenBuild.yml
HeikoKlare: Sounds good 👍

.github/workflows/mavenBuild.yml
```diff
@@ -81,19 +85,15 @@ jobs:
...
-      uses: suisei-cn/actions-download-file@818d6b7dc8fe73f2f924b6241f2b1134ca1377d9 # v1.6.0
-      id: api-tools-matcher
-      with:
-        url: "https://raw.githubusercontent.com/eclipse-platform/eclipse.platform.releng.aggregator/master/.github/matcher/api-tools.json"
-        target: .github/matcher
-    - run: echo "::add-matcher::.github/matcher/${{ steps.api-tools-matcher.outputs.filename }}"
+      working-directory: .github/matcher
+      run: |
+        curl -o api-tools.json https://raw.githubusercontent.com/eclipse-platform/eclipse.platform.releng.aggregator/master/.github/matcher/api-tools.json
+        echo "::add-matcher::.github/matcher/api-tools.json"
```
HeikoKlare, line 91: Is this still correct when running this in working directory `.github/matcher`? I have not clue about this statement/syntax, so just wondering whether the path inside the echo call may be relative to the current location.

HeikoKlare (commented):
Changes still look good to me. Using curl instead of the download actions seems to be fine as we don't use much of the functionality of that download action anyway.

.github/workflows/mavenBuild.yml
laeubi: I think we can simply remove the matcher as we have better ways today to get notified about it and I think it even won't work anymore.

HannesWell: It should be, because that just echos the plain string without any knowledge of the working directory.
In general this is a command send to the runner:
https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/workflow-commands-for-github-actions#using-workflow-commands-to-access-toolkit-functions



> I think we can simply remove the matcher as we have better ways today to get notified about it and I think it even won't work anymore.

That would be even better. You are referring to the Jenkins quality-gate?
Can you tell (again), how the matcher in GH is supposed to work? Are these the markers that are shown in the GH-UI in the `Files changed` tab, if some API errors or warnings occur? Because IIRC I have seen them in the past. But maybe it's from the long past. In general I think such display in the GH-UI would still be convenient and worth to fix, if it's not too complicated.

laeubi: As mentioned I don't think it work anymore, it was done at a time where we have no Jenkins parser for the API problems and as we do not print the API problems to the log it won't work anyways.

If we want something similar to Jenkins for Github actions we should aim for:
- https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/issues/1733

HeikoKlare: Thank you for the pointer to https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/issues/1733! I wasn't aware of that issue existing, but I think it becomes more than relevant now that we think about how to be/become more independent of Jenkins / EF infrastructure anyway.

.github/workflows/mavenBuild.yml
HannesWell: Done. :)

.github/workflows/mavenBuild.yml
HannesWell: > As mentioned I don't think it work anymore, it was done at a time where we have no Jenkins parser for the API problems and as we do not print the API problems to the log it won't work anyways.

API problems are still reported, even in GH workflows:
```
[INFO] --- tycho-apitools:4.0.12:verify (verify) @ org.eclipse.e4.core.commands ---
[...]
[INFO] API Analysis finished in 6 s.
[WARNING] [API WARNING] File ExpressionContext.java at line 27: ExpressionContext illegally implements IEvaluationContext (location: D:\a\eclipse.platform.ui\eclipse.platform.ui\bundles\org.eclipse.e4.core.commands\src\org\eclipse\e4\core\commands\ExpressionContext.java)
[WARNING] 1 API problems can't be mapped to the compiler log!
[INFO] 0 API ERRORS
[INFO] 1 API warnings
```
So I currently don't understand why it shouldn't work anymore, but I'm testing it with:
https://github.com/eclipse-platform/eclipse.platform.ui/pull/2940

HannesWell: Even after enhancing the pattern it still does not work so I moved that part into:
- https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/pull/3002

If it still can't be made to work, we can still remove it from there.

.github/workflows/mavenBuild.yml
HannesWell: > I even remember we have used that option in the past and the people complained that the check shows green even it was failing tests

I would say that each failing check should be inspected/understood in detail.

> * The results would be obviously "cleaner" as a test failure only makes one check fail, so it removes the manual effort take a look at multiple check results in case of a failure.

That's my main motivation as I find it very annoying to have to check two logs for one failure, just to be sure nothing is hidden. Of course we could also 'just' fix all flaky tests.

> * People would be better notified about any kind of failure not related to tests. I have seen merges despite different kinds of failures as they were "hidden" by the fact that there was a test failure which was just considered to be the only cause for the workflow to fail.

There can be actually two kinds of hidden. Hidden in the sense that the committer does not inspect the build-log because it is assumed that the test-failure is unrelated. And hidden that a failing test-project prevents the execution of another one that also has test-failures that don't occur because they are not executed. Although this is currently unlikely as we run with `-fae` and test inter-dependencies are at least very uncommon. But that's basically Heiko's last point.

PR #3001, opened 2025-04-27
[Build] Don't fail shared GH-workflow on test-failures

HannesWell:
Don't fail the build on test-failures because the the additional test-result check fails in case of test-failures. This simplifies the distinction between 'hard' failures, like compile-errors and test-failures.

### DIFF
--- a/.github/workflows/mavenBuild.yml
+++ b/.github/workflows/mavenBuild.yml
@@ -84,7 +84,7 @@ jobs:
         -Dcompare-version-with-baselines.skip=false
         -Pbree-libs
         -Papi-check
-        --fail-at-end
+        --fail-at-end -Dmaven.test.failure.ignore=true
         ${{ inputs.maven-goals }}
     - name: Upload Test Results for ${{ matrix.config.name }}
       if: always()


.github/workflows/mavenBuild.yml
```diff
@@ -97,7 +97,7 @@ jobs:
         -Dcompare-version-with-baselines.skip=false
         -Pbree-libs
         -Papi-check
-        --fail-at-end
+        --fail-at-end -Dmaven.test.failure.ignore=true
```
laeubi, line 100:
```suggestion
```
As mentioned in the other PR I think we better should even remove `--fail-at-end` (and possibly `-Pbree-libs`  / `-Papi-check`)

The workflow should just replicate the usual workflow/setup and if projects want they can set whatever option seems suitable for them (and change it without affecting others) e.g. in `maven.config` or with maven options specified.

Even though it might feels a bit "duplicated" it makes it much easier to run a local build with the recommended options without the need to dig into CI files to find out the desired options, and gives much more freedom.

laeubi (changes requested):

.github/workflows/mavenBuild.yml
HeikoKlare: I still think we must not remove that (https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/pull/2998#discussion_r2062500096):
> This is something we cannot do as we have sporadically failing tests and people are (unfortunately) used to merging PRs with such a failure referring to the issue reporting that failure. So this could easily lead to test failure that would occur after such a known one to stay undiscovered.

HeikoKlare (commented):
In my opinion, this cannot be solved by simply ignoring the test results in the build workflow, as mentioned in https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/pull/2998#discussion_r2062500096. There is the additional test results check, but:
> Even if the additional check does not fail, it takes some time (seconds) until its done and shows up. That check is not shown as "pending" but just appears several seconds after the builds finished. I have seen people waiting actively for the builds to finish to directly merge the change, so this could result in PRs being merged right after all checks become green even though few seconds later a failure pops up.
Thus, I think we should somehow ensure that either the test results check shows up as "pending" or that the builds wait for the test results to show up before finishing (but then a build would only be marked as completed when all of them are).

And even if the test results check shows up early enough, I bet it will happen that people see that the three GHA builds (and maybe Jenkins) ran green and then merge without waiting for the test results check. You will say that every commiter _must_ have a proper look at every check and I totally agree with that, which I why I think we have to accept that risk.

PR #3002, opened 2025-04-27
[Build] Remove unusable 'api-problems'-matcher for shared GH workflows

HannesWell:
The current api-problems matcher seems not to work anymore.

I have the impression that could be because the current pattern doesn't match the print-out, which is in the 'raw'-view
`[ERROR] [API ERROR] File MANIFEST.MF at line 7: The minor version should be incremented in version 1.1.600, since new APIs have been added since version 1.1.500 (location: /home/runner/work/eclipse.platform.ui/eclipse.platform.ui/bundles/org.eclipse.e4.core.commands/META-INF/MANIFEST.MF)`
respectively in the 'default' view:
`Error:  [API ERROR] File MANIFEST.MF at line 7: The minor version should be incremented in version 1.1.600, since new APIs have been added since version 1.1.500 (location: /home/runner/work/eclipse.platform.ui/eclipse.platform.ui/bundles/org.eclipse.e4.core.commands/META-INF/MANIFEST.MF)`

Although the new pattern matches the different beginning of the line better, it looks like the pattern still does not match. But according to the documentation of [`problem-matchers`](https://github.com/actions/toolkit/blob/main/docs/problem-matchers.md) everything else seems to be in place.

That workflow run shows that the matcher is added:
https://github.com/eclipse-platform/eclipse.platform.ui/actions/runs/14694871712/job/41235517380

So probably the pattern still does not work somehow.


laeubi:
@HannesWell the pattern matcher was always fragile, I therefore recommend to simply remove this entirely. We already have issues reported by Jenkins now in a much better way.

HannesWell:
I was hoping for a quick fix, but it looks like something more has to be done and I doubt it's worth the time now. I'll change this PR to remove the problem matcher part this evening.
If somebody want's to fix it later, this can then cleanly be reverted.

PR #3004, opened 2025-04-28
Move shared workflows into folder '.github/workflows/shared'

HannesWell:
We have more and more GH workflows in this repository that are shared with other repos.
While this is good from a unification POV, it becomes more and more difficult to distinguish which workflow is a shared one and which workflow is just used in this repository.

Therefore this PR proposes to move all shared workflows into a sub-folder named `shared`, in in order to helps to distinguish workflows shared with and called from other repositories and workflows only used in this repository.

If we have consensus on this change I can create corresponding PRs to adapt all calling projects in the SDK.
This would also be a good opportunity to rename workflows if we think there current name does not fit.


HeikoKlare:
This is a good idea! For me, it would also help to give more structure to the shared workflows, i.e., to better separate them by purpose. So would it make sense to take the chance and also give the shared workflows more structure (e.g., by placing them in further subfolders)? Would probably be a good point in time since we need to touch consumers anway.

One proposal (without thinking much about it, so feel free to propose something else):
- `prerequisites` containing the freeze period verification, merge commit checks etc.
- `updates` containing the target and release update workflows
- `verification` containing the build and test publish worklows

HannesWell:
> Also I'm not sure what an "Eclipse top-level project" is at least other projects often follow what _Eclipse Platform_ do and I don't see much benefit in breaking them just for the sake of a filename change.

The eclipse top level projects is the parent of the projects forming the SDK: https://projects.eclipse.org/projects/eclipse
Namely, Eclipse-Platform, Equinox, JDT and PDE.
And while I think it's good if other (Eclipse) projects follow what we do, I don't consider these workflow a deliverable whose primary use-case is reuse by everybody. Of course other can use them too, but have to adapt if they change. Otherwise the workflows would be set in stone and I don't think we want to carry that bourdon too.

But any ways, this will cause a lot of work even just internally and I think it doesn't carry that weight at the moment, thus I'm closing this.

PR #3007, opened 2025-05-01
Use trim() instead of strip() because Jenkins doesn't like it now.

merks:


merks:
I previously tested this with a replay.

HannesWell:
We have to be careful with this as `trim()` only removes blank spaces while `strip()` removes all kind of white space. And sometimes the called CLI tools also include a line-break to beautify their output.
Therefore I requested the permission of `String.strip()` back then, in https://gitlab.eclipse.org/eclipsefdn/helpdesk/-/issues/2936
I hope it's added back.

PR #3009, opened 2025-05-01
[Build] Use String.strip() instead of trim()

HannesWell:
This also reverts
- https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/pull/3007


HannesWell:
In fact `String.trim()` already does trim line-breaks. Its Javadoc says:
```
Returns a string whose value is this string, with all leading and trailing space removed,
where space is defined as any character whose codepoint is less than or equal to 'U+0020' (the space character).
```
Thanks @merks for pointing this out.
So this change is not necessary in practice. Closing.

PR #3021, opened 2025-05-05
Enable 'Remove unnecessary array creation' in the cleanup profile

HannesWell:
As discussed in
- https://github.com/eclipse-equinox/equinox/pull/936#discussion_r2072660096


laeubi (commented):
Looks good! PRs are already created from the nightly run.

PR #3023, opened 2025-05-06
Add terminal control to update site

laeubi:
Since we migrated it from CDT, there should be a way for CDT to get rid of this bundle. Adding it to the update site would allow this without platform explicitly referencing it (yet).

### DIFF
--- a/eclipse.platform.releng.tychoeclipsebuilder/eclipse.platform.repository/category.xml
+++ b/eclipse.platform.releng.tychoeclipsebuilder/eclipse.platform.repository/category.xml
@@ -40,4 +40,5 @@
    <bundle id="jakarta.annotation-api" version="1.3.5"/>
    <bundle id="jakarta.inject.jakarta.inject-api" version="1.0.5"/>
    <bundle id="org.eclipse.equinox.slf4j"/>
+   <bundle id="org.eclipse.tm.terminal.control"/>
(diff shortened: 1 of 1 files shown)


laeubi:
The license check likley fails because we need an ibuild to update the branch pointers?!

As Jenkins succeed I think it can be merged.

mickaelistria:
Shouldn't it be added to some features/products as well?

laeubi:
@mickaelistria this will be automatically pulled by features/products once there is a reference so there is usually no need to explicitly include it.

merks:
Moreover, CDT currently has its own version that is included in a feature so, given this bundle is a singleton, so we need to be careful during the transition period that there isn't more than one feature including it because if those includes are of different versions, the two feature could not both be installed.

FYI @jonahgraham

mickaelistria:
> this will be automatically pulled by features/products once there is a reference so there is usually no need to explicitly include it.

What is the dependency chain that includes it eg in the Eclipse SDK product right now?

merks:
Nothing includes it yet.  So indeed, it's not in the product, yet either.   That's why it's added to the category.xml directly, like is the case for other bundles mentioned directly in the category.xml, e.g., `org.eclipse.equinox.slf4j` so that we can make it available for consumption by CDT.

PR #3026, opened 2025-05-07
YBuild for Java25

MohananRahul:
Update YBuild config for Java 25


MohananRahul:
Merging now , PR build will not verify it.

PR #3032, opened 2025-05-08
Limit number of cleanup PRs

laeubi:
currently an unlimited number of PRs are created for cleanups. Especially when enabling new cleanups this can lead to a massive amount of PRs created what are hard to manage and probably block the CI.

This now enables a limit of 10 open PRs at a time.

### DIFF
--- a/.github/workflows/cleanCode.yml
+++ b/.github/workflows/cleanCode.yml
@@ -78,16 +78,28 @@ jobs:
       max-parallel: 1
       fail-fast: false
     steps:
+      - name: List number of open PRs
+        id: list-prs
+        run: |
+            json_output=$(gh pr list -l 'cleanup' -R $OWNER/$REPO -L 100 --json id)
+            echo "prs=$json_output" | tee -a "$GITHUB_OUTPUT"
+        env:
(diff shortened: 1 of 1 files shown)


laeubi:
Build will not test this!

.github/workflows/cleanCode.yml
```diff
@@ -78,16 +78,28 @@ jobs:
       max-parallel: 1
       fail-fast: false
     steps:
+      - name: List number of open PRs
+        id: list-prs
+        run: |
+            json_output=$(gh pr list -l 'cleanup' -R $OWNER/$REPO -L 100 --json id)
```
fedejeanne, line 84: Since you only want to allow 10 clean ups to run in parallel anyway, wouldn't it suffice to use the parameter `-L 100` ?

fedejeanne: Just out of curiosity. I'm no expert in CI/CD 😅

laeubi: Yes I think so, but it would also not harm here as we only fetch the ID, I used this mainly to debug in the output how many are actually matching.

PR #3034, opened 2025-05-08
Use the latest feature of the CBI p2 Aggregator for SDK4Mvn.aggr

merks:
- Specify the root features explicitly to avoid aggregating tests.
- Use includedIUPattern to specify which IUs we want to aggregated, i.e., only JDT, PDE, and Platform IUs.
- Use validateNexusPublishingRequirements to ensure we don't miss any required content in the pom.
- Use excludeFeatures to exclude features, binary content, and purely metadata content.


merks:
Has no build impact.

PR #3035, opened 2025-05-08
Simplify publishToMaven.jenkinsfile

merks:
- There is no need to filter out wanted content because the aggregator only aggregates what we want to publish.
- There is no need to enrich the poms because the aggregator creates poms the meet the nexus publishing requirements and tests that the poms do actually conform.
- There is no need to create javadoc stubs because the aggregator generates one (for every artifact with a corresponding `sources.jar`.
- There is no need to recompute checksums because nothing is modified.

### DIFF
--- a/JenkinsJobs/Releng/publishToMaven.jenkinsfile
+++ b/JenkinsJobs/Releng/publishToMaven.jenkinsfile
@@ -15,7 +15,6 @@ pipeline {
 	}
 	environment {
 		// Filter out all the feature, test, and product IUs that are not published. Escape dots to match them literally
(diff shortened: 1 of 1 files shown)


merks:
Has no build impact.

HannesWell:
Thanks  a lot @merks for this work. It's really great to see how the publication to Maven-Central becomes easier and easier!

Regarding the points you mentioned I just wanted to record that they are not necessary _anymore_, due your enhancements in
- https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/pull/3034

respectively in the CBI aggregator.

PR #3036, opened 2025-05-08
Remove now unused POM-enrichment for publication to Maven-Central

HannesWell:
And downgrade the JDK used in the publishToMaven Jenkins pipeline as we don't need the latest Java-22/23 features anymore, i.e. the ability to launch multi-file source-code programs.

Follow-up on
- https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/pull/3035

This is currently a draft, because I want to test this at least once before submission.


merks (approved):
Very good.

HannesWell:
Thanks.
Just tested it and now it worked well. Submitting immediately as the verification build doesn't test this.

PR #3042, opened 2025-05-12
[Build] Run mavenBuild-workflow with CMD shell on Windows

HannesWell:
As currently some P2 tests fail and the executing shell might be the cause.

Follow-up on:
- https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/pull/2998


HannesWell:
Successfully tested in
- https://github.com/eclipse-equinox/p2/pull/825

Jenkins build failure is unrelated.

PR #3046, opened 2025-05-14
Add CDT core native API bundle for compilation

laeubi:
This is required for 

- https://github.com/eclipse-platform/eclipse.platform/pull/1762

currently it contains the cdt-main, but probably should reference the M2 repo (where relevant API was added) or a future CDT release.

Even though the bundle is **named cdt** it actually depends on nothing than JNA, therefore it can be used without really depend on CDT releases like any other API bundle (we just can't depend on new API if such would be ever added).

In a product one then would need to include the terminal feature (that includes the API and native fragments) to actually use the terminal console feature of platform.

FYI @merks @jonahgraham @akurtakov @stbischof

### DIFF
--- a/eclipse.platform.releng.prereqs.sdk/eclipse-sdk-prereqs.target
+++ b/eclipse.platform.releng.prereqs.sdk/eclipse-sdk-prereqs.target
@@ -68,6 +68,16 @@
       <unit id="org.eclipse.license.feature.group" version="2.0.2.v20181016-2210"/>
       <repository location="https://download.eclipse.org/cbi/updates/license/2.0.2.v20181016-2210"/>
     </location>
+    
+    <!--  
+      We consume the bundle form CDT unless we can fully migrate it to platform, this is only used for compilation
+      If one wants to use the terminal console feature in a product (e.g. EPP) one needs to include the CDT terminal support,
+      together with bundle org.eclipse.debug.terminal.
+     -->
+    <location includeAllPlatforms="true" includeConfigurePhase="false" includeMode="slicer" includeSource="true" type="InstallableUnit">
+       <repository location="https://download.eclipse.org/tools/cdt/releases/12.1/cdt-12.1.0/"/>
+       <unit id="org.eclipse.cdt.core.native" />
+    </location>
 
     <!-- uncomment 'eclipse_home' location, with text editor, for use in Eclipse IDE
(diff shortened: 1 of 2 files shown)


laeubi:
> currently it contains the cdt-main, but probably should reference the M2 repo (where relevant API was added) or a future CDT release.

Now changed it to RC2 (what is what CDT currently contributes to simrel at the moment).

laeubi:
Adjusted with final CDT release.

akurtakov:
I guess this is first step before moving it to platform so CDT doesn't have to become a -1 project ? @jonahgraham Are you fine with releasing CDT 1 week before each Platform M/RC ?

laeubi:
As long as we do not add new API (or more formally use that new API at platform), one won't need to update this in an way, this is more like Junit or any other dependecy, we use it and might update it when a new version become available, but it is not mandatory to update.

jonahgraham:
I agree with what @laeubi says - these things haven't changed in a long time. And I am happy to accommodate @akurtakov request as follows: If there are any changes that may affect Platforms consumption of these bundles then CDT will publish as a -1 for that change. 

xref new releasing documentation for CDT in https://github.com/eclipse-cdt/cdt/pull/1190

chirontt:
It looks like there's no terminal native fragment for Linux `riscv64` in CDT. Yet the Eclipse SDKs (and EPP too) have builds for Linux `riscv64` platform.

Is there a plan to add this `riscv64` native fragment soon?

laeubi:
@chirontt as mentioned by @merks contributions in this area has to go to CDT for the moment. In general if the usual Terminal View in the EPP works it should work here as well.

Beside that there are two ways:

1. As you mentioned, one might create a new native fragment, that would probably require some (cross) compilation and some setup to actually publish this new project
2. One could use JNA for the relevant parts and make the code independent from an actual binary, this would be the best solution as it would cover other architectures as well and would allow us to easier migrate the code to platform directly. Sadly there is no tool I know about at the moment to automate such process.

laeubi:
I now added a short introduction here:

https://github.com/eclipse-platform/www.eclipse.org-eclipse/blob/master/news/4.37/platform.md

jonahgraham:
> Perhaps I should whip up a PR for such new terminal native fragment for `riscv64`...

Yes please. See https://github.com/eclipse-cdt/cdt/issues/980

PR #3048, opened 2025-05-16
Touch org.eclipse.sdk.tests

merks:


merks:
The I-build is broken and this is the only way to fix it, so best to run the I-build again.

Issue #3070, opened 2025-05-29
Preparation work for 4.37(2025-09) and open master for development


MohananRahul:
This preparation work involves the following tasks. For previous issue please refer to eclipse-platform/eclipse.platform.releng.aggregator#2849 

- [x] Create R4_36_maintenance branch
- [x] eclipse-platform/eclipse.platform.releng.aggregator#3073
- [x] eclipse-platform/eclipse.platform.releng.aggregator#3075
- [x] eclipse-platform/eclipse.platform.swt#2192
- [x] eclipse-platform/eclipse.platform.releng.aggregator#3077
- [x] Create 4.37-I-builds repo
- [x] #3088
     - [x] Update product version number to 4.37 across build scripts
     - [x] Move previous version to 4.36RC2 across build scripts
- [x] #3089
- [x] Clean forceQualifierUpdate files for doc bundles
- [x] #3085
- [x] Create new I-build job for 4.37 release
- [x] Update check composites script to verify 4.37 repositories
- [x] #3090
- [x] #3093
- [x] #3114
- [x] Update build calendar for platform 4.37 release
- [x] Update generic repos I-builds, Y-builds, P-builds to point to 4.37 repos

HannesWell:
@MohananRahul when exactly do you plan to conduct this preparatory work?
I'm asking because I'm currently working on a Jenkins pipeline that automates and combines at least the following steps
> - POM and product version changes for 4.37 release
> - Update product version number to 4.37 across build scripts
> - Move previous version to 4.36RC2 across build scripts
> - Version bumps for 4.37 stream

maybe even for
> - Update version number in Mac's Eclipse.app for 4.37

It should make the tedious version bumps in all submodules obsolete because it improves the timings, and ensures the new parent-pom is already deployed so that the automated version-bump workflow can already do the heavy lifting.

It would be nice to use that pipeline immediately, but at the same time I don't want to delay the re-opening of the master (a lot).

MohananRahul:
> [@MohananRahul](https://github.com/MohananRahul) when exactly do you plan to conduct this preparatory work? I'm asking because I'm currently working on a Jenkins pipeline that automates and combines at least the following steps

Will soon after promoting RC2. 

> > * POM and product version changes for 4.37 release
> > * Update product version number to 4.37 across build scripts
> > * Move previous version to 4.36RC2 across build scripts
> > * Version bumps for 4.37 stream

[...]

PR #3094, opened 2025-06-04
[Build] Update snapshots in Version-Increment workflow

HannesWell:
This ensures that the workflow for automated version increments always operates on the latest content from the I-builds repository and discards cached data.
At least that's my hope and that worked locally for me. Let's see if it also helps in the CI.


HannesWell:
At least in https://github.com/eclipse-platform/eclipse.platform.swt/pull/2198 this seems to have helped to get the fresh composite content from the EF download-server, after latest I-build was marked unstable so that the 4.37-I-builds composite only contained the reference to the previous 4.36-RC2 (see also https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/issues/3070#issuecomment-2940680659).

In general I have the impression that the download-server caches returned results for quite a long time.

merks:
> In general I have the impression that the download-server caches returned results for quite a long time.

https://gitlab.eclipse.org/eclipsefdn/helpdesk/-/issues/6088

PR #3100, opened 2025-06-06
Eliminate all source feature dependencies from org.eclipse.sdk

merks:


iloveeclipse:
Ed, I'm not fully understand the implication of your proposal. 
Does it mean, SDK will be built without sources?

merks:
Also related:

https://github.com/eclipse-pde/eclipse.pde/pull/1806

HannesWell:
Thanks Ed for this change, it seems to be a follow-up on https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/pull/2777#issuecomment-2613214044 and will also enabled further clean-up of the SDK target.

I want to add, that in a target-file one can also specify `includeSource="true"` for `IU`-locations to include sources automatically, if they are available.

Furthermore PDE is about to get a new marker that tools like Tycho can use to add all sources in a feature automatically:
- https://github.com/eclipse-pde/eclipse.pde/issues/1613

> And I will strongly assert that these things are almost completely useless. Because who would want to install such a feature in product. Who would want to add a dependency to such a feature anywhere? No one...

Agree. If we don't use the source-features at all, we could also stop generating them in the build. On the other hand some might have referenced them explicitly and might be worried. They could just remove these references, but it's also not much work for us to provide source features.

PR #3107, opened 2025-06-07
Remove explicit inclusion of source-features in Eclipse SDK repository

HannesWell:
They are automatically included by Tycho in the build:
https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/blob/b8d3d2bd0fe900443277bd97654a2ceb60dd843e/eclipse.platform.releng.tychoeclipsebuilder/eclipse.platform.repository/pom.xml#L52


HannesWell:
It turned out that actually source features are not automatically added to a p2-repo and this was submitted too early.
The removal of source-features is reverted in
- https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/pull/3112

PR #3119, opened 2025-06-09
Add a manifest cleanup profile and workflow option

laeubi:
See:
- https://github.com/eclipse-tycho/tycho/pull/5037


laeubi:
Tested with:

https://github.com/laeubi/equinox/pull/55/files#diff-98e2e2939a98231d03432cd076fd94aa6396649b18bebff255b384f91134567cR17

PR #3130, opened 2025-06-14
Add org.eclipse.debug.terminal to be part of the repository

laeubi:
`org.eclipse.debug.terminal` is currently not meant to be referenced in any feature but to be consumed for example by an EPP. To do so we need to have it part of the update-site.


merks (commented):
👍

merks (approved):

PR #3136, opened 2025-06-17
Warn if workflow for automated version increments cannot modify PR

HannesWell:
It it's a constant source of confusion that the commit with version increments is not pushed to the PR, in case edits by maintainers are disabled or the PR is created from an organization repository (which makes edits in general impossible).
This now adds a warning for that case to at least explain contributors the situation.

FYI @HeikoKlare and @fedejeanne, as you or your colleagues have encountered this problems regularly in the past.


HannesWell:
The verification build doesn't test this. Therefore I submit this immediately so that it can be used sooner.

laeubi:
By the way you can use "Warning blocks" if you want:

https://github.com/orgs/community/discussions/16925

PR #3138, opened 2025-06-18
Silence deprecation warnings for junit jars

akurtakov:
API stability for JUnit is not guaranteed by Eclipse project thus it shouldn't be included in reports like
https://download.eclipse.org/eclipse/downloads/drops4/I20250617-1800/apitools/deprecation/apideprecation.html


akurtakov:
These reports are generated only by I-builds thus merging directly.

PR #3078, opened 2025-06-01
[Build] Add Jenkins pipeline to improve release preparation automation

HannesWell:
This introduces a new Jenkins job `Prepare next Development Cycle` that carries out many of the preparation tasks and aims to further enhance the release preparation automation. This includes
- Updating the version of the Maven parent, all references to it and the Eclipse products
- Updating the release version across build scripts
- Updating the previous release version to the current Release-Candidate

These tasks are currently done partly manually and partly automatically, but through a mixture of the (hidden) Jenkins job to [create-milestones](https://ci.eclipse.org/releng/job/create-milestones/) and distributed GH workflows that perform the Maven version updates upon the creation of new milestones.
Having one central job simplifies the understanding of the preparation process.

This PR is just the first step and it's the goal to move more preparation steps into this Jenkins job in order to automate currently manually executed step or to reduce the number of involved Jenkins jobs. All of this also ensures that the steps are always executed in the right order.

When applying the version updates, the script 'getPreviousRelease.sh' is not considered as it seems to be unused and is intended to be removed soon.

Issue #3123, opened 2025-06-10
Remove 4.36 from 4.37-I-builds


merks:
To keep 

- https://download.eclipse.org/eclipse/updates/4.37-I-builds/

from being empty at the start of the cycle, we composed

- https://download.eclipse.org/eclipse/updates/4.36-I-builds/I20250528-1830

So currently it looks like this:

![Image](https://github.com/user-attachments/assets/dd8e8b68-8b97-4ffc-b209-a704fbc80740)

We should improve the processes such that 4.36 is removed as soon as a proper 4.37 I-build is added and should remove 4.36-I-builds as soon as possible, i.e., especially before the 4.36 I-builds are deleted.

FYI @MohananRahul

HannesWell:
I'm currently on it. It should be gone soon. I added it with a manually assembled prototype job and I'm currently using a modified version to remove it. If I fail to remove it this morning (I only have limited available), I'll continue this evening.
The most important part is that nobody else deletes the referenced 4.36-I-builds repo in the meantime.
I'll let you know once the reference is removed.

> We should improve the processes such that 4.36 is removed as soon as a proper 4.37 I-build is added and should remove 4.36-I-builds as soon as possible, i.e., especially before the 4.36 I-builds are deleted.

Yes, absolutely. I think it should be sufficient to update https://ci.eclipse.org/releng/job/Cleanup/job/pruneDailyRepos/, to always just keep three (or whatever threshold is specified builds).

I think it's even worth to consider to implement a hard coded limit of references that are retained in the composite when a new build is added to the composite. Then when adding a new build an old one is immediately removed and we don't need an extra job to clean-up.
The current approach only has the advantage that, if a build is marked unstable later, that then the clean-up might not have happened and more repos remain in the composite. But since the limit is currently three (we could increase it to four if necessary) there should be sufficient back-up in the composite to not run empty. Unless we mark three consecutive builds unstable. But for that case we could also enhance the mark-unstable job to at least warn about it or by adding back older builds?

HannesWell:
> I think it's even worth to consider to implement a hard coded limit of references that are retained in the composite when a new build is added to the composite. Then when adding a new build an old one is immediately removed and we don't need an extra job to clean-up.

This is now implemented with
- https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/pull/3153

PR #3146, opened 2025-06-22
[Build] Create new stream repositories in release preparation pipeline

HannesWell:
Additionally add the previous release RC as first children immediately to the new I-build repository so that it isn't empty and the RC is used as baseline for the first builds of the new stream as also described in
- https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/issues/3070#issuecomment-2937128287

Follow-up on
- https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/pull/3078

This is currently a draft as it requires some final testing.


HannesWell:
Testing was successful and this is ready for submission.

@MohananRahul FYI, with this, when the 4.38 development cycle is prepared in late summer, this pipeline will also create the repositories
- https://download.eclipse.org/eclipse/updates/4.38
as empty composite
- https://download.eclipse.org/eclipse/updates/4.38-Y-builds/
as empty composite
- https://download.eclipse.org/eclipse/updates/4.38-I-builds/
with a reference to the then specified RC2 build of the 4.37 release as only child of the composite

So that the problem described in https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/issues/3070#issuecomment-2937128287 ff should not occur.
I have also updated the release documentation that the new stream repos don't have to be created in a separate step and have removed the definition of the corresponding job.

As next in this effort I'll update the pruning of the composite repository so that https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/issues/3123 does not happen anymore and the reference to the previous RC2 is automatically removed when the first I-builds of the then new stream have succeed.

PR #3154, opened 2025-06-24
Remove now obsolete eclipse-platform-sources Maven project

HannesWell:
It seems now unused (I couldn't find any reference to it) and maintaining it further complicates the RelEng.
So just removing it seems to be the path forward.
Last part of
- https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/issues/882


akurtakov (approved):

akurtakov:
As this PR only removes things, it can not create new license issues itself. Merging.

PR #3157, opened 2025-06-25
[Build] Use own cache in Version-Increment workflow

HannesWell:
As it was pointed out by @akurtakov, the version-check for SWT maybe fails because the used caches interfere with the caches of the verification workflows.
This change ensures that the version-check workflow has its own cache.


HannesWell:
The testing seemed to be successful:
- https://github.com/eclipse-platform/eclipse.platform.swt/pull/2266

Lets hope that it also works in practice.

FYI @fedejeanne and @HeikoKlare

PR #3159, opened 2025-06-26
Refine check if PR can be updated in version increments workflow

HannesWell:
Maintainers and consequently the organization's Eclipse Bot can always modify a PR from a branch in the same repository, even if edit's by maintainers are not (explicitly) enabled.

This avoids a false positive warning that a required version-increment cannot be pushed, if for example a Bot has created a PR from a branch within the target repository.

Follow-up on
- https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/pull/3136


HannesWell:
The verification build doesn't test this, thus submitting this immediately.

PR #3160, opened 2025-06-27
Do not include source features in SDK

akurtakov:
All the content should still be there as materialize-products has includeSources=true property set.


laeubi (approved):
Yes since a while we already include the sources anyways, so it is just shipping the feature itself... (so removing it will not remove any sources).

laeubi:
Its quite unlikley that the licenses make any problem for this change. The run says "Received an error response from ClearlyDefined." seems the service is currently down...

akurtakov:
I know that, also clearlydefined gives errors quite often and the previous run failed with few of our own plugins being wrongly flagged by the tool and I want to have iplab issues for them so they don't fail on other PRs.

akurtakov:
Anyway, merging and starting new I-build to verify state.

PR #3162, opened 2025-06-28
Improve pom files

akurtakov:
* Do not duplicate setting tycho-version to same constant
* Set the Equinox repository name proper.


akurtakov:
Failing licensechecks are project's own content - aka issue with the tool. Merging.

PR #3164, opened 2025-06-29
[Build] Add missing version updates in release preparation pipeline

HannesWell:
Follow-up on
- https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/pull/3078


HannesWell:

PR #3165, opened 2025-07-01
Add dedicated job to modify a p2 composite repository

HannesWell:
- unifies the code to modify a composite p2-repository, which was previously written in multiple locations
- the new job performs all operations locally within the job's workspace and only copies the resulting files to the download-server. This therefore eliminates the remaining executions of Eclipse directly on the download-server.
- A dedicated job, prevents concurrent modifications of the same repository
- A dedicated job, allows run a specific modification manually if necessary (for out of order trouble-shooting).

Follow-up on
- https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/pull/3153


HannesWell:
All my testing was successful and as the verification build doesn't test this, I submit this immediately so that the next I-build can test it in production.

PR #3176, opened 2025-07-09
Do not include tips source feature

akurtakov:
Source bundles are handled just fine by Tycho and PDE so this is useless artifact.


merks:
Thanks @akurtakov

PR #3179, opened 2025-07-10
[RelEng] Migrate Maven publication to new Central Portal API

HannesWell:
This is currently a draft as the tokens are not yet available for all three namespaces (`platform`, `jdt` and `pde`).

### DIFF
--- a/JenkinsJobs/Releng/FOLDER.groovy
+++ b/JenkinsJobs/Releng/FOLDER.groovy
@@ -15,7 +15,7 @@ then deploys the artifacts for <code>Eclipse-Platform</code>, <code>JDT</code>, 
 Snapshots are deployed to <a href="https://repo.eclipse.org/content/repositories/eclipse-snapshots/">https://repo.eclipse.org/content/repositories/eclipse-snapshots/</a>.
 </li>
 <li>
-Releases are deployed to <a href="https://repo1.maven.org/maven2/org/eclipse/">Maven central</a> by deploying to a <a href="https://oss.sonatype.org/#stagingRepositories">staging repository</a>.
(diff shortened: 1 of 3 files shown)


HannesWell:
@merks FYI this pipeline passed one first run, but only for the platform namespace (for the other two namespaces the tokens are missing).

@elsazac can you please check the new Central Portal if there has been an deployment for the o.e.platform namespace and check if it's correct and discard it so that it is ** !NOT! RELEASED**!
With the old OSSRH it was the `Drop` button as you described in https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/pull/2730#discussion_r1904083740, but I don't know how this works with the new Portal. If you are in doubt, please just do nothing and let's conduct some research. Or maybe Ed can tell from his experience with EMF?
Thanks in advance.

elsazac:
But there’s also an option to publish components by selecting files from the local file system in the OS.



Screenshot 2025-07-11 at 7 03 38 PM

elsazac:
And then just this also in my notifications tab, This is regarding the platform namespace; Is this what you were looking for ?? apart from that there aren't much options available too.

Screenshot 2025-07-11 at 7 09 39 PM

HannesWell:
Thank you @elsazac for checking and providing all these information. That makes me confident that the deployment actually worked.

> Are we sure (or expecting) that the problem of only being able to see a deployment made with one's own ID has been addressed already?

Good point, that's probably the reasons why Elsa cannot see the full deployment.

In the meantime, I'll just continue to use the Publisher API to manage (inspect and drop) the deployment:
- https://central.sonatype.org/publish/publish-portal-api/#verify-status-of-the-deployment

In general we could also built that into a Jenkins job, but that would mean that any Committer can publish a release (which is technically already the case today, but one has to know about the mentioned API and how to use it).

PR #3181, opened 2025-07-11
[RelEng] Specify P2 repository to publish to Maven as job parameter

HannesWell:
instead of changing the `SDK4Mvn.aggr` files to just run the publication job subsequently and then changing it again to the next release cycle's I-build repository.
This avoids the described changes and simplifies the release process.

### DIFF
--- a/JenkinsJobs/Builds/build.jenkinsfile
+++ b/JenkinsJobs/Builds/build.jenkinsfile
@@ -298,10 +298,12 @@ spec:
 			steps {
 				script {
 					for (c in BUILD.testConfigurations) {
-						build job: "${BUILD.testJobFolder}/${BUILD.testPrefix}-${c.os}-${c.arch}-java${c.javaVersion}", parameters: [string(name: 'buildId', value: "${BUILD_IID}")], wait: false
+						build job: "${BUILD.testJobFolder}/${BUILD.testPrefix}-${c.os}-${c.arch}-java${c.javaVersion}", wait: false, parameters: [
+							string(name: 'buildId', value: "${BUILD_IID}")
+						]
 					}
 				}
-				build job: 'SmokeTests/Start-smoke-tests', parameters: [string(name: 'buildId', value: "${BUILD_IID}")], wait: false
+				build job: 'SmokeTests/Start-smoke-tests', wait: false, parameters: [string(name: 'buildId', value: "${BUILD_IID}")]
 			}
 		}
 		stage('Trigger publication to Maven snapshots repo') {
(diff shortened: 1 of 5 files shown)


HannesWell:
@elsazac this makes changes like https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/pull/3105 unnecessary during the release process. You can just specify the p2-repository to publish by its URL as parameter when triggering the Maven-publish job.
I have also adjusted the documentation and hope this is sufficient.

HannesWell:
@stephan-herrmann, this also would make it possible to deploy Y-build snapshots in a simple way, as for example requested in https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/issues/2625#issuecomment-2671168267.
Probably the only additionally required change would be specifying the target (snapshot) maven repository, as I assume that Y-build snapshots should go to another Maven-repository.

Would you be in favor of such deployment? And if yes, where should it go? The JDT eclipse-repo?

stephan-herrmann:
> @stephan-herrmann, this also would make it possible to deploy Y-build snapshots in a simple way, as for example requested in [#2625 (comment)](https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/issues/2625#issuecomment-2671168267). Probably the only additionally required change would be specifying the target (snapshot) maven repository, as I assume that Y-build snapshots should go to another Maven-repository.
> 
> Would you be in favor of such deployment? 

I have no strong opinion either way. 

IMHO deploying the entire Y-build would be too much. In the comment you quote only ecj was requested. 

> And if yes, where should it go? The JDT eclipse-repo?

I don't even know of any such JDT eclipse-repo :-/  -- but it might indeed be a good idea to keep Y-build artifacts separate, just in case there's an unintended overlap of versions between master an the beta branch.

mickaelistria:
> I don't even know of any such JDT eclipse-repo :-/

See https://github.com/eclipse-jdt/eclipse.jdt.core/blob/master/pom.xml#L52 , a p2 repo is created for jdt.core if built with `-Pp2-repo` profile.

stephan-herrmann:
> > I don't even know of any such JDT eclipse-repo :-/
> 
> See https://github.com/eclipse-jdt/eclipse.jdt.core/blob/master/pom.xml#L52 , a p2 repo is created for jdt.core if built with `-Pp2-repo` profile.

This seems to answer a different question. @HannesWell  asked "where should it go?", but the pom you reference doesn't show a URL where JDT can deploy maven snapshot artifacts for users to consume.

PR #3183, opened 2025-07-13
Enable "Remove unused private members" cleanups

laeubi:
Currently some other cleanups can lead to a private method become unused, what then results in a warning and maybe build failure.

This now enables the cleanups to remove unused private methods/fields/constructors to avoid that situation.


laeubi:
Tested locally and build will not verify this.

PR #3184, opened 2025-07-14
Enable Manifest Cleanup "removeUnusedDependencies"

laeubi:
I have now tested this with Equinox repo (that is the only one where it is currently enabled) and beside some edge cases it seem to work quite fine now.

The remaining issues are better addressed on concrete code samples that allow discussion of possible mitigates directly at the PR.


laeubi:
The build will not test this.

PR #3187, opened 2025-07-16
[Build] Restore compressed content.xml for platform repository

HannesWell:
Include the additional IU `org.eclipse.equinox.executable` directly in the platform P2-repository via the defining `category.xml`. This avoids the post-processing of the raw `content.xml` file using an XSL transformation and consequently ensures the `content.xml` is only provided in compressed form as 'content.jar' and `content.xml.xz`.
Due to the previous post-processing the `content.xml` was only provided uncompressed, which increases the load on the download-server, slows down metadata downloads and also complicates the RelEng process.

Fixes https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/issues/1912


eclipse.platform.releng.tychoeclipsebuilder/eclipse.platform.repository/category.xml
```diff
@@ -19,4 +19,9 @@
    <bundle id="jakarta.inject.jakarta.inject-api" version="1.0.5"/>
    <bundle id="org.eclipse.equinox.slf4j"/>
    <bundle id="org.eclipse.debug.terminal"/>
+   <iu>
+     <query> 
+       <expression type="match">id == 'org.eclipse.equinox.executable'</expression>
+     </query>
+   </iu>
```
HannesWell, line 26: This is an alternative approach to explicitly include the `org.eclipse.equinox.executable` generated by Equinox.
I copied this approach (and simplified it) from
https://github.com/eclipse-m2e/m2e-core/blob/f64b9290f2e3dd0c7b6f14a185edc36286590933/org.eclipse.m2e.repository/category.xml#L21-L30

merks: That's certainly MUCH simpler!

merks:
That’s really great. 👍

PR #3191, opened 2025-07-18
[Build] Restore compressed content.xml for platform repository

HannesWell:
Include the additional IU 'org.eclipse.equinox.executable' directly in the platform P2-repository via the defining 'category.xml'. This avoids the post-processing of the raw 'content.xml' file using an XSL transformation and consequently ensures the 'content.xml' is only provided in compressed form as 'content.jar' and 'content.xml.xz'. Due to the previous post-processing the content.xml was only provided uncompressed, which increases the load on the download-server, slows down metadata downloads and also complicates the RelEng process.

Fixes https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/issues/1912

This was originally introduced via
- https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/pull/3187

and reverted due to a regression in I-builds via
- https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/pull/3189

The change is introduced again unchanged as the cause of that regression is fixed via
- https://github.com/eclipse-equinox/equinox/pull/1089

This reverts commit d4d05423bd2f76fa721006d35bf576bd8f833c77.


HannesWell:
As https://github.com/eclipse-equinox/equinox/pull/1089 is submitted and my local testing successful this is now ready to be applied again.
I'll start an I-build immediately to verify it works as desired in that case too.

HannesWell:
> I'll start an I-build immediately to verify it works as desired in that case too.

The build https://download.eclipse.org/eclipse/downloads/drops4/I20250718-1800/ succeeded, lets see if the tests show any issues.

PR #3202, opened 2025-07-25
Allow to configure branch for "Automatic Code Cleanups"

ruspl-afed:
Currently `master` branch is hardcoded and it is not possible to apply this very useful workflow to a repositories where branches are named differently.
This change adds `branch` configuration parameter


laeubi:
By the way if one uses a manual trigger to run the workflow one can even select a branch, so it would maybe good to even read it from the inputs first and then fall back to the configured default (e.g. timed execution), but that's another story.

PR #3203, opened 2025-07-25
Publish o.e.sdk.examplese not source

akurtakov:
Sources will be auto published without the use of deprecated (at least by tycho) source feature usage.


akurtakov:
Licensecheck misdetects junit sampleproject as external project. Merging.

Issue #3201, opened 2025-07-24
Release P2 repositories miss p2.mirrorsURL property since 4.28


HannesWell:
Since Eclipse 4.28 respectively 2023-06 the `p2.mirrorsURL` is not defined for the release repositories anymore.
I assume that therefore mirrors don't work anymore for all release p2-repositories since then. Further I assume this reduces download performance and increases the load for the EF servers.

The last repository to contain this property is https://download.eclipse.org/eclipse/updates/4.27/R-4.27-202303020300/:
`<property name="p2.mirrorsURL" value="https://www.eclipse.org/downloads/download.php?format=xml&file=/eclipse/updates/4.27/R-4.27-202303020300"/>`

This is probably because the `Rename and Promote`/`Promote build` job, which creates the final release repositories did run with Java-11 up until I migrated it to Jenkins pipeline recently (and removed that JDK because I assumed it's unused as it's not sufficient for some time):
https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/blob/40720432177635a5a7a14369bf6660589a35f5e5/JenkinsJobs/Releng/renameAndPromote.groovy#L33

That job launches an Eclipse installation to add said property, which fails (unfortunately silently) since then because since 4.28 a 17 JRE is required.

akurtakov:
Tycho can inject this Metadata directly. See https://github.com/eclipse-linuxtools/org.eclipse.linuxtools/blob/master/releng%2Forg.eclipse.linuxtools.releng-site%2Fpom.xml#L53 .

merks:
> But in general I wonder why didn't received any complains due to that, or at least I didn't notice it.

Who would complain? The webmaster?  It's not generally obvious why the server is taxed more than necessary.  The users?  Do they even notice where things are coming from to wonder why it's not coming from a mirror?

> I disabled all mirror processing.

I guess you mean in Tycho?

> Beside that is only the update repo affected or also the simrel? 

SimRel does specify a mirrors URLs in each simple repository, e.g.,

<img width="1127" height="22" alt="Image" src="https://github.com/user-attachments/assets/0f4fc4b9-7bb0-4c5c-825b-0e00a8712ccd" />

But the Platform does add itself to the preference:

<img width="532" height="82" alt="Image" src="https://github.com/user-attachments/assets/42519a11-498a-4542-a781-b428c8c7e8a6" />

[...]

PR #3204, opened 2025-07-25
[RelEng] Simplify creation of release P2 repo and restore its mirrorsURL

HannesWell:
Since 4.28 respectively 2023-06 the property 'p2.mirrorsURL' is not added to release repositories anymore, which effectively disables the use of mirrors.
This happened accidentally because the JDKs installed at the Jenkins agents running the release process became insufficient then.

In order to simplify the release process, the P2-repositories produced by I-builds alredy have their final name set.

Fixes https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/issues/3201

This is currently a draft as it has to be tested first.

### DIFF
--- a/eclipse.platform.releng.tychoeclipsebuilder/eclipse.platform.repository/pom.xml
+++ b/eclipse.platform.releng.tychoeclipsebuilder/eclipse.platform.repository/pom.xml
@@ -50,6 +50,7 @@
         <configuration>
           <includeAllDependencies>true</includeAllDependencies>
           <includeAllSources>true</includeAllSources>
+          <repositoryName>Eclipse ${releaseVersion}</repositoryName>
         </configuration>
       </plugin>
       <plugin>
--- a/JenkinsJobs/Releng/promoteBuild.jenkinsfile
+++ b/JenkinsJobs/Releng/promoteBuild.jenkinsfile
@@ -11,9 +11,13 @@ pipeline {
 	environment {
 		HIDE_SITE = 'true'
 		// Download Server locations (would very seldom change)
-		BUILD_ROOT = '/home/data/httpd/download.eclipse.org'
-		EP_ECLIPSE_ROOT = "${BUILD_ROOT}/eclipse"
-		EP_EQUINOX_ROOT = "${BUILD_ROOT}/equinox"
+		EP_ROOT = '/home/data/httpd/download.eclipse.org'
+		EP_ECLIPSE_ROOT = "${EP_ROOT}/eclipse"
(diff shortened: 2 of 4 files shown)


eclipse.platform.releng.tychoeclipsebuilder/eclipse.platform.repository/pom.xml
```diff
@@ -50,6 +50,11 @@
         <configuration>
           <includeAllDependencies>true</includeAllDependencies>
           <includeAllSources>true</includeAllSources>
+          <repositoryName>Eclipse ${releaseVersion} repository (for ${releaseYear}-${releaseMonth})</repositoryName>
+          <extraArtifactRepositoryProperties>
+            <p2.statsURI>https://download.eclipse.org/stats</p2.statsURI>
```
HannesWell, line 55: Do we want to add this `p2.statsURI`?

laeubi:
The less tools, the better, so if we can use plain Tycho it would be preferable... If anything is missing there we can add it.

To account for different types of builds, one can have a profile that only enable the settings if a given ENV variable is set.

HannesWell:
> The less tools, the better, so if we can use plain Tycho it would be preferable... If anything is missing there we can add it.
> 
It would be great if this could be simplified, but currently I didn't found a better solution to add or just alter a property in an existing P2 repository. Neither with Tycho nor with the P2 ant tasks.
But if you know better, please let me know.
I'm not sure if it's worth just for this case to add an extra mojo to add a property to a repository.


> To account for different types of builds, one can have a profile that only enable the settings if a given ENV variable is set.

The problem is that we don't have a final release build but just promote a selected I-build repository (which is basically copying it and adding that property).

HannesWell:
I know all of them and tried to use them for this task, but didn't find a solution.

> and [tycho-p2-repository:fix-artifacts-metadata](https://tycho.eclipseprojects.io/doc/latest/tycho-p2-repository-plugin/fix-artifacts-metadata-mojo.html) modify the metadata repo (and can recompress it as well afterwards).

If you think it's suitable I can look into creating a `modify-artifacts-metadata` or alike mojo for the `tycho-p2-repository-plugin` plugin. Which would then support modifying the name adding or removing properties. Later more modifications could be implemented, if desired (thinking about the extra IU of the equinox.executable feature were about to eliminate).
One question would be if it should just be for `artifacts` or also for `content` repos? Or the mojo could be generalized to be `modify-metadata` and have an option to select `artifacts`, `content` or `both` (by default).
And in order to save bandwidth, the source could be any remote and only the resulting artifact/content.jar/xml.xz would be outputted locally (but could also overwrite a local repository).
WDYT?

PR #3222, opened 2025-08-13
[Build] Set up I-build tests for Java-25 EA

HannesWell:
As discussed in today's PMC meeting this adds a new I-build test configuration for Java-25 EA. For Y-builds such a configuration already exists.

If we consider constant EA testing important, we could also have a permanent Java-_EA_ configuration using the `openjdk-ea-latest` JDK tool:
https://github.com/eclipse-cbi/jiro/blob/faaca45c9cbaaeaa59b0433c41d02ae9e639badd/templates/jenkins/partials/tools-jdk.hbs#L5-L6

But I think it's also fine to just the configuration for a new Java already when it's EA. The EF-infra team usually adds the corresponding tools in time (and if not, we can ask them).
Another option would be adding just a Smoke test for a Java-EA configuration.

@akurtakov or anybody else, what do you think?

### DIFF
--- a/JenkinsJobs/AutomatedTests/I_unit_tests.groovy
+++ b/JenkinsJobs/AutomatedTests/I_unit_tests.groovy
@@ -3,7 +3,8 @@ def STREAMS = config.Streams
 
 def TEST_CONFIGURATIONS = [
 	[os: 'linux' , ws:'gtk'  , arch: 'x86_64' , javaVersion: 21, agentLabel: 'ubuntu-2404'        , javaHome: "tool(type:'jdk', name:'temurin-jdk21-latest')" ],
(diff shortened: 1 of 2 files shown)


iloveeclipse:
> Using the openjdk-ea-latest JDK tool sounds best to me . The less custom configs, the less misalignments.

Fully agree.

HannesWell:
> But I think it's also fine to just the configuration for a new Java already when it's EA. The EF-infra team usually adds the corresponding tools in time (and if not, we can ask them).

On the other hand just adding the new test-configuration (long) before the release would probably lead to less changes, respectively less overlap because I assume there is certainly some time where `openjdk-ea-latest` and e.g. `openjdk-jdk25-latest` point to the same JDK.

PR #3225, opened 2025-08-16
[RelEng] Inline promoteSites script and streamline promotion work

HannesWell:
Perform the copy and file-renaming directly on the download/storage-server instead of coping all files to the Jenkins build-agent and back to the server, just to rename them.


HannesWell:
@MohananRahul I tested this change as good as I can, but to be sure everything works fine, please let me know when you (have) run the `Promote Build` job for the final release promotion so that I can check it in detail before the it's finally published.

PR #3232, opened 2025-08-20
New and Noteworthy 4.37 (Sample Preview)

elsazac:
Note: This is a sample PR that demonstrates how the Help page is rendered and is not the final PR.

@merks 
The New and Noteworthy entries in the news repo were recently migrated from HTML to Markdown as you know. Earlier, we used a script to pull the HTML content directly into the help pages, but since the format has changed, the old scripts no longer work.

To address this, I took a copy of the Markdown files, encoded and converted them into HTML, and then replaced the corresponding files in the repo were the help page is rendered (releng repo). I also updated the references in the toc.xml files. Simply embedding Markdown-encoded HTML wasn’t rendering consistently (issues with fonts, colors, etc.), so I also adjusted the styling for proper display.

With these changes, the rendering is now working as expected. The only issue I’m noticing is that some links in the help pages open as blank pages. Is this a known issue? I think it could be fixed by adjusting the link attributes to open in a new tab using `href` along with `target="_blank"`. Let me know your thoughts on that.

Also, the images folders in the respective directories have been refilled with the latest images and deleted the old images.

Having said all this, I want to confirm that the news repo where contributors add Markdown files remains untouched. These changes only affect how the Help page is rendered.

I’ve put all of these changes into a sample PR. This is not the final PR for the release; I understand that the New and Noteworthy updates may still be added to the news repo. The purpose of this PR is just to provide a preview of how the Help page would look after rendering.

Could you please test these changes locally and share your suggestions? If this approach looks fine to you, we can proceed with it for this release. Compared to the last release (where we only provided a link to the N&N pages), this rendering feels like a better option to me.


merks:
@akurtakov 

If we have "dead" content we probably should remove it, right?

merks:
There is news

https://eclipse.dev/eclipse/markdown/?f=news/4.37/pde.md

But I don't know what the plans are around the PDE docs I see here.  If they're useless, we should delete them, if not,  I'd expect to see the in the help system like I do when I debug launch.

Like Hannes, I also didn't know about plans to "copy/map" the online content into the help system docs.

elsazac:
> Like Hannes, I also didn't know about plans to "copy/map" the online content into the help system docs.

Rather than just pointing to a link, I feel it’s better to render the content directly in the help pages, as we used to in past releases. That approach feels more complete to me, but I’m fine going with what the majority decides. Let me know your thoughts on this.

elsazac:
If we want to render the help pages directly in the right-hand pane with the actual content (as we did in previous releases), it involves quite a bit of manual work since we have now fully shifted to Markdown. This PR was an attempt to preserve that style of rendering, but it comes with certain tasks, such as:

1.   wrapping the Markdown files with an HTML header
2.   adjusting links to open in a new tab
3.   resizing images 
4.   handling theme selection
5.   copying images into the platform.common 

These problems are all still solvable only, but they do require manual effort.

The main advantage of this approach is that we can continue to render the content directly within the help pages, providing a more complete view to the user.

That said, since the majority prefers simplifying the releng process, I'm fine going with the alternative approach ie, having a forward link that points to the corresponding md files.

merks:
It just seems like quite some work, work that's manual and error prone, work that happens very late in the cycle, and work that might be better invested elsewhere.  But that's just my **personal** opinion.

elsazac:
Closing the discussion here and Final PR for New and Noteworthy 4.37 created via https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/pull/3244

PR #3235, opened 2025-08-21
[RelEng] Keep up to ten builds of the Promotion/Publication jobs

HannesWell:
This way at least one release cycle is available in the history, as these jobs run at least six times per release (M1, M2, M3, RC1, RC2 and R/GA).

Also extend the timeout for the publication job, which runs usually at most 7min (for a release), to 45min. This is safety factor is similarly high to the promotion-job (which runs about 20min max and has a 120min timeout, as I don't want to have a promotion/publication aborted, just due to a slow network or busy Jenkins.


HannesWell:
This is a pure RelEng change that's intended to be used for tomorrows RC1 promotion. Therefore I'm submitting this now despite we are currently in a freeze-period.

PR #3246, opened 2025-08-25
[RelEng] Simplify/shorten 'sign-off page' link in the download page

HannesWell:
Makes the page content more compact and the promotion pipeline a little bit simpler.
The GH-issue style label was added in https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/pull/3225, but seems not to help that much.


HannesWell:
This is a pure RelEng change that's intended to be used for the RC2 promotion. Therefore I'm submitting this now despite we are currently in a freeze-period.

PR #3252, opened 2025-08-26
[RelEng] Simplify input processing in release preparation pipeline

HannesWell:
Unify the printing of processed/computed parameters with the other RelEng jobs (which is more compact).

Enhance processing of the `PREVIOUS_RELEASE_CANDIDATE_ID` parameter to
- handle RC2a/b/c... builds
- service versions


HannesWell:
This is a pure RelEng change that's intended to be used for the upcoming release preparation. Therefore I'm submitting this now despite we are currently in a freeze-period. Also the verification build doesn't test this, so I'm submitting this immediately.

PR #3254, opened 2025-08-26
[RelEng] Update p2-repos in products in release preparation pipeline

HannesWell:
This was for example done manually in
- https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/pull/3209

and as it's not documented it was forgotten in the initial release preparation and done later.
On the long run it would be great if we could just reference the corresponding variables here, like in other places (but I have to admit that I haven't tested if it's already possible).


HannesWell:
This is a pure RelEng change that's intended to be used for the upcoming release preparation. Therefore I'm submitting this now despite we are currently in a freeze-period.

PR #3255, opened 2025-08-26
[RelEng] Update build configuration in release preparation pipeline

HannesWell:
and simplify the definition and activation of the RC schedule: Update the cron-trigger to it's final value immediately. The extra builds are then triggered when the RC period arrives without activating that schedule manually.

This also simplifies the cron-triggers by just using the complement for the RC schedule (at least for I-builds) and by continuing the RC schedule for the sign-off Thursday. Nowadays we don't have to worry too much about extra triggers as an I/Y build is skipped when nothing has changed:
- https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/pull/2706,

From a RelEng POV this has the advantage that the RC schedule doesn't have to be activated by an extra config/code-change but is just specified as necessary from the beginning.

FYI @akurtakov, @MohananRahul
and @jarthana, @mpalat, @stephan-herrmann as this also affects JDT's Y-build.
Are you all fine with this?

This is currently a draft as the update of the Y-build is currently missing and I have to check if the new I/Y-build jobs can be created in disabled state (and be activated once everything is ready).

### DIFF
--- a/JenkinsJobs/Builds/FOLDER.groovy
+++ b/JenkinsJobs/Builds/FOLDER.groovy
@@ -14,18 +14,11 @@ for (STREAM in config.Streams){
 				triggers {
 					cron {
 						spec('''TZ=America/Toronto
-# format: Minute Hour Day Month Day of the week (0-7)
-
+# Format: Minute Hour Day Month Day-of-week (1-7)
 # - - - Integration Eclipse SDK builds - - - 
-# 2025-09 Release Schedule
-# Normal : 6 PM every day (11/6 - 27/8)
-# 0 18 * * *
-
-
-# RC Schedule 
-# Post M3, no nightlies, I-builds only. (Be sure to "turn off" for tests and sign off days)
-# 0 6 15-27 8 5-7,1-3
-# 0 18 15-27 8 5-7,1-3
+# Schedule: 6 PM every day until end of RC2
+0 18 * 5-7 *
+0 18 1-27 8 *
 ''')
 					}
 				}
--- a/JenkinsJobs/Releng/prepareNextDevCycle.jenkinsfile
+++ b/JenkinsJobs/Releng/prepareNextDevCycle.jenkinsfile
@@ -58,6 +58,17 @@ pipeline {
 					assignEnvVariable('NEXT_RELEASE_MONTH', String.format("%02d", gaDate.monthValue))
 					assignEnvVariable('NEXT_RELEASE_NAME', "${NEXT_RELEASE_YEAR}-${NEXT_RELEASE_MONTH}")
 					assignEnvVariable('MAINTENANCE_BRANCH', "R${PREVIOUS_RELEASE_VERSION_MAJOR}_${PREVIOUS_RELEASE_VERSION_MINOR}_maintenance")
+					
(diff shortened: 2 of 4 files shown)


HannesWell:
> Then for a moment I was thinking if #2706 is actually good for Y-builds (which have lower frequency of actual changes). In the light of intermittent test failures I actually appreciate when more builds are running on the same input to give an idea if the failure indicates a systematic bug.

That feature was introduced mainly for RC-schedule times where we have more frequent I-builds but at the same time have a great chance of no changes (if no bug is found...). or times around new-year where only little change happens.

As Y-builds by defaults runs less often, it would probably not be unreasonable to turn that off for Y-builds. OTOH, the last time a Y-build was skipped due to no changes was in March (probably short before the Java-24 release?):
https://www.eclipse.org/lists/jdt-dev/2025/Mar/index.html
At the same time this can be an argument pro disabling build-skipping for Y-builds.
However, let's discuss that separately.

> OTOH, I don't want to be a resource hog, so should the above scenario be resolved by touching some `forceQualifierUpdate.txt` file or similar?

Does enforcing a new qualifier really helps with test-failures? I assumed they only helped with comparator errors.

akurtakov:
For the record - 4.37 RC1 I-builds happen only  once per day and 2 I-builds per day for RC2 . As RC2 is too late in the cycle for significant work having 2 builds doesn't make sense and no one complaint about the single build per day for RC1 - I propose to remove this complication and have single I-build per day at all time even in RC phase.

HannesWell:
This PR has been updated with the suggestion from my previous message.
All my testing so far was successful.

This is a pure RelEng change that's intended to be used for the upcoming release preparation. Therefore I'm submitting this now despite we are currently in a freeze-period. And as the verification build doesn't test this, I'm submitting this immediately.

stephan-herrmann:
We have a build failure in https://ci.eclipse.org/jdt/job/eclipse.jdt.core-run.javac-25/26/console which tells me that that composite repo https://download.eclipse.org/eclipse/updates/Y-builds already points to 4.38-Y-builds, which IMHO doesn't make sense before Java 25 GA / final freeze of BETA_JAVA25.

Is there any connection to this PR?

Will Y-builds start to upload to 4.38-Y-builds already?

As per https://github.com/eclipse-jdt/eclipse.jdt.core/pull/4366 I'm preparing to tie the branch build to the more specific 4.37-build repo, which assumes that this will still be populated by current Y-builds. Is this assumption correct?

stephan-herrmann:
> Will Y-builds start to upload to 4.38-Y-builds already?

no, today's Y-build still uploaded to 4.37-Y-builds.

HannesWell:
> We have a build failure in https://ci.eclipse.org/jdt/job/eclipse.jdt.core-run.javac-25/26/console which tells me that that composite repo https://download.eclipse.org/eclipse/updates/Y-builds already points to 4.38-Y-builds, which IMHO doesn't make sense before Java 25 GA / final freeze of BETA_JAVA25.
> 
> Is there any connection to this PR?

Yes that was intentionally changed by this PR, but after your remarks, I think that this point in time might be too early.

> There's more to it, over to #3273

Thanks, I'll continue there.

PR #3256, opened 2025-08-27
docs/help: fix DTD for P2 feature.xml

chrisrueger:
Replace pipe with comma, since pipe means "or" / choice. But that would mean you can only use one of the elements. Comma means that feature must contain the following elements, which is what we want here

I found an archived site which uses comma: https://archive.eclipse.org/eclipse/downloads/documentation/2.0/html/plugins/org.eclipse.platform.doc.isv/reference/misc/feature_archive.html


merks:
@akurtakov 

I believe this change is an accurate correction according to `Elements with Children (sequences)` here:

https://www.w3schools.com/xml/xml_dtd_elements.asp

Given it's a doc only change, should we merge it now?

chrisrueger:
Thanks @merks 
I stumbled over this while working on a [testcase the P2Export in bnd](https://github.com/bndtools/bnd/commit/58db435d736674d1863995fc4ecb95632f609e21) and I wanted to add a validation of the feature.xml bnd generates.

PR #3259, opened 2025-08-27
Migrate "reference" documents to HTML5

ptziegler:
This change includes:

- Changing the doctype to "html".

- Simplifying the "Content-Type" meta element to only retain the encoding. The remaining attributes match the default settings and can be omitted.

- Removing the obsolete "charset attribute of the "link" element. The encoding is specified via the "Content-Type" header of the linked document.

- Removing the redundant "type" attribute for the "style" element. CSS is the default type and can therefore be omitted.

- Removing the obsolete "Content-Style-Type" meta element. Styles use CSS by default, so there is no need to set it globally for the document.

- Removing the obsolete "Content-Language" meta element. The document language is specified at the root element.

- Converting the obsolete "name" attribute of anchor points to "id" attributes of their nearest container and removing the "a" element all-together.

- Translating the styling of HTML elements to CSS.

- Cleanup of broken HTML or broken references.

### DIFF
--- a/eclipse.platform.common/bundles/org.eclipse.platform.doc.isv/reference/misc/multi_user_installs.html
+++ b/eclipse.platform.common/bundles/org.eclipse.platform.doc.isv/reference/misc/multi_user_installs.html
@@ -1,15 +1,14 @@
-<!DOCTYPE HTML PUBLIC "-//W3C//DTD HTML 4.01 Transitional//EN">
+<!DOCTYPE HTML>
 <html lang="en">
 <head>
 
 <meta name="copyright" content="Copyright (c) IBM Corporation and others 2000, 2005. This page is made available under license. For full details see the LEGAL in the documentation book that contains this page." >
 
-<meta http-equiv="Content-Language" content="en-us">
-<meta http-equiv="Content-Type" content="text/html; charset=iso-8859-1">
-<link REL="STYLESHEET" HREF="../../book.css" CHARSET="ISO-8859-1" TYPE="text/css">
+<meta charset="utf-8">
+<link REL="STYLESHEET" HREF="../../book.css" TYPE="text/css">
 <title>Eclipse Multi-user Installs</title>
 </head>
-<body text="#000000" bgcolor="#ffffff">
+<body>
 <h1>Eclipse Multi-user Installs</h1>
 <p>
   Eclipse provides a number of strategies for supporting multi-user installs.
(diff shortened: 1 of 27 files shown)


eclipse.platform.common/bundles/org.eclipse.platform.doc.isv/reference/misc/multi_user_installs.html
```diff
@@ -1,15 +1,14 @@
...
 <meta name="copyright" content="Copyright (c) IBM Corporation and others 2000, 2005. This page is made available under license. For full details see the LEGAL in the documentation book that contains this page." >
 
-<meta http-equiv="Content-Language" content="en-us">
-<meta http-equiv="Content-Type" content="text/html; charset=iso-8859-1">
-<link REL="STYLESHEET" HREF="../../book.css" CHARSET="ISO-8859-1" TYPE="text/css">
+<meta charset="utf-8">
+<link REL="STYLESHEET" HREF="../../book.css" TYPE="text/css">
 <title>Eclipse Multi-user Installs</title>
 </head>
-<body text="#000000" bgcolor="#ffffff">
```
ptziegler, line 12: Black text on white background is already the default. Keeping those values might also mess with e.g. a dark theme.

eclipse.platform.common/bundles/org.eclipse.platform.doc.isv/reference/misc/project_description_file.html
```diff
@@ -1,16 +1,16 @@
...
-  <link REL="STYLESHEET" HREF="../../book.css" CHARSET="ISO-8859-1" TYPE="text/css">
+   <meta charset="utf-8">
+  <link REL="STYLESHEET" HREF="../../book.css" TYPE="text/css">
   <title>Project description file</title>
 </head>
-<body link="#0000FF" vlink="#800080">
+<body>
 
-<h1 align="left">
+<h1>
```
ptziegler, line 13: The header is already left-aligned by default.

ptziegler:
I've checked the documents using the [Nu HTML checker](https://validator.github.io/validator/), which is the same validator used by e.g. https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/pull/3238.

I've fixed all errors and warnings, outside of the deprecated \<tt\> element, because that would overshadow all other changes.

akurtakov (approved):

akurtakov:
Sorry for the noise! I misclicked and created a merge request and had to remove it after that.

PR #3268, opened 2025-08-29
Remove images repository from Oomph setup

HeikoKlare:
The images repository has been archived. This also removes it form the SDK Oomph setup.


merks:
I discussed this with @akurtakov and given this has no impact on the artifacts produced by the build, we can merge this now.

PR #3274, opened 2025-08-31
Update the platform setup to redirect 4.36 to 4.37 and 4.37 to 4.38

merks:


merks:
This has no build impact.  Merging immediately.

PR #3283, opened 2025-09-03
Remove "porting" docs for Eclispe 4.28 and older

akurtakov:
Actually most of the documents are empty and this matches the https://github.com/eclipse-platform/eclipse.platform/blob/master/docs/ApiRemovalProcess.md period. It makes no sense to have documents that can point to no longer existing API.


HannesWell:
@akurtakov I see you added version bumps manually, was this just in anticipation or did it not work automatically? Because it should work for the doc bundles.

akurtakov:
It didn't work automatically and I decided to not bother you with that.

Issue #3275, opened 2025-09-01
Fix htmlvalidator problems


akurtakov:
In https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/pull/3238 a validator is being added in order to prevent issues like https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/issues/3236 . This ticket is to aggregate as a tracker for https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/pull/3259 , https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/pull/3240 and etc.

akurtakov:
All is done!

PR #3285, opened 2025-09-04
HTML validation fixes for jdt.doc.user

akurtakov:
Contributes to

### DIFF
--- a/eclipse.platform.common/bundles/org.eclipse.jdt.doc.user/META-INF/MANIFEST.MF
+++ b/eclipse.platform.common/bundles/org.eclipse.jdt.doc.user/META-INF/MANIFEST.MF
@@ -2,7 +2,7 @@ Manifest-Version: 1.0
 Bundle-ManifestVersion: 2
 Bundle-Name: %pluginName
 Bundle-SymbolicName: org.eclipse.jdt.doc.user; singleton:=true
-Bundle-Version: 3.15.2601.qualifier
+Bundle-Version: 3.15.2700.qualifier
 Bundle-Vendor: %providerName
 Bundle-Localization: plugin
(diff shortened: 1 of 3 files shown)


akurtakov:
@HannesWell version bump didn't kick in here. If you don't have time to look into it I'll bump manually tomorrow.

HannesWell:
> version bump didn't kick in here. If you don't have time to look into it I'll bump manually tomorrow.

The job is currently simply not configured to run in this directory:
- https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/blob/950d14faa86b041591e7742fb0ee87d34f5ce5c6/.github/workflows/pr-checks.yml#L25

I'll try to extend the pipeline to allow multiple working directories tomorrow.

HannesWell:
> I'll try to extend the pipeline to allow multiple working directories tomorrow.

- https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/pull/3293

laeubi:
> I'll try to extend the pipeline to allow multiple working directories tomorrow.

Per definition a working-directory is a single one and can't be multiple. I also doubt that the version increment workflow is really that useful here to complicate things more. Actually the goal should be to move the docs to JDT / platform (like we did for PDE) instead of make the aggregator requirements more complex.

HannesWell:
> > I'll try to extend the pipeline to allow multiple working directories tomorrow.
> 
> Per definition a working-directory is a single one and can't be multiple.

The variable can be renamed if you think another name fits better.



> Actually the goal should be to move the docs to JDT / platform (like we did for PDE) instead of make the aggregator requirements more complex.

Sure that would be better, but as long as this isn't the case I think it helps to have the version increment working.
Or will you (or somebody else) soon work on moving them?

PR #3286, opened 2025-09-05
Fix more docs html validation problems

akurtakov:
Contributes to


akurtakov:
Manually verified that the changed bundle version has been bumped thus merging without waiting for verification which is supposed to be fixed by linked PR.

PR #3287, opened 2025-09-05
Further htmlvalidor fixes in platform.doc.user

akurtakov:
Contributes to
https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/pull/3238


akurtakov:
Merging as due to https://github.com/eclipse-platform/eclipse.platform/pull/2140 verification will fail and the changed bundle has its version already bumped.

Issue #3264, opened 2025-08-28
Preparation work for 4.38 (2025-12) and open master for development


HannesWell:
This preparation work involves the following tasks. For previous bug please refer to eclipse-platform/eclipse.platform.releng.aggregator#3070.

- [ ] Splash Screen for 4.38 (2025-12)
- [ ] Run [`Prepare Next Development Cycle`](https://ci.eclipse.org/releng/job/Releng/job/prepareNextDevCycle/) job.
Review and submit the created changes
- [ ] Clean forceQualifierUpdate files for doc bundles
- [ ] Update build calendar for platform 4.38 release

MohananRahul:
> [@MohananRahul](https://github.com/MohananRahul), [@subyssurendran666](https://github.com/subyssurendran666) I created this for the release preparation as it was not created yet? If that was intentionally, sorry for disturbing your plans, but I wanted to discuss a few points (and didn't want to use the release issue).

Thanks @HannesWell 

> [@MohananRahul](https://github.com/MohananRahul) is that fine for you or do you think something is missing for that? Cleaning the `forceQualifierUpdate` files and updating the Splash Screen can be done afterwards everything else was prepared.

Looks fine to me, but we’ll also need to update the maintenance branch once it’s created.

- Move 4.37-I builds to R4_37_maintenance branch
- Update R4_37 maintenance branch with release version for 4.37+ changes

> For the milestone/RC/GA dates I have looked at the [SimRel 2025-12 schedule](https://github.com/eclipse-simrel/.github/blob/main/wiki/SimRel/2025-12.md) and derived the following dates from it:
> 
> * M1: `2025-10-03`
> * M2: `2025-10-24`
> * M3: `2025-11-14`
> * RC1: `2025-11-21`
> * RC2: `2025-11-28`
> * GA: `2025-12-05`
> 
> [@MohananRahul](https://github.com/MohananRahul) is that all correct? 

All good

> Last but not least, I have disabled the [I-build-4.37](https://ci.eclipse.org/releng/job/Builds/job/I-build-4.37/) build as there are no more builds planned, but the cron-schedule was not properly updated. I suggest to now create a PR for that but change it manually once we have the maintenance branch (and also move it to the maintenance branch until the job can be deleted). I'll also think about how this could be automated as well (but not for this time, we should do it manually).

Yes, I have updated now.

MohananRahul:
[...]

PR #3289, opened 2025-09-05
Add a dependency mapping rule to avoid bogus SWT fragment dependencies

merks:
https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/issues/3264


merks:
Has no build impact, merging immediately.

PR #3290, opened 2025-09-05
[RelEng] Move Maven staging behind build-update in release promotion

HannesWell:
The Staging to Maven-Central is just triggering the Maven-deployment job and is therefore easier to do manually in case something fails before. It should therefore happen later, as it's also a stage that's not much tested in advance.


HannesWell:
The verification build does not test this, thus I'm submitting it immediately.

PR #3293, opened 2025-09-05
[RelEng] Support multiple working-directories in Version-Increment workflow

HannesWell:
And check the versions within the `eclipse.platform.releng` and `eclipse.platform.common` directory in this repositories.

### DIFF
--- a/.github/workflows/checkVersions.yml
+++ b/.github/workflows/checkVersions.yml
@@ -22,16 +22,25 @@ on:
         required: false
         default: ''
       working-directory:
-        description: Optional additional arguments to specify the directory in which maven build is executed
+        description: Optional, comma-separated list of directories in which this version-check is executed
         type: string
         required: false
         default: '.'
+      submodules:
+        description: |
+          Whether to checkout submodules: `true` to checkout submodules or `recursive` to recursively checkout submodules.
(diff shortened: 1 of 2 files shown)


HannesWell:
In general this seems to work, but now fails with
```
Error:  Failed to execute goal org.eclipse.tycho.extras:tycho-document-bundle-plugin:5.0.0:schema-to-html (schema-to-html) on project org.eclipse.platform.doc.isv: java.io.FileNotFoundException: /home/runner/work/eclipse.platform.releng.aggregator/eclipse.platform.releng.aggregator/eclipse.platform.common/bundles/org.eclipse.platform.doc.isv/../../../eclipse.platform/ant/org.eclipse.ant.core/plugin.xml (No such file or directory) -> [Help 1]
```
Seems like we have to checkout sub-modules too.

HannesWell:
@laeubi in https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/pull/3285#issuecomment-3259274806 you said, that this would be too complicated.
Do you or anybody else plan on moving the doc bundles to the appropriated eclipse.platform respectively JDT repositories so that we don't need this anymore?
If we keep them doc bundles here as they are, I think this still would be useful. But if they are moved, I'm also happy to keep the workflow simpler.

PR #3299, opened 2025-09-06
Migrate "Ant" Guides to HTML5

ptziegler:
This cleanup up the metadata and fixes broken links.
Contributes to


akurtakov:
@ptziegler Would you put errors on purpose in an .html and in .htm file so we can see what gets reported?

ptziegler:
I did some tests and `.html` files still are validated, while `.htm` files are not. Though I still find the 3-4s to be quite low. Does it perhaps only check changed files?

ptziegler:
i.e. this change is clearly invalid
image

ptziegler:
Merging, as .htm files are only validated after https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/pull/3342. I've reverted the change to the file extension, so this commit only contains the HTML cleanup again.

PR #3305, opened 2025-09-08
[RelEng] Introduce dry-run parameter in promotion pipeline

HannesWell:
This introduces a `DRY_RUN` parameter for the promotion pipeline, which simplifies future testing and debugging and hopefully reduces the chance for regression (recent changes where not tested well enough and had numerous regression).

### DIFF
--- a/JenkinsJobs/Releng/promoteBuild.jenkinsfile
+++ b/JenkinsJobs/Releng/promoteBuild.jenkinsfile
@@ -11,6 +11,7 @@ pipeline {
 	environment {
 		// Download Server locations (would very seldom change)
 		EP_ROOT = '/home/data/httpd/download.eclipse.org'
+		DRY_RUN_PREFIX = "${params.DRY_RUN ? 'eclipse/try-outs/' : ''}"
 	}
 	tools {
 		jdk 'temurin-jdk21-latest'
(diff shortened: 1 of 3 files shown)


JenkinsJobs/Releng/promoteBuild.jenkinsfile
```diff
@@ -321,7 +343,7 @@ pipeline {
 
 def renameBuildDrop(String baseDropPath, String oldDropID, String oldBuildLabel, String newDropID, String newBuildLabel, Closure extraTasks=null) {
 	def sourcePath="${EP_ROOT}/${baseDropPath}/${oldDropID}"
-	def targetPath="${EP_ROOT}/${baseDropPath}/${newDropID}"
+	def targetPath="${EP_ROOT}/${DRY_RUN_PREFIX}${baseDropPath}/${newDropID}"
```
MohananRahul, line 346: @HannesWell  I believe this should be:
def targetPath="${EP_ROOT}/${env.DRY_RUN_PREFIX}${baseDropPath}/${newDropID}"
as the environment variables need to be explicitly referenced via the env object.
I noticed this issue when attempting the promotion for M1.

HannesWell: I'm having a look at this.

PR #3310, opened 2025-09-09
[RelEng] Make read of releaseName more robust in publication pipeline

HannesWell:
Don't launch Maven to output the interpolated `releaseName` property, but instead read the `releaseYear` and `releaseMonth` properties from the raw (`eclipse-platform-parent/pom.xml`) file.
This avoids problems if at the time of the release (candidate) build a snapshot version was used, e.g. of Tycho, that's not available anymore at the time of the publication (because e.g. that Tycho version was released in the meantime).


HannesWell:
The verification build doesn't test this, thus I'm submitting this immediately to fix this blocker (i.e. the use of Tycho 5.0.0-SNAPSHOT at the time of the release build) for tomorrows release publication.

PR #3322, opened 2025-09-14
[Build] Unify clean-ups jobs as Jenkins pipelines and inline scripts

HannesWell:
- Migrate all clean-up jobs to pure Jenkinsfile/pipeline based jobs.
- Unify the daily clean-up into one job that handles Eclispe and Equinox builds in parallel.
- Move remaining two jobs to the 'Releng' folder as it already contains the manually maintained variant of the clean-up job for release artifacts.
- Automatically run clean-up of release artifacts when promoting a release.


HannesWell:
This is a draft as it requires:
- https://gitlab.eclipse.org/eclipsefdn/helpdesk/-/issues/6628

And a few more things have to be cleaned-up.

JenkinsJobs/Cleanup/cleanupReleaseArtifacts.jenkinsfile
```diff
@@ -0,0 +1,125 @@
...
+						} // Y-build drops are deleted continuously based on their count, nothing to do here
+						
+						// --- Update sites ---
+						releaseVersions.each{ release, versions ->
+							utilities.removeDropsOnRemote("${EP_ECLIPSE_UPDATES}", ["${versions.version}-I-builds"])
+							if (release != latestMinorRelease || (getBuildProperty("${EP_ECLIPSE_DROPS}", release, 'NEXT_JAVA_RELEASE_DATE')?.isEmpty() ?: false)) {
+								// Delete Y-build p2-repositories for the latest release only if no Java release is scheduled shortly after that release
+								utilities.removeDropsOnRemote("${EP_ECLIPSE_UPDATES}", ["${versions.version}-Y-builds"])
+							}
+							//TODO: maybe keep always the last 10 children, if this is the latest release? Regardless of a new Java-release is close or not
```
HannesWell, line 78: This considers the new `NEXT_JAVA_RELEASE_DATE` that was introduced with
- https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/pull/3339
- https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/pull/3351

... to determine if the Y-build p2-repositories should be cleaned up or if a java release is imminent and the p2-repo should be kept until the next release, just as you said in
- https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/issues/3328#issuecomment-3295127759

@mpalat, @jarthana or @stephan-herrmann, but I think the clean-up of be more robust (i.e. robust against that this variable was accidentally not set when the release cycle was prepared and it wasn't noticed and added afterwards), if we for example would just keep always a sufficient number Y-build p2-repositories created for the latest release, regardless of a Java release is imminent or not. For the final P-builds you usually just use the latest one, don't you?
So my suggest would be to always keep the last 10 Y-build p2-repos of the latest release.

Would that be fine for you?

HannesWell:
> As suggested in [#3328 (comment)](https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/issues/3328#issuecomment-3309137932), this also changes the `Daily Cleanup of old Builds` to keep always the **Drops! at the download page of the latest six Y-builds**. This corresponds to two weeks of regular Y-builds. The P2-repository is not affected by this!

Unfortunately the I didn't get any response to this and the other questions as the addressees are on vacation or unavailable.
But as I want to proceed as this contains many other changes I have further relaxed the policy for now to keep the drops of nine Y-builds and to always keep all Y-build p2-repos of the latest release, regardless of if a java release is due. As Y-builds are currently not active, no new builds are added and thus nothing will be deleted.
I'll create a follow-up PR soon with the goal to tighten these bounds again, which will hopefully receive more feedback. Ideally this is done before Y-builds are reactivated for Java-26, then nothing should be deleted without consent.

stephan-herrmann:
> Unfortunately the I didn't get any response to this and the other questions as the addressees are on vacation or unavailable.

Sorry for not responding. I personally don't see much need for retaining lots of "old" Y-builds. _My_ only use for previous Y-builds is for comparing test results. And every now and then the latest **successful** Y-build will be used to produce a P-build. That's it.

How Y-builds are consumed by early adopters I simply don't know.

PR #3333, opened 2025-09-16
[RelEng] Move previous release's I-build job to maintenance branch

HannesWell:
when preparing a new dev cycle. This used to be done in the past but was forgotten or just stopped to be done and was then done manually. On the master keep the old and new job until the old job is finally removed when the release is promoted.


HannesWell:
The verification builds doesn't test this and my other testing was successful. Submitting immediately.

PR #3340, opened 2025-09-17
[RelEng] Relax replacement patterns in release preparation pipeline

HannesWell:
This makes the preparation pipeline more flexible in case some build-properties have changed to something else than exactly the previous version as value in-between releases.


HannesWell:
The verification build doesn't check this (manual testing was successful), thus submitting this immediately.

PR #3368, opened 2025-10-01
Update org.hamcrest to 3.0.

merks:


HannesWell:
I recall that some version ranges had to be adjusted in JDT to work with the previous new hamcrest minor.
So it might be necessary again?

PR #3376, opened 2025-10-03
Exclude junit-platform-launcher from analyzed bundles

akurtakov:
Deprecations in it are outside of project control so shouldn't be reported. (https://download.eclipse.org/eclipse/downloads/drops4/I20251002-1800/apitools/deprecation/apideprecation.html )


akurtakov:
Merging as it's used only by I-builds.

PR #3380, opened 2025-10-04
[RelEng] Fix access to empty DRY_RUN_PREFIX in promotion pipeline

HannesWell:
If an environment variable is defined with an empty value (as it's the case for 'DRY_RUN_PREFIX' if a build is not a dry-run) it's effectively discarded, not passed to scripts and is not available as environment variable binding respectively in the special 'env' map/variable.

Implement corresponding special handling for that case of a non-dry-run.

Refers to
- https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/pull/3305#discussion_r2403771309

CC @MohananRahul


HannesWell:
> @MohananRahul I'll restart the promotion with this fix applied and will promote the latest I-build from tonight (so one day later):
> 
>     * https://download.eclipse.org/eclipse/downloads/drops4/I20251003-1800/

Successfully done.

And thanks for the report. I'm glad this corner-case is now sanitized as well.

PR #3381, opened 2025-10-04
[Build] Streamline and speed-up test-result collection

HannesWell:
Instead of downloading the entire drop of a build to the build agent, just download the required test-result XML files of other test-runs of the same build.
Also avoid use of `raw.githubusercontent` URLs as their use is rate-limited.
Furthermore remove unused elements in `publish.xml` and `eclipse/helper.xml` after the performance-test infrastructure was removed.


iloveeclipse:
Please check https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/issues/3386 , I guess this PR caused regression.

PR #3382, opened 2025-10-04
Convert "Editors" documents to  HTML5

ptziegler:
Contributes to


ptziegler:
It looks like the HTML validator is only executed when htm/html files are modified, but it still validates _all_ files. So the check fails because of files that aren't modified by this PR.

ptziegler:
Regardless, the validator doesn't complain about any file starting with `editor_`, which is exactly what I fixed...

PR #3384, opened 2025-10-04
Migrate "Advanced workbench concepts" to HTML5

ptziegler:
Contributes to

### DIFF
--- a/eclipse.platform.common/bundles/org.eclipse.platform.doc.isv/guide/wrkAdv_markers.htm
+++ b/eclipse.platform.common/bundles/org.eclipse.platform.doc.isv/guide/wrkAdv_markers.htm
@@ -1,12 +1,11 @@
-<!DOCTYPE HTML PUBLIC "-//W3C//DTD HTML 4.0 Transitional//EN"><html lang="en">
+<!DOCTYPE HTML><html lang="en">
 <HEAD>
 
 <meta name="copyright" content="Copyright (c) IBM Corporation and others 2000, 2005. This page is made available under license. For full details see the LEGAL in the documentation book that contains this page." >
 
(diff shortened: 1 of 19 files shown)


eclipse.platform.common/bundles/org.eclipse.platform.doc.isv/guide/wrkAdv_markers.htm
```diff
@@ -47,12 +46,12 @@ <h3>Marker Help and Resolution</h3>
 
 
 <p>Markers can be added to a <b>.readme</b> file using the readme editor's popup
-menu.&nbsp; (The popup menu actions are added dynamically in <b>ReadmeTextEditor.<font SIZE="2" COLOR="#000000">editorContextMenuAboutToShow(IMenuManager</font>
-</b><font SIZE="2" COLOR="#000000"><b>parentMenu)</b>).&nbsp; Once added,
-the markers appear on the left side of the editor and in the tasks view.</font></p>
+menu.&nbsp; (The popup menu actions are added dynamically in <b>ReadmeTextEditor.<span style="font-size: smaller;">editorContextMenuAboutToShow(IMenuManager
+parentMenu)</span></b>).&nbsp; Once added,
+the markers appear on the left side of the editor and in the tasks view.</p>
```
ptziegler, line 51: I think the idea behind this font was to make the method name appear smaller than the class name. The problem is that the tags were set improperly and as a result, the rest of the text was also rendered smaller, which just looks weird. This is how it looks now:

image

akurtakov: Looks good.

Issue #3386, opened 2025-10-05
No test results published for I20251004-1800


iloveeclipse:
See
* https://download.eclipse.org/eclipse/downloads/drops4/I20251004-1800/
* https://ci.eclipse.org/releng/job/Releng/job/collectTestResults/4/console

Probably regression from https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/pull/3381

```
05:37:51  + rsync -avzh '--exclude=*_ep438I-unit-macosx-x86_64-java21_*' genie.releng@projects-storage.eclipse.org:/home/data/httpd/download.eclipse.org/eclipse/downloads/drops4/I20251004-1800/testresults/xml /home/jenkins/agent/workspace/Releng/collectTestResults/postingDir/I20251004-1800/testresults
05:37:53  receiving incremental file list
05:37:53  rsync: [sender] link_stat "/home/data/httpd/download.eclipse.org/eclipse/downloads/drops4/I20251004-1800/testresults/xml" failed: No such file or directory (2)
05:37:53  
05:37:53  sent 50 bytes  received 8 bytes  23.20 bytes/sec
05:37:53  total size is 0  speedup is 0.00
05:37:53  rsync error: some files/attrs were not transferred (see previous errors) (code 23) at main.c(1865) [Receiver=3.2.7]
```

HannesWell:
> Probably regression from [#3381](https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/pull/3381)

Yes, definitively. I'm working on it. Thanks for the report.

iloveeclipse:
Broken again in https://download.eclipse.org/eclipse/downloads/drops4/I20251005-1800/
See https://ci.eclipse.org/releng/job/Releng/job/collectTestResults/10/console

PR #3387, opened 2025-10-05
[Build] Fix collection of the first test results

HannesWell:
Check if results for other test configurations have been published yet and if the corresponding directory already exists on the storage server.

Additionally use the `eclipse-platform` instead of the `eclipse-SDK` product. The former is smaller, but sufficient for the executed tasks.

Fixes https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/issues/3386


HannesWell:
The verification build doesn't test this, but my manual testing was successful, thus I'm submitting this immediately.

PR #3388, opened 2025-10-05
Fix html warnings in colordialog.htm

akurtakov:
Contributes to


akurtakov:
Html validation fails as there are many other warnings but none in this file. The rest of validations are pointless for this change. Merging.

PR #3392, opened 2025-10-06
Remove com.github.weisj.jsvg version 1.7.2 from target platform

HeikoKlare:
⚠️ To be merged after https://github.com/eclipse-platform/eclipse.platform.swt/pull/2597 as SWT still depends on JSVG 1.x until that PR is processed.


HeikoKlare:
Started an I-Build to verify that the removal does not break anything: https://ci.eclipse.org/releng/job/Builds/job/I-build-4.38/58

PR #3396, opened 2025-10-06
[Build] Remove P-build reference from Y-Builds Jenkins folder

HannesWell:
since P-builds are now managed in the JDT JIPP.

If applied the build-tools have to be adjusted too:
- https://github.com/eclipse-platform/eclipse.platform.releng.buildtools/blob/c5f7ecf1951d44311e24ce7bd6b505189aabb4da/bundles/org.eclipse.build.tools/src/org/eclipse/releng/generators/TestResultsGenerator.java#L812

@stephan-herrmann, @jarthana and @mpalat is this fine for you?
I don't know if you reference these jobs directly in Jenkins. If yes, you would have to adjust some URLs


jarthana:
I have no idea. @stephan-herrmann  is the right person to answer. So, let's wait for his response.

stephan-herrmann:
If the question is only about names of build jobs, then you are 100% free to decide. I then only need to update one bookmark where I monitor these builds :)

The only point of technical interaction is the URL of the resulting p2 repo (as referenced from PR builds and P-builds). This URL, too, can be freely chosen at the start of a branch (we need to set up this configuration anew anyway), if only during development in the branch it remains stable.

PR #3398, opened 2025-10-07
Exclude asm commons from deprecations report

akurtakov:
Third party dependencies are beyond project's control so deprecations in them shouldn't be reported.
https://download.eclipse.org/eclipse/downloads/drops4/I20251006-1800/apitools/deprecation/apideprecation.html


akurtakov:
Used only by I-builds so verification will not test it. Merging.

PR #3399, opened 2025-10-07
Added JUnit 6 to platform

trancexpress:


laeubi:
Build fails with

> Target definition is not well-formed XML: The element type "version" must be terminated by the matching end-tag "`</version>`"

eclipse.platform.releng.prereqs.sdk/eclipse-sdk-prereqs.target
```diff
@@ -903,4 +985,4 @@
 	  </location>
   </locations>
   <targetJRE path="org.eclipse.jdt.launching.JRE_CONTAINER/org.eclipse.jdt.internal.debug.ui.launcher.StandardVMType/JavaSE-21"/>
-</target>
```
merks, line 906: Maybe we could avoid this whitespace change?

eclipse.platform.releng.prereqs.sdk/eclipse-sdk-prereqs.target
```diff
@@ -792,6 +792,88 @@
...
+				  <artifactId>apiguardian-api</artifactId>
+				  <version>1.1.2</version>
+				  <type>jar</type>
+			  </dependency>
+			  <dependency>
+				  <groupId>org.opentest4j</groupId>
+				  <artifactId>opentest4j</artifactId>
+				  <version>1.3.0</version>
+				  <type>jar</type>
+			  </dependency>
```
merks, line 808: These two are duplicates.  That's might become a point of confusion.

merks (changes requested):
Probably we should avoid duplicates and unnecessary whitespace changes.

trancexpress:
> /request-license-review

Let me get the build to work locally, then we can do this. Likely the set of libraries will change.

merks:
@trancexpress 

I had similar problems locally when updating Orbit.   Just at a glance I noticed you added 11 things but in Orbit I only added 10 things:

https://github.com/eclipse-orbit/orbit-simrel/blob/db315d1b0b270ed6e5350ce209cf8780bcf44a1a/maven-osgi/tp/Maven.target#L1337-L1468

Maybe I missed something?  Or maybe one of the things you added was bogus (which happened to me too).

Maybe this summary helps:

image

https://download.eclipse.org/tools/orbit/simrel/orbit-aggregation/table.html

trancexpress:
> I had similar problems locally when updating Orbit. Just at a glance I noticed you added 11 things but in Orbit I only added 10 things:

> Maybe I missed something? Or maybe one of the things you added was bogus (which happened to me too).

One of the libraries is removed (`junit-platform-runner`), I overlooked that. Likely that is the difference.

trancexpress:
```
[ERROR] Failed to execute goal org.eclipse.tycho:tycho-compiler-plugin:5.0.0:validate-classpath (default-validate-classpath) on project org.eclipse.test: Execution default-validate-classpath of goal org.eclipse.tycho:tycho-compiler-plugin:5.0.0:validate-classpath failed: org.osgi.framework.BundleException: Bundle org.eclipse.test cannot be resolved:org.eclipse.test [147]
[ERROR]   Unresolved requirement: Import-Package: org.junit.platform.commons.util
[ERROR] 
[ERROR] -> [Help 1]
org.apache.maven.lifecycle.LifecycleExecutionException: Failed to execute goal org.eclipse.tycho:tycho-compiler-plugin:5.0.0:validate-classpath (default-validate-classpath) on project org.eclipse.test: Execution default-validate-classpath of goal org.eclipse.tycho:tycho-compiler-plugin:5.0.0:validate-classpath failed: org.osgi.framework.BundleException: Bundle org.eclipse.test cannot be resolved:org.eclipse.test [147]
  Unresolved requirement: Import-Package: org.junit.platform.commons.util
```

With the latest changes.

laeubi:
> I guess we can replace the call with just the contents of the method... Its nothing special.

 I'm not sure what is the suggestion here? We should really not "reimplement" what JUnit already offers, if there is any issue in the usage of the JUnit in **client code** then it should be reported to JUnit team.

If it is in the code we use to **run the JUnit platform** then it exactly falls under

> These utilities are intended solely for usage within the JUnit framework itself.

and we should just include whats required here.

trancexpress:
> I'm not sure what is the suggestion here? We should really not "reimplement" what JUnit already offers, if there is any issue in the usage of the JUnit in **client code** then it should be reported to JUnit team.

```
diff --git a/eclipse.platform.releng/bundles/org.eclipse.test/META-INF/MANIFEST.MF b/eclipse.platform.releng/bundles/org.eclipse.test/META-INF/MANIFEST.MF
index c77d2d68a4..f8ea7bbc71 100644
--- a/eclipse.platform.releng/bundles/org.eclipse.test/META-INF/MANIFEST.MF
+++ b/eclipse.platform.releng/bundles/org.eclipse.test/META-INF/MANIFEST.MF
@@ -20,7 +20,6 @@ Import-Package: org.junit.jupiter.api,
  org.junit.platform.launcher,
  org.junit.platform.launcher.core,
  org.junit.platform.commons,
- org.junit.platform.commons.util,
  org.junit.jupiter.engine
 Bundle-ActivationPolicy: lazy
 Bundle-RequiredExecutionEnvironment: JavaSE-17
diff --git a/eclipse.platform.releng/bundles/org.eclipse.test/src/org/eclipse/test/LegacyXmlResultFormatter.java b/eclipse.platform.releng/bundles/org.eclipse.test/src/org/eclipse/test/LegacyXmlResultFormatter.java
index 0c4b1c3cba..bb2b0f4a24 100644
--- a/eclipse.platform.releng/bundles/org.eclipse.test/src/org/eclipse/test/LegacyXmlResultFormatter.java
+++ b/eclipse.platform.releng/bundles/org.eclipse.test/src/org/eclipse/test/LegacyXmlResultFormatter.java
@@ -15,7 +15,9 @@ package org.eclipse.test;
 
 import java.io.IOException;
 import java.io.OutputStream;
+import java.io.PrintWriter;
 import java.io.Reader;
+import java.io.StringWriter
 import java.util.Date;
 import java.util.Map;
 import java.util.Optional;
@@ -333,7 +335,11 @@ public class LegacyXmlResultFormatter extends AbstractJUnitResultFormatter {
                                writer.writeAttribute(ATTR_TYPE, t.getClass().getName());
 
                                if (writeExceptionStackTrace) {
-                                       writer.writeCharacters(ExceptionUtils.readStackTrace(t));
+                                       StringWriter stringWriter = new StringWriter();
+                                       try (PrintWriter printWriter = new PrintWriter(stringWriter)) {
+                                               t.printStackTrace(printWriter);
+                                       }
+                                       return stringWriter.toString();
                                }
                        }
                        writer.writeEndElement();
```

> If it is in the code we use to **run the JUnit platform** then it exactly falls under
> 
> > These utilities are intended solely for usage within the JUnit framework itself.
> 
> and we should just include whats required here.

This is not how I understand "internal" code, but what is your proposal? Do we have a way to avoid the error from my comment with some `MANIFEST.MF` change?

Or from your POV we ask the JUnit contributors to relax the restriction on the package?

trancexpress:
@merks maybe it will be faster to just ask you, where do we include JUnit now? So that its actually added in the SDK. I see its no longer in the JDT feature: https://github.com/eclipse-jdt/eclipse.jdt/commit/bae402ea86337f00545bc74e8fb1f8c9a1d52c4b

merks (approved):
It looks okay.  The whitespace change in the *.target is a nitty annoyance, but one we could overlook.

merks:
> But if I build from the aggregator repository, the resulting SDK doesn't have JUnit 6.

That's kind of what I would expect (and hope).  I would have assumed that consumers of junit 5 typically have upper bounds that exclude 6.x, but that seems most often not the case.   E.g., things like this:

image

Is there some agree-to plan to abandon JUnit?

merks:
> So we need some plug-in to have a `MANIFEST.MF` that requires a JUnit 6 bundle (`[6.0,7.0)`)? Then (assuming all are depended on) we would have the new bundles in the platform?

Yes if some things require < 6 and nothing requires >=6 then the transitive closure does need 6.x so won’t include it. 

So I think this PR is a step in the right direction for making 6.x available for upcoming jdt changes.  Note that 6.x does not need to be in the I-build result.  The root pom will be published and subsequent builds will have the target platform changes of this commit available and that is sufficient.

trancexpress:
> So I think this PR is a step in the right direction for making 6.x available for upcoming jdt changes. Note that 6.x does not need to be in the I-build result. The root pom will be published and subsequent builds will have the target platform changes of this commit available and that is sufficient.

Should I look for the `MANIFEST.MF` from which this is coming from?

```
$ find . -name "*6.0.0*"
./eclipse.platform.releng.tychoeclipsebuilder/eclipse-junit-tests/target/repository/plugins/junit-jupiter-api.source_6.0.0.jar
./eclipse.platform.releng.tychoeclipsebuilder/eclipse-junit-tests/target/repository/plugins/junit-jupiter-api_6.0.0.jar
./eclipse.platform.releng.tychoeclipsebuilder/eclipse.platform.repository/target/repository/plugins/junit-jupiter-api.source_6.0.0.jar
./eclipse.platform.releng.tychoeclipsebuilder/eclipse.platform.repository/target/repository/plugins/junit-jupiter-api_6.0.0.jar
```

Seems like something we should fix before merging, ensuring `junit-jupiter-api` is depended on only `[5.0,6.0)]` for the current code bases.

merks:
Personally I would merge and deal with the fallout afterwards.  Do you need someone else to push the green button?

trancexpress:
> Personally I would merge and deal with the fallout afterwards.

Well the SDK itself doesn't have `junit-jupiter-api` `6.0.0`, its "only" in the built repository. I also checked adding JUnit to a plug-in project and a Java project, running tests, that works fine.

> Do you need someone else to push the green button?

@iloveeclipse are we merging this **now**? From my POV we are not in a rush here, I still need to add dependencies to JUnit 6 and check what happens with the built SDK - if there are bugs due to having 2 bundles with the same symbolic name of most JUnit 5/6 jars. So I'm not sure merging this **now** will help much with adding JUnit 6 support. But we will find problems with this particular change sooner...

iloveeclipse:
> To my thinking finding problems sooner rather than later is better.

Agree. The change here shouldn't be breaking, and if it is, better to be informed ASAP. Let's merge ans see what happens with tests tomorrow.

iloveeclipse:
> Let's merge ans see what happens with tests tomorrow.

SDK tests seem to be not affected, however looking on these strange failures I fear they could be related: https://github.com/eclipse-platform/eclipse.platform.ui/pull/3382#issuecomment-3382178174

eclipse.platform.releng/features/org.eclipse.sdk.tests/feature.xml
```diff
@@ -2,7 +2,7 @@
 <feature
       id="org.eclipse.sdk.tests"
       label="%featureName"
-      version="4.38.0.qualifier"
+      version="4.38.100.qualifier"
```
HannesWell: This version increment should not be necessary (a enforced qualifier-update, if necessary at all, could probably have been sufficient).
In general this version should be aligned with the eclipse release version and the current state breaks the release preparation workflow (respectively the workflow detects the missing version increment and fails).

This is work-around with
- https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/pull/3508

PR #3400, opened 2025-10-07
[Build] Stop unused deployments of releng scripts to storage server

HannesWell:
Except for the `getEBuilder.xml` ANT scripts nothing from https://download.eclipse.org/eclipse/relengScripts is used anymore. Therefore the deployment can be stopped.

In fact the entitlements are in use at the moment by
https://github.com/eclipse-equinox/equinox/blob/88ad15678b2607670c883ec3ef5309d9ef7f6ccb/Jenkinsfile#L248

but we can use a github-raw link as well there, which should be sufficient in that case.

In the foreseeable future probably even the `getEBuilder.xml` script becomes obsolete and the entire deployment of releng-scripts to the storage server can be stopped.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -36,7 +36,7 @@ pipeline {
 						return !sh(script:'''
 							latestCommitID=$(curl https://download.eclipse.org/eclipse/relengScripts/state)
 							git diff --name-only ${latestCommitID} HEAD \\
-								production/ cje-production/ eclipse.platform.releng.tychoeclipsebuilder/
+								production/testScripts/bootstrap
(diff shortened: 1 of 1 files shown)


HannesWell:
> In fact the entitlements are in use at the moment by https://github.com/eclipse-equinox/equinox/blob/88ad15678b2607670c883ec3ef5309d9ef7f6ccb/Jenkinsfile#L248
> 
> but we can use a github-raw link as well there, which should be sufficient in that case.

This is done in
- https://github.com/eclipse-equinox/equinox/pull/1150

PR #3402, opened 2025-10-07
[I/Y-Build] Unify build configurations into buildConfigurations.json

HannesWell:
The buildConfigurations.json becomes the central and only source for configurations of all I/Y-builds and their associated tests. It's read by the seed-job to create the corresponding Jenkins jobs and during I/Y-builds executions.

This avoids a definition of test configurations split over multiple files and avoids continuous changes in the build pipeline file due to regular version, Java or infrastructure updates.


HannesWell:
@MohananRahul, @subyssurendran666 FYI. After this is submitted, adding new test configurations or updating their JDK is all done in the new `buildConfigurations.json` file.

In general the idea is to have all configuration of I/Y-builds, that changes regularly, like versions, test configurations or Java version, at one central place in the `buildConfigurations.json` file.
The pipeline file or the scripts are only changed, if the build changes structurally.

PR #3416, opened 2025-10-14
[I/Y-Build] Unify definition of test-jobs into one pipeline file

HannesWell:
This also has the advantage that input parameters for the test builds are now always trimmed.
It also become easier to run other/smaller test-suites than the `all` suite.

Last part of
- https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/issues/2625


iloveeclipse:
Could it be, this caused https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/issues/3417 ?

JenkinsJobs/shared/utilities.groovy
```diff
@@ -104,4 +110,31 @@ private void removeDropsOnRemote(String remoteDirectory, List<String> drops) {
...
+	// Translate os/arch names that are different in the Adoptium API
+	if (arch == 'x86_64') {
+		arch = 'x64'
+	}
+	if (os == 'macosx') {
+		os = 'mac'
+	} else if (os == 'win32') {
+		os = 'windows'
+	}
+	return installDownloadableTool('jdk', "https://api.adoptium.net/v3/binary/latest/${version}/${releaseType}/${os}/${arch}/jdk/hotspot/normal/eclipse") + (os == 'mac' ? '/Contents/Home' : '')
```
MohananRahul, line 125: Hi @HannesWell  just a small observation for whenever Java 27 test configurations get added here , it looks like Temurin hasn't published GA binaries for Java 27 yet (per [Adoptium's release roadmap](https://adoptium.net/support), it's currently listed as "Latest Release: None, yet". So the case 'temurin': return utilities.downloadTemurinJDK(...) path would likely not work for a javaVersion: 27 entry just yet.
Would it make sense to use the case 'install' path (utilities.installDownloadableTool('jdk', test.jdk.url)) instead, pointing at an OpenJDK.net GA archive in the meantime?

PR #3432, opened 2025-10-25
Build with Tycho 5.0.1-SNAPSHOT

HannesWell:
This makes the fix to not mix JUnit-5 and JUnit-6 bundles in a test-runtime, if both are in the target-platform but only JUnit-5 is actually referenced:
- https://github.com/eclipse-tycho/tycho/pull/5526

The back-port is what will be actually used:
- https://github.com/eclipse-tycho/tycho/pull/5529


HannesWell:
> The master build currently fails due to a compiler error:

I'm now trying to unblock this by running the verification build without updateing the submodules to the tip of the master branch, but instead on the state recorded in the aggregator, i.e. at the state of the latest I-build.
I think the very latest changes are not that relevant for this, but I would like to have a generally good verification build first.

Issue #3436, opened 2025-10-26
OutOfMemory errors in PDE build since Tycho 5.0.1-SNAPSHOT


HannesWell:
Since we upgraded to Tycho 5.0.1-SNAPSHOT in https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/pull/3432, the PDE master build is failing constantly with OutOfMemoryErrors when building the o.e.pde.ui jar:
```
 [INFO] --- tycho-packaging:5.0.1-SNAPSHOT:package-plugin (default-package-plugin) @ org.eclipse.pde.ui ---
 [INFO] Building jar: /home/jenkins/agent/workspace/eclipse.pde_master/ui/org.eclipse.pde.ui/target/org.eclipse.pde.ui-3.16.300-SNAPSHOT.jar
 [578.399s][warning][gc,alloc] ForkJoinPool-1-worker-10: Retried waiting for GCLocker too often allocating 62502 words
 java.lang.OutOfMemoryError: Java heap space
 Dumping heap to jvmHeapDump.hprof ...
 Heap dump file created [1201636756 bytes in 8.056 secs]
 
 Exception: java.lang.OutOfMemoryError thrown from the UncaughtExceptionHandler in thread "HttpClient-9-SelectorManager"
 Exception: java.lang.OutOfMemoryError thrown from the UncaughtExceptionHandler in thread "HttpClient-7-SelectorManager"
 Exception: java.lang.OutOfMemoryError thrown from the UncaughtExceptionHandler in thread "HttpClient-10-SelectorManager"
 Exception: java.lang.OutOfMemoryError thrown from the UncaughtExceptionHandler in thread "HttpClient-1-SelectorManager"
 Exception: java.lang.OutOfMemoryError thrown from the UncaughtExceptionHandler in thread "Worker-0"
 Exception: java.lang.OutOfMemoryError thrown from the UncaughtExceptionHandler in thread "HttpClient-8-SelectorManager"
 Exception: java.lang.OutOfMemoryError thrown from the UncaughtExceptionHandler in thread "Worker-1"
 Exception: java.lang.OutOfMemoryError thrown from the UncaughtExceptionHandler in thread "HttpClient-6-SelectorManager"
 Exception: java.lang.OutOfMemoryError thrown from the UncaughtExceptionHandler in thread "HttpClient-2-SelectorManager"
[...]

PR #3437, opened 2025-10-26
Temporarily use tycho-packaging-plugin with plexus-archiver 4.10.0

HannesWell:
Workaround for
- https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/issues/3436

until the bug is fixed and the fixed version is integrated in Tycho.


HannesWell:
The test in PDE showed that this workaround seems to be effective.

PR #3460, opened 2025-11-04
Exclude com.ibm.icu from deprecation report

akurtakov:
As it's third party lib it doesn't play by the project rules thus just polutes https://download.eclipse.org/eclipse/downloads/drops4/I20251103-1800/apitools/deprecation/apideprecation.html


akurtakov:
Merge as this report is generated only by I-build.

PR #3465, opened 2025-11-08
[Build] Remove explicit configuration of (now) default archive formats

HannesWell:
Creating `zip` files on Windows and `tar.gz` on Unix is the default in Tycho for quite some time now.


HannesWell:
Currently this also tests the attempted I-build failure fix
- https://github.com/eclipse-platform/eclipse.platform.ui/pull/3521

PR #3473, opened 2025-11-11
Migrate remaining HTML files in org.eclipse.platform.doc.user to HMTL5

ptziegler:
Contributes to


akurtakov:
In book.css there is class table.border which will save you a lot of copy/paste in different files if you're interested to use it.

ptziegler:
> In book.css there is class table.border which will save you a lot of copy/paste in different files if you're interested to use it.

Thanks for the hint. That made things much cleaner :smile:

PR #3477, opened 2025-11-13
Fix o.e.sdk.tests.feature.group -> o.e.sdk.tests 4.38.0.v20251113-0510

merks:


merks:
@iloveeclipse  @akurtakov 

I don't know if this will help, but I assume it does no harm.

See details here:

- https://github.com/eclipse-equinox/p2/pull/968#issuecomment-3528205921
- https://github.com/eclipse-equinox/p2/pull/968#issuecomment-3528234505

PR #3480, opened 2025-11-14
[Build] Redesign build of Eclipse products and P2-repositories

HannesWell:
Migrate all Eclipse Products and P2-repositories build in this repository to a default Tycho build layout with one product or site per Maven module.
This simplifies the configuration and enables better parallelization during the build.

This also implies the following changes
- Sources are removed (again) from the eclipse-platform product
- The Mac entitlements are unified into one file
- The not effective windows-signing via maven-antrun-plugin is removed from the `EclipseRTOSGiStarterKit` product
- The 'org.eclipse.equinox.executable' feature is not contained in the zipped 'org.eclipse.platform' repository anymore, which is part of each build drop.

~This is currently an work-in-progress early draft, but I've created it already to be able to test it on the Eclipse infrastructure with full signing.~

TODOs:
- [x] Unify/clean-ups icons (who needs Solaris icons nowadays?)
- [x] Check PGP signing: Can we just sign the artifacts individually instead of the build products and p2-repos?
- [x] Complete adaption of I-build script
- [x] Introduce eclipse projects for products and sites


HannesWell:
> * Check PGP signing: Can we just sign the artifacts individually instead of the build products and p2-repos?

@merks and/or @laeubi can you tell if it would be simply possible to sign each project individually and just assemble the signatures together with other P2 metadata when assembling the repository?
This would avoid signing for each repo and could also improve the parallelism of the signing process.

merks:
I believe the associated PGP signatures propagate wherever the artifact goes, including into the artifact repository (artifacts.xml) of a product.

HannesWell:
And I also wonder if we should introduce separate projects (i.e. `.project` files) for the `products` and `sites` folder (like for example `eclipse.platform.releng.tychoeclipsebuilder`) or if it's just fine to have them in a folder of the root project of this repository?

HannesWell:
Although the pgp signing of each product's own repository happens before the product materialization the SDK and platform product don't contain the expect PGP signatures: 
```
[INFO] --- tycho-gpg-bump:5.0.1:sign-p2-artifacts (default) @ org.eclipse.platform.ide ---
...
[INFO] --- tycho-p2-director:5.0.1:materialize-products (materialize-products) @ org.eclipse.platform.ide ---
```

I'm investigating, but if anybody knows why this happens or has at least an idea: any help would be welcome.

HannesWell:
> Although the pgp signing of each product's own repository happens before the product materialization the SDK and platform product don't contain the expect PGP signatures:

I expect a fix to be
- https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/pull/3549

PR #3482, opened 2025-11-14
[Build] Unify and clean-up eclipse product executable icons

HannesWell:
Explicitly use the eclipse.ico icon for the Windows executable, even though the executable binary produced by Equinox contains that icon already by default in the eclipse case.
Unify the name of the linux icon,  too.


HannesWell:
> What is the benefit here?

I'd say consistency across platforms. It's more difficult to understand why Windows is different, which would also make it more difficult to update the icon if desired.

> This will likely invalidate the signature of the `eclipse.exe` as Tycho will try to replace the icon (again) and I'm not sure it is really a no-op then...

The Windows executable is signed during the build anyways. So it the signature should stay intact.

HannesWell:
> If we could convince the signing service to process a zip containing executables, that would helpful (I think). I might have an opportunity to discuss that next week...

As far as I understood the documentation of the `eclipse-winsigner-plugin` used in 
https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/blob/45c733165ed55680555148a1ed12d4ae1c413dc5/eclipse.platform.releng.tychoeclipsebuilder/eclipse.platform.repository/pom.xml#L217-L229

... it just signs the exe file(s). And if that happens before the zip is packed it should be fine, shouldn't it?
Of course then the question is which file is added to the P2 repository? I assume that's what you mean when asking to process a zip? The zips in the `binaries` folder of the repo that contain the two executables for Windows (the `.exe` and the `c.exe`).
After signing, we probably would also have to fix the metadata to adjust the checksums (there is a Tycho mojo for that).
But yes, that's all very convoluted and cumbersome.

HannesWell:
> I might have an opportunity to discuss that next week...

So do you think there is a chance for 'custom' images in the foreseeable future with proper signing?
Do you therefore suggest to keep the Windows icon or are you fine with this change?

PR #3485, opened 2025-11-16
Html cleanups

akurtakov:
Contributes to
https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/issues/3275 .
Hanldes o.e.jdt.doc.user/reference/menus.


akurtakov:
Around ~300 problems less . Merging as the other verifications will not check anything useful for this check.

PR #3508, opened 2025-11-25
[RelEng] Ignore misaligned version of org.eclipse.sdk.tests feature

HannesWell:
The version of the org.eclipse.sdk.tests feature was increased twice in this release cycle and therefore isn't properly updated in the release preparation anymore.
To unbreak the release preparation it's misaligned version is temporarily ignored/accepted.

Caused by
- https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/pull/3399

With the preparation of the upcomming release cycle this should be reverted again (and the version of the `org.eclipse.sdk.tests` feature updated to `4.39`.


HannesWell:
This a required but RelEng only change.
Therefore I'm submitting it despite the current freeze period.

PR #3527, opened 2025-11-29
Align version of org.eclipse.jdt.doc.isv with release version

HannesWell:
this simplifies the RelEng process because now bundle's version is automatically increment as part of the usual release preparation. This is already the case for the other doc-bundles that contain New&Noteworthy links (and are updated too during the preparation of a new development cycle).

At the moment the bundles version has to be incremented manually:
- https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/pull/3525

And there are valid objections against running the version increment check for the containing folders too:
- https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/pull/3293

@akurtakov, @iloveeclipse or @jarthana and @mpalat, any objections?


akurtakov:
Makes total sense. Current version is totally useless artifact and having one more thing automated is more than welcome.

PR #3557, opened 2025-12-10
Disable platform setup to redirect 4.37 to 4.38 and 4.38 to 4.39

merks:
- Flesh in the values needed for the end of the 4.39 cycle


merks:
Has no build impact.  Merging immediately.

PR #3549, opened 2025-12-07
[Build] Materialize products from local repository to add pgp signatures

HannesWell:
By default products are materialized from the target-platform, but the `pgp-sigatures` produced in a previous execution of the `tycho-pgp-plugin` are only added to the local repository of the product (in it's `target/repository` directory).
In order to make sure that the produced `pgp-sigatures` are included into the products (i.e. their embedded repositories) the later repository has to be the source of the materialization.

Follow-up on
- https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/pull/3480

Similar to (errors repeat themself...)
- https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/pull/133

PR #3559, opened 2025-12-11
[Build] Add back sources to Eclipse-SDK product

HannesWell:
The sources of all contained bundles were accidentally removed from the eclipse-SDK product in
- https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/pull/3549

That is because the p2-repository assembled for the eclipse-SDK product didn't contain sources before, as that repository is not used further. But now that the product is materialized from that p2-repository (since that mentioned change), sources went missing in the product itself too.

Consequently the property `repository.includeAllSources` must now be true if `product.installSources` is true. Otherwise the latter has no effect. If no sources are available in the source-repository, of course Tycho cannot install them.

Additionally remove dependencies again from eclipse-junit-tests product. They were added unintentionally in the mentioned change, too.

Fixes
- https://github.com/eclipse-jdt/eclipse.jdt.core/issues/4684

### DIFF
--- a/products/eclipse-junit-tests/pom.xml
+++ b/products/eclipse-junit-tests/pom.xml
@@ -39,7 +39,17 @@
       </resource>
     </resources>
 
-
+    <pluginManagement>
+      <plugins>
+        <plugin>
+          <groupId>org.eclipse.tycho</groupId>
+          <artifactId>tycho-p2-repository-plugin</artifactId>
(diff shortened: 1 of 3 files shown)


HannesWell:
I just aborted the scheduled I-build so this can be included too.
Will restart it when this is submitted.

PR #3562, opened 2025-12-11
Version bump for 4.39 stream of org.eclipse.test feature

HannesWell:
The master (and probably also the I-build) fails because of this.


HannesWell:
At least based from the time-stamp this chain of problems seems to be caused by
- https://github.com/eclipse-equinox/equinox/pull/1220

Probably the change in the org.eclipse.equinox.core.feature causes quite some follow-ups.

But I'm giving up on this now. Everyone is invited to help applying all the other bumps that still will be required until this succeeds.

HannesWell:
Ok, it finally succeeded. New I-build is triggered.
Good night.

PR #3577, opened 2025-12-23
Make o.e.test Screenshots work when Platform is not there

akurtakov:
Needed by SWT tests screenshoting.


akurtakov:
The change to Screenshots I did was to allow tests that run standalone (like SWT running tests with maven surefire not tycho) to still save screenshots  in "default" folder instead of failing when Platform (org.eclipse.core.runtime.jar) is not there.

PR #3581, opened 2025-12-27
Restrict updates to ignore poor BND dependencies

merks:
https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/pull/3579


laeubi:
> There's really no way to test this without committing it is there?

You can just execute the following maven command in the root of the aggregator repo:

`mvn -ntp --non-recursive tycho-version-bump:update-target@update-target`

then check if/what was changed in the target file looks like what you expect it to be.

PR #3590, opened 2026-01-04
Add org.eclipse.terminal.test to daily build tests

HeikoKlare:
The terminal tests are missing in the I-Build tests. This change adds the according configuration.

This requires the addition of an according `test.xml` to the test plugin first via:
- https://github.com/eclipse-platform/eclipse.platform/pull/2406

### DIFF
--- a/eclipse.platform.releng.tychoeclipsebuilder/eclipse/publishingFiles/testManifest.xml
+++ b/eclipse.platform.releng.tychoeclipsebuilder/eclipse/publishingFiles/testManifest.xml
@@ -174,6 +174,9 @@
       name="org.eclipse.team.tests.core"
       type="test" />
     <logFile
+      name="org.eclipse.terminal.test"
+      type="test" />
+    <logFile
       name="org.eclipse.tests.urischeme"
       type="test" />
     <logFile
--- a/eclipse.platform.releng/features/org.eclipse.sdk.tests/feature.xml
+++ b/eclipse.platform.releng/features/org.eclipse.sdk.tests/feature.xml
@@ -391,4 +391,8 @@
          id="com.github.weisj.jsvg"
          version="0.0.0"/>
 
+   <plugin
+         id="org.eclipse.terminal.test"
+         version="0.0.0"/>
+
 </feature>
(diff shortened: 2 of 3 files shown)


HannesWell (approved):
Looks good to me, thanks.

In general I think we should avoid the need for a `testManifest.xml` and just drive the list of expected tests dynamically.

HeikoKlare:
From my understanding, the version increment workflow fails because it is executed on the `org.eclipse.releng` folder, thus the build of the features in that folder consumes the included plugins from a referenced p2 repository. This will be the I-Build repository that does currently not contain the `org.eclipse.terminal.test` bundle (e.g. https://download.eclipse.org/eclipse/updates/4.39-I-builds/I20260104-1800/), which is why the build fails accordingly:
```
Cannot resolve dependencies of project eclipse.platform.releng:org.eclipse.sdk.tests:eclipse-feature:4.39.0-SNAPSHOT
Error:   with context {osgi.os=linux, osgi.ws=gtk, org.eclipse.update.install.features=true, org.eclipse.jdt.buildtime=true, osgi.arch=x86_64, org.eclipse.update.install.sources=true}
Error:    Software being installed: org.eclipse.sdk.tests.feature.group 4.39.0.qualifier
Error:    Missing requirement: org.eclipse.sdk.tests.feature.group 4.39.0.qualifier requires 'org.eclipse.equinox.p2.iu; org.eclipse.terminal.test 0.0.0' but it could not be found: See log for details
```

With the test bundle being included in `org.eclipse.sdk.tests` via this PR, it will be promoted to the I-Build's p2 repository due to the inclusion via:
https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/blob/77f4e89deb3155ebd93bb918e326cdc2d0c47567/sites/eclipse-platform-repository/category.xml#L3-L5

@HannesWell is this assessment correct, so that we can/have to ignore the version increment workflow failure here?

HannesWell:
> If I am not mistaken, this is caused by the terminal not being part of the SDK, such that the production bundle to be tested by the added test bundle is not present in the I-Build test product.

Yes, the director logs seems to confirm this:
https://download.eclipse.org/eclipse/downloads/drops4/I20260105-1800/testresults/ep439I-unit-linux-x86_64-java21_linux.gtk.x86_64_21/directorLogs/director-org.eclipse.terminal.test.log

PR #3615, opened 2026-01-17
Try to update jetty-servlet-api from 4.0.6 to 4.0.8

merks:


laeubi:
The Java Version/BREE was a red herring ... the actual problem is that the jetty-servlet bundle is not passing to the resolver at all and only the source bundle is there, so for some reason it does not pass through some check...

PR #3638, opened 2026-02-03
Remove Outdated Screenshots From "Keys" Help Page

BeckerWdf:


akurtakov (commented):
Outdated screenshots are a problem in many many places in the documentation. I agree that showing outdated screenshots is bad (mitigated a bit by the fact that some screenshots are are on Mac or Linux so overall lack of consistency makes it hard to spot it's outdated rather than another OS screenshot) but having such massive amount of text for such a common feature is not good either (will anyone read few pages for changing shortcut?).
With the above said I leave it to you to decide what to so as it's neither better nor worse IMO.

akurtakov:
Rewriting the page to be smaller/easier to understand and follow.

PR #3642, opened 2026-02-05
Remove reference to releng.buildtools form the SDK configuration

merks:


merks:
Has no build impact.  Merging.

PR #3651, opened 2026-02-14
[RelEng] Fix determination of baseline drop and git diff in promotion

HannesWell:
Fix the computation of the baseline tag by ignoring the just promoted drop.
Compare against the git-tag of the input dropId since the tag of the promoted build is not yet applied at the time when the git comparison is performed.
Additionally unify the computation of the git-tag for the promoted drop.


HannesWell:
I'll have a look at it over the weekend and will fix it. 
Thanks for the report.

PR #3656, opened 2026-02-18
Add org.eclipse.core.tools to org.eclipse.releng.testsIU

merks:
- This is the category labeled `Eclipse Tests, Tools, Examples, and Extras`  with description `Collection of Misc. Features, such as unit tests, SWT and e4 tools, examples, and compatibility features not shipped as part of main SDK, but which some people may desire in creating products based on previous versions of Eclipse.`.

https://github.com/eclipse-platform/eclipse.platform/issues/2457

### DIFF
--- a/sites/eclipse-platform-repository/category.xml
+++ b/sites/eclipse-platform-repository/category.xml
@@ -36,6 +36,9 @@
       <category name="org.eclipse.releng.pde.categoryIU"/>
    </feature>
    <feature id="org.eclipse.pde.osgitest.dependencies.feature"/>
+   <feature id="org.eclipse.core.tools">
+      <category name="org.eclipse.releng.testsIU"/>
+   </feature>
    <bundle id="jakarta.annotation-api" version="1.3.5"/>
    <bundle id="jakarta.inject.jakarta.inject-api" version="1.0.5"/>
    <bundle id="org.eclipse.equinox.slf4j"/>


merks:
> If you want, just trigger IBuild right now, we will see. If it fails, we can revert today.

I'll give @HannesWell or @laeubi  some time to comment.  With the pomless structure I'm not really sure what drives things being built, but given it's parallel to the core.runtime feature, I expect it to just be built...

HannesWell:
> With the pomless structure I'm not really sure what drives things being built, but given it's parallel to the core.runtime feature, I expect it to just be built...

It can either be a an aggregator module, which indeed automatically discovers all PDE type of projects (Bundles, Features, p2-repos, products) in a directory and only that aggregator-directory has to be listed in a `pom.xml`.
Or, and this seems to be the case here, a project (of whatever type) is listed explicitly in a `pom.xml`:
https://github.com/eclipse-platform/eclipse.platform/blob/acd8d2e28426241909c776d63a8b2d7391c94428/runtime/pom.xml#L26

> Is there some chicken-end-egg problem here where we really need to do an I-build for this problem to go away?

No, it looks like the feature is just not build.
We can either update the linked pom, to list the other feature too or, and that's my suggestion, change the listing and just include the `features` directory. Then all included projects are build, which in this case, just adds the desired feature.
Once that's on the master of the `eclipse.platform` repository this build should succeed.

> It looks like we really haven't been building the feature, but that doesn't make sense:

Sometimes the unbelievable happens :)

PR #3688, opened 2026-02-28
Update platform setup to redirect 4.38 to 4.39 and 4.39 to 4.40

merks:


merks:
This does not affect the build. Merging.

PR #3690, opened 2026-02-28
Test if build aggregator PR build fails

merks:


merks:
The build is stuck at this point for going on 10 minutes:

image

HannesWell:
> The build is stuck at this point for going on 10 minutes:

I experienced this yesterday evening too when running the first I-build.
My impression is that some pods in the Jenkins kubernetes cluster are terribly slow.

After a restart the build just worked fine.

PR #3696, opened 2026-03-03
Limit PR creation in checkDependencies.yml to 10 open PRs

laeubi:
Add PR limiting mechanism matching the pattern in cleanCode.yml:
- Add list-prs step to count open PRs with 'dependency-check' label
- Add if conditions to skip steps when 10+ PRs already exist
- Label created PRs with 'dependency-check' for tracking

Marking this as a draft for now as we need to first cleanup the queue on P2 before we can apply the limits!

### DIFF
--- a/.github/workflows/checkDependencies.yml
+++ b/.github/workflows/checkDependencies.yml
@@ -73,32 +73,46 @@ jobs:
       max-parallel: 1
       fail-fast: false
     steps:
+      - name: List number of open PRs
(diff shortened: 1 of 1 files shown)


akurtakov:
Should this one be rebased and merged or there is no interest in it and should be closed?

laeubi:
I think it should be merged in any case but now has conflicts I need to resolve.

PR #3708, opened 2026-03-09
Update for first Orbit 4.40 milestone

merks:
- Update Lucene to 10.4.0
- Update httpcomponents.core5 to 5.4.2.


akurtakov:
Let's merge this one as Orbit verified them and the EF licensecheck is unreliable and there is more than enough time till release.

PR #3756, opened 2026-04-10
[Build] Remove unnecessary resources-plugin executions and set encoding

HannesWell:
The `filter-resources` execution of the `maven-resources-plugin` seems not any effect at any project and the specified directory is always just skipped.

Additionally specify the the encoding explicitly when filtering resources, to prevent the INFO from below.
Java properties files are encoded using ISO-8859-1 encoding. Therefore, when filtered by the `maven-resources-plugin`, these files should be written with that encoding.
```
The encoding used to copy filtered properties files has not been set.
This means that the same encoding will be used to copy filtered properties files as when copying other filtered resources.
This might not be what you want! Run your build with --debug to see which files might be affected.
Read more at
https://maven.apache.org/plugins/maven-resources-plugin/examples/filtering-properties-files.html
```

### DIFF
--- a/eclipse-platform-parent/pom.xml
+++ b/eclipse-platform-parent/pom.xml
@@ -277,18 +277,12 @@
         <artifactId>maven-resources-plugin</artifactId>
         <executions>
           <execution>
-            <id>filter-resources</id>
+            <id>saveproperties</id>
             <goals>
-              <goal>resources</goal>
+              <goal>copy-resources</goal>
             </goals>
             <phase>process-resources</phase>
-            <configuration>
-              <outputDirectory>${project.build.directory}/resources</outputDirectory>
-              <escapeString>\</escapeString>
-            </configuration>
-          </execution>
-          <execution>
-            <id>saveproperties</id>
+            <inherited>false</inherited>
             <configuration>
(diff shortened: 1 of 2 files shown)


HannesWell:
I found one actual usage of the removed execution:
```
[INFO] --- resources:3.5.0:resources (default-resources) @ org.eclipse.jdt.ui.junit.sampleproject ---
[INFO] Copying 7 resources from src/main/resources to target/classes
[INFO] 
[INFO] --- resources:3.5.0:resources (filter-resources) @ org.eclipse.jdt.ui.junit.sampleproject ---
[INFO] Copying 7 resources from src/main/resources to target/resources
```
But since that project has packaging type `jar` the default lifecycle binding implies the default `resources:resources` execution:
https://github.com/eclipse-jdt/eclipse.jdt.ui/blob/26b3e0973e9870995b1cf72e8d68ea807a0bf146/org.eclipse.jdt.ui.junit.sampleproject/pom.xml#L27

... which should be suitable for the project's content:
https://github.com/eclipse-jdt/eclipse.jdt.ui/tree/master/org.eclipse.jdt.ui.junit.sampleproject/src/main/resources/

Therefore the removal should still be fine.

PR #3758, opened 2026-04-10
[I/Y-Build] Replace base-builder installation with tycho eclipse-run

HannesWell:
Replace creation and use of a base-builder Eclispe installation, by invocations of the `eclipse-run` goal of the `tycho-eclipse` Maven plugin.

Also remove unused `eclipserun-repo` property.


JenkinsJobs/Builds/build.jenkinsfile
```diff
@@ -482,6 +467,7 @@ pipeline {
 									# Build Equinox Launcher archives
 									pushd $CJE_ROOT
 									mkdir -p $ECLIPSE_BUILDER_DIR/equinox/$TMP_DIR
+									#FIXME: Replace this too!!
```
HannesWell: Is is an important, currently missing, part.
But I'll probably resolve that in a separate PR in advance.

HannesWell: This is now done in
- https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/pull/3759

JenkinsJobs/Builds/build.jenkinsfile
```diff
@@ -413,7 +399,7 @@ pipeline {
 									java \
 										-DdropDirectory=${DROP_DIR}/${BUILD_ID} \
 										${SCRIPTS}/releng/CompilerSummaryGenerator.java
-								'''
+								'''.stripIndent().replace('\t', ' ')
```
HannesWell: This is a workaround for the fact that the `eclipse-run` Tycho Mojo splits the [`appArgLine`](https://tycho.eclipseprojects.io/doc/5.0.2/tycho-eclipse-plugin/eclipse-run-mojo.html#appArgLine) and [`argLine`](https://tycho.eclipseprojects.io/doc/5.0.2/tycho-eclipse-plugin/eclipse-run-mojo.html#argLine) only at spaces, respectively the called `CommandLineUtils.translateCommandline` does it:
https://github.com/eclipse-tycho/tycho/blob/d219bd5a3ed07e3dcdf34a0706bda2aa0b78bc6b/tycho-eclipse-plugin/src/main/java/org/eclipse/tycho/eclipserun/EclipseRunMojo.java#L480-L486

But for example the bash reference manual indicates that space, tab and newline is considered as separator:
https://www.gnu.org/software/bash/manual/bash.html#Word-Splitting-1
```
If IFS is unset, word splitting behaves as if its value were <space><tab><newline>, and treats these characters as IFS whitespace.
```
As far as I can tell, Windows `cmd` behaves similar. I'm not sure about `powershell`.

I know that the documentation discourages from using these parameters and to use the explicit list equivalents instead, but I'm not aware how to fill these lists from the CLI, so I assume we are stuck with `appArgLine` and `argLine`here.

@laeubi do you have a better idea?
Or is this something we should 'fix' in Tycho (i.e. also split at tabs) or even upstream in `org.codehaus.plexus.util.cli.CommandLineUtils`?

merks: OMG. That tab thing wasted so many hours of my life. Completely confusing, unexpected, and arguably wrong. I recall the two things being deprecated but I begged to undo that because it is really hard to synthesize a sequence of XML elements but really easy to  just create space separated string of arguments. Just don’t format and end up with tab in the string.

HannesWell: Yes it's really not trivial to identify this issue. I reported it upstream:
- https://github.com/codehaus-plexus/plexus-utils/issues/329

In this case, I think I could/will solve this by having two explicit executions configured in the parent-pom, one for each use-case. Requiring certain type of white-space seems to be too brittle to me.

HannesWell: > In this case, I think I could/will solve this by having two explicit executions configured in the parent-pom, one for each use-case. Requiring certain type of white-space seems to be too brittle to me.

This is now implemented.

PR #3763, opened 2026-04-13
Enable/Update Ybuild for Java27

MohananRahul:


HannesWell:
Thanks Rahul for this, but this missed the permanent enablement of the Y-builds for 4.40. I just did this with
- https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/pull/3781

Without it, the Y-build gets disabled each time the seed job runs.

Furthermore I deleted the now unused `Y-build-4.39` job and its tests.

PR #3767, opened 2026-04-14
[I/Y-Build] Unify and streamline build and naming of products

HannesWell:
Build OSGi Starter Kit for all supported platforms.
FYI @tjwatson and @laeubi (but I assume almost nobody really cares).

Unify names of Eclipse products built for Windows, i.e. use the same `<ws>-<os>-<arch>` naming schema like for Linux and macOS, even though `ws` and `os` segments are the same.
I assume nobody really cares about the exact names either (except that existing scripts would not require special handling for Windows anymore).
We just have to adjust the api-tools generation after next release at
- https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/blob/8210d2514f24f669091a97b8575e76e3e82e2127/eclipse-platform-parent/pom.xml#L557-L559
- https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/blob/8210d2514f24f669091a97b8575e76e3e82e2127/eclipse.platform.releng.tychoeclipsebuilder/eclipse/buildScripts/api-tools-builder.xml#L187

if it's then still using the Eclipse SDK product for Windows only (I hope not).


HannesWell:
> We just have to adjust the api-tools generation after next release at
> 
>     * https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/blob/8210d2514f24f669091a97b8575e76e3e82e2127/eclipse.platform.releng.tychoeclipsebuilder/eclipse/buildScripts/api-tools-builder.xml#L187
> 
In fact that particular change is already necessary now.

PR #3769, opened 2026-04-15
Add skija 1.7.0 to eclipse-sdk target

DenisUngemach:
For using skija in SWT, the maven jars must be loaded. 

Are they added here in the right way?

Required for:
https://github.com/eclipse-platform/eclipse.platform.swt/pull/3231#issuecomment-4242232863


merks:
See my comment here:

https://github.com/swt-initiative31/eclipse.platform.releng.aggregator/pull/1

merks:
@DenisUngemach 

I'll close this in favor of https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/pull/3800

PR #3770, opened 2026-04-16
Do not use new Color(display,...) in docs

akurtakov:
Adjust docs to not recommend what is discussed to be deprecated and is totally not needed.
Contributes to eclipse-platform/eclipse.platform.swt#3233 .


akurtakov:
License vetting return error 504 again. Ignore for this PR as no dependencies change.

PR #3772, opened 2026-04-16
[Build] Remove test Plug-ins from Eclipse P2-repository

HannesWell:
Tests are not intended for downstream consumers at all and therefore just increases the repositories size without any value. A smaller repository can also accelerates the Maven build itself due to an increased parallelism and is of course faster to transfer between build- and storage-server.

Besides all 'test'-Plug-ins, this also removes the `org.eclipse.sdk.tests` feature from the Eclipse p2 repository.

If one really needs the built test Plug-in jars, one can download them with the `eclipse-Automated-Tests` zip from each build drop website.

The only production bundle that is currently removed is `org.apache.commons.commons-fileupload` since it's seems to be only required by tests and only optionally required by `org.eclipse.equinox.http.servlet`. I aim to avoid that via
- https://github.com/eclipse-equinox/equinox/pull/1274

Since this is change affects the entire Eclipse top level project repository I'd like to make all project leads aware and hope everyone is fine with it:
- for Platform: @akurtakov, @iloveeclipse, @HeikoKlare
- for Equinox: @tjwatson and @laeubi
- for JDT: @jarthana and @mpalat
- for PDE: myself (I'm fine with it).

### DIFF
--- a/sites/eclipse-platform-repository/category.xml
+++ b/sites/eclipse-platform-repository/category.xml
@@ -1,8 +1,5 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <site>
-   <feature id="org.eclipse.sdk.tests">
-      <category name="org.eclipse.releng.testsIU"/>
-   </feature>
    <feature id="org.eclipse.equinox.p2.sdk" version="0.0.0"/>
    <feature id="org.eclipse.equinox.p2.discovery.feature" version="0.0.0"/>
    <feature id="org.eclipse.core.runtime.feature" version="0.0.0"/>
@@ -49,9 +46,6 @@
    <iu id = "org.eclipse.platform.ide">
      <category name="org.eclipse.platform.ide.categoryIU"/>
    </iu>
-   <iu id = "eclipse-junit-tests">
-     <category name="org.eclipse.releng.testsIU"/>
-   </iu>
    <category-def name="org.eclipse.equinox.target.categoryIU" label="Equinox Target Components">
       <description>
          Features especially useful to install as PDE runtime targets.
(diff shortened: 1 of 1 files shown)


mpalat:
+1 from JDT. @HannesWell  - Thanks for the suggestion.

akurtakov:
I like the idea and it's better to check sooner whether anything breaks due to this removal. Go from Platform.

merks:
How are the tests available for testing after the  initial I-build step completes?

HannesWell:
> How are the tests available for testing after the initial I-build step completes?

The test Plug-ins together with the `org.eclipse.sdk.tests` feature  are built into the `eclipse-junit-tests` product:
https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/blob/96c2d381cd25c102f2cd4832e6523127984d6663/products/eclipse-junit-tests/eclipse-junit-tests.product#L152-L163

That product respectively actually only the p2-repo build from it is contained in the `eclipse-Automated-Tests-<BUILD_ID>.zip` that's provided on each build's drop website. It's content is defined in the [products/eclipse-junit-tests/src/main/assembly/assembly.xml](https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/blob/96c2d381cd25c102f2cd4832e6523127984d6663/products/eclipse-junit-tests/src/main/assembly/assembly.xml).

That zip is then downloaded and extracted during the bootstrapping phase of I/Y-build tests:
https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/blob/00d59267887b473f19f95efb33440639defa145d/production/testScripts/runTests2.xml#L79-L83

To my knowledge, the I/Y-build tests don't use the Eclipse updates p2-repository at all to receive the test Plug-ins.

Since we have +1 from three of the four TLPs (counting myself for PDE), I plan to submit this PR this (late European) evening, unless somebody objects until then, to have time over the weekend to react if necessary.

merks:
Thanks for the details.  It makes good sense to proceed!

HannesWell:
Thanks everyone, then let's have this now. 

~~Until a solution is decided about how to keep Apache fileupload the corresponding test of `org.eclipse.equinox.http.servlet` will probably fail temporarily.~~
Edit: In fact `org.apache.commons.commons-fileupload` is included in the `eclipse-Automated-Tests-<BUILD_ID>.zip` and therefore available to tests:
https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/blob/e5e1abf7c2ad16b4ae78e97187d746f3a3b85d07/eclipse.platform.releng/features/org.eclipse.sdk.tests/feature.xml#L314-L316

HannesWell:
In the context of https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/issues/3784, I also noticed that the automated version bump will probably not work for test projects anymore. But since tests are now not in the baseline anymore, they also won't have (or at least shouldn't need) version problems anymore.
And applying strict versioning to tests probably has little value in general anyways.

PR #3776, opened 2026-04-17
[I/Y-Build] Use new names of Windows products in test resource download

HannesWell:
This was missed in

and lead to failures of the Windows tests tonight.


production/testScripts/runTests2.xml
HannesWell, line 102: This download has to be adapted after the next release.

PR #3777, opened 2026-04-17
Silense deprecation report for jetty.session

akurtakov:
As reported in
https://download.eclipse.org/eclipse/downloads/drops4/I20260416-1800/apitools/deprecation/apideprecation.html


akurtakov:
Merge as this will not get verified by anything else but I-build.

PR #3778, opened 2026-04-18
[I/Y-Build] Use new names of Windows products in test bootstrap

HannesWell:
This was missed in

respectively in
- https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/pull/3776


HannesWell:
Only I-builds will verify this (the changed file is packed into the eclipse-Automated-Tests zip), thus I'm submitting this immediately and will trigger another I-build.

I also checked if there are more potentially affected locations but didn't find I any (should have done that with my previous round of fixes).

PR #3780, opened 2026-04-18
[I/Y-Build] Use new name for previous release Windows product in test

HannesWell:
The new naming schema is expected for all products. Therefore the previous release platform product has to use the new schema locally too, even if it's not named like that at the download page.

This was missed in

respectively in
- https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/pull/3778


HannesWell:
Again, like in https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/pull/3778#issuecomment-4273097708, only I-build (tests) will verify this.
Submit this immediately and restart another test (aborting the currently running ones).

Issue #3784, opened 2026-04-19
Baseline comparision failes for tests removed from Eclipse P2-repository


HannesWell:
After the removal of test Plug-ins from the Eclipse p2-repository (via https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/pull/3772) many verification builds with baseline comparision errors on test projects, like the one below.

This can also be verified locally by running:
`mvn clean verify -DskipTests=true "-Dcompare-version-with-baselines.skip=false"`

Actually I assumed that artifacts not present in the baseline are simply ignored by the `compare-version-with-baselines` Mojo. But I have to check the details.

As last resort the execution of `compare-version-with-baselines` could forcefully be skipped for all test projects, but it would be cumbersome to configure them all.
Another option could be, if skipping missing artifacts doesn't work, to configure `compare-version-with-baselines` to skip projects of certain type. For this case specifically to skip projects with type `eclipse-test-plugin`.

```
 [ERROR] Failed to execute goal org.eclipse.tycho.extras:tycho-p2-extras-plugin:5.0.3-SNAPSHOT:compare-version-with-baselines (compare-attached-artifacts-with-release) on project org.eclipse.jdt.core.tests.builder.mockcompiler: Baseline and reactor have the same fully qualified version, but different content
 [ERROR] different
 [ERROR]    META-INF/ECLIPSE_.RSA: present in baseline only
 [ERROR]    META-INF/ECLIPSE_.SF: present in baseline only
 [ERROR] -> [Help 1]
 org.apache.maven.lifecycle.LifecycleExecutionException: Failed to execute goal org.eclipse.tycho.extras:tycho-p2-extras-plugin:5.0.3-SNAPSHOT:compare-version-with-baselines (compare-attached-artifacts-with-release) on project org.eclipse.jdt.core.tests.builder.mockcompiler: Baseline and reactor have the same fully qualified version, but different content
 different
    META-INF/ECLIPSE_.RSA: present in baseline only
    META-INF/ECLIPSE_.SF: present in baseline only

     at org.apache.maven.lifecycle.internal.MojoExecutor.doExecute2 (MojoExecutor.java:333)
     at org.apache.maven.lifecycle.internal.MojoExecutor.doExecute (MojoExecutor.java:316)
     at org.apache.maven.lifecycle.internal.MojoExecutor.execute (MojoExecutor.java:212)
[...]

PR #3785, opened 2026-04-19
[Build] Skip baseline version comparision for eclipse-test-plugin

HannesWell:
Fixes
- https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/issues/3784

Additionally I wonder if API checks should be skipped too.
IIRC I saw some projects explicitly disabling it, I'll check if some test plugins were among them.

### DIFF
--- a/eclipse-platform-parent/pom.xml
+++ b/eclipse-platform-parent/pom.xml
@@ -903,6 +903,35 @@
       </properties>
     </profile>
     <profile>
+      <id>skip-baseline-comparision-of-tests</id>
+      <activation>
+        <property>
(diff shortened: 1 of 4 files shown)


HannesWell:
The previous build verified that this works with for the entire aggregator build once the pending changes from JDT are submitted:
- https://github.com/eclipse-jdt/eclipse.jdt.core/pull/5012
- https://github.com/eclipse-jdt/eclipse.jdt.core.binaries/pull/26

Nevertheless submitting this now will at least unblock the verification builds of all other repositories.

PR #3786, opened 2026-04-20
Remove org.eclipse.test.source from o.e.test feature

HannesWell:
Or does anybody know a reason why to keep that particular source bundle?
Git blame doesn't indicate anything.


akurtakov:
My guess would be that back in the days that was done to save having one more feature. Go for it.

merks (approved):
Yes, looks like a remnant.

HannesWell:
Then let's have it. Thanks.

PR #3791, opened 2026-04-22
Silence deprecation warnings from bcutil

akurtakov:
This is beyond project control so doesn't make sense to report it. Can be seen at
https://download.eclipse.org/eclipse/downloads/drops4/I20260421-1800/apitools/deprecation/apideprecation.html


akurtakov:
Purely reporting change that would only affect generation of https://download.eclipse.org/eclipse/downloads/drops4/I20260421-1800/apitools/deprecation/apideprecation.html in I-builds. Merging.

PR #3800, opened 2026-04-29
Add skija to the target platform

merks:


DenisUngemach:
Hello @merks 

I did now the following:

clone your repo and checked out the releng. There load the target :

image


Next, in the manifest of org.eclipse.swt.skia i set dependencies:

image

But there i can add shared, types and **for all the **OS dependent only the source jars, not the real OS-dependent jars**. The OS-dependent jars won't be found.

And then if i execute the SnippetCanvasCompare from /org.eclipse.swt.examples  (in my skia branch), the library for windows won't be loaded:

java.lang.UnsatisfiedLinkError: Library file skija.dll not found in io/github/humbleui/skija/windows/x64/
	at io.github.humbleui.skija.impl.Library._extract(Library.java:160)
	at io.github.humbleui.skija.impl.Library.load(Library.java:101)
	at io.github.humbleui.skija.impl.Library.staticLoad(Library.java:20)
	at io.github.humbleui.skija.DirectContext.<clinit>(DirectContext.java:7)
	at org.eclipse.swt.internal.skia.SkiaGlCanvasExtension.<init>(SkiaGlCanvasExtension.java:70)
	at org.eclipse.swt.internal.canvasext.SkiaCanvasFactory.createCanvasExtension(SkiaCanvasFactory.java:33)
	at org.eclipse.swt.internal.canvasext.ExternalCanvasHandler.createHandler(ExternalCanvasHandler.java:140)
	at org.eclipse.swt.widgets.Canvas.<init>(Canvas.java:85)
	at org.eclipse.swt.examples.skia.SnippetCanvasCompare.main(SnippetCanvasCompare.java:21)



Am i doing something wrong?
To be honest, i have no idea how this releng plugin works and how the target configuration works....

In my target configuration, the bundles seem to be there:
image

merks:
You cannot generally add a requirement to a fragment so it will not be listed when editing the MANIFEST.MF.

For SWT folks added requirements from to host to each fragment so the right fragment is typically added to the launch.    So these two do that:

image


That's not been done for skija (in the bundle that requires the skija shared host bundle, it should do something analogous with filtered requirements on each of the os/arch combinations) so it's highly likely that you need to manually add it to the launch.  But if it's a-stand alone launch, it's not so easy to add a jar.

image

But I'd tried that first just for quick feedback...

I guess you need help with this, but I don't know how to reproduce your setup to provide such help (and I have limited time).

The p2.inf should have stuff like this, one per supported os/arch:
```
requires.1.namespace = org.eclipse.equinox.p2.iu
requires.1.name = io.github.humbleui.skija-windows-x64
requires.1.filter = (&(osgi.os=win32)(osgi.arch=x86_64))
```
Each combination needs a new number, e.g., requires.2.*, requires.3.*

General docs:

https://eclipse.dev/eclipse/markdown/?file=eclipse-equinox/p2/master/docs/Customizing_Metadata.md

DenisUngemach:
Hello @merks 


I just wanted to check whether i get the native libs somehow to run. 
I took your target as described before and i used the official SWT repository. 

Now in the project explorer i imported:
org.eclipse.platform.releng.prereqs.sdk  <-- from your PR.
org.eclipse.swt
org.eclipse.swt.examples
org.eclipse.swt.win32.win32.x86_64


-> To SWT i added in the manifest all humbleui bundles and reexported them.
And in org.eclipse.swt to the p2.inf i added:


requires.10.namespace = org.eclipse.equinox.p2.iu
requires.10.name = io.github.humbleui.skija-windows-x64
requires.10.range = [0.0.0,100.0.0]
requires.10.filter = (&(osgi.os=win32)(osgi.ws=win32)(osgi.arch=x86_64)(!(org.eclipse.swt.buildtime=true)))


But maven install for org.eclipse.swt fails, because the skija dependencies won't be found by the standard tycho build:

[INFO] [1m------------------------------------------------------------------------[m
[INFO] [1;31mBUILD FAILURE[m
[INFO] [1m------------------------------------------------------------------------[m
[INFO] Total time:  10.122 s
[INFO] Finished at: 2026-04-30T13:32:28+02:00
[INFO] [1m------------------------------------------------------------------------[m
[ERROR] Cannot resolve dependencies of project org.eclipse.platform:org.eclipse.swt:eclipse-plugin:3.134.0-SNAPSHOT
[ERROR]  with context {osgi.os=win32, org.eclipse.update.install.features=true, org.eclipse.swt.buildtime=true, osgi.arch=x86_64, org.eclipse.update.install.sources=true, osgi.ws=win32, org.eclipse.jdt.buildtime=true}
[ERROR]   Software being installed: org.eclipse.swt 3.134.0.qualifier
[ERROR]   Missing requirement: org.eclipse.swt 3.134.0.qualifier requires 'osgi.bundle; io.github.humbleui.types [0.2.0,1.0.0)' but it could not be found: See log for details
[ERROR] -> [1m[Help 1][m
[ERROR] 
[ERROR] To see the full stack trace of the errors, re-run Maven with the [1m-e[m switch.
[ERROR] Re-run Maven using the [1m-X[m switch to enable full debug logging.
[ERROR] 
[ERROR] For more information about the errors and possible solutions, please read the following articles:
[ERROR] [1m[Help 1][m http://cwiki.apache.org/confluence/display/MAVEN/ResolverException


-> to examples i added a new snippet, which uses skija:

```
package org.eclipse.swt.examples;

import io.github.humbleui.skija.Surface;

public class SnippetSkia {

	public static void main(String[] args) {

		Surface surface = Surface.makeRasterN32Premul(256, 256);

		surface.getCanvas().clear(0xFFFF0000);


	}

}
```


No syntax error. the code will be executed, but the skia windows lib won't be found:

Exception in thread "main" java.lang.UnsatisfiedLinkError: Library file skija.dll not found in io/github/humbleui/skija/windows/x64/
	at io.github.humbleui.skija.impl.Library._extract(Library.java:160)
	at io.github.humbleui.skija.impl.Library.load(Library.java:101)
	at io.github.humbleui.skija.impl.Library.staticLoad(Library.java:20)
	at io.github.humbleui.skija.Surface.<clinit>(Surface.java:9)
	at org.eclipse.swt.examples.SnippetSkia.main(SnippetSkia.java:9)




I have no idea how i can fix this, resp make this run. 

---------------------

...
Maybe is should just create a new dependency plugin myself and use classic maven to load the dependencies in there.
...

---------------------

merks:
It's this that's missing so not really missing from the target platform:
```
      <unit id="io.github.humbleui.types" version="0.2.0.v20260416-1000"/>
```
Did you specify a dependency on that in the MANIFEST.MF?  Maybe you don't need to though...

I can help better if I can reproduce what you are doing.  I'm like 99.9% sure you are launching without the necessary library on the classpath, but given I am 0% sure how you are launching, I can only stab in the dark.

Compare what you are doing to the SVG bundle:

image

Do you see all the expected things in the plug-in dependencies in your bundle?  Especially the io.github.humbleui.skija-windows-x64 fragment?

Have you looked at the properties of the launched executable in the debug view to see what's on the classpath?  E.g.,

image
 
Looking closely it seems like you are trying to add this directly to SWT?  We didn't even do that with the SVG support.  It's provided by a fragment.

@HeikoKlare 

What exactly is the plan here? I had assumed that skija would be a separate optional thing, but maybe that's a bad assumption?  Given not all of SWT's platforms are supported, I'm confused about the design goal here.

DenisUngemach:
@merks 
I only want to make it run at all. 
Trying to make it run on the org.eclipse.swt.skia fragment only adds one more complexity layer.

These are in the classpath. 

image

The native jars are not there. Maybe we have to setup a meeting.

merks:
BTW, if you are doing this in the SWT bundle (which confuses me), use this so that it's not ignored a build time either.

```
requires.10.filter = (&(osgi.os=win32)(osgi.arch=x86_64))
```

Add it manually to the launch configuration:

image

Use this to find/copy the location:

image

Yes, we could have a meeting with screen sharing.  But I'm about to leave and tomorrow I assume is a holiday for most everyone...

merks:
Yes,it's all quite tricky.  😦 

I don't know how best this should work at development time.  

I tried this example:

image

I can see that the induced p2 metadata has the additional requirement but that doesn't show up in the plugin dependencies at development time.

The example (attached below) includes two launches, one for standalone and for an application.

I added this folder to the stand alone launch to make it work:

image

The application launcher can be minimized and does know to include the fragment.

image

Here's the example:

[org.example.skija.zip](https://github.com/user-attachments/files/27304046/org.example.skija.zip)

I'm just not sure how to make this simpler for development time standalone launching...

DenisUngemach:
@merks 
For setting up a meeting: 
can you send me a mail to denis.ungemach@sap.com or give me your email adresse.

Adding the windows skija jar to the launch configuration was not possible, because at the location you described is a folder not the jar.

But i added the windows skija jar as implicit dependency in the target and then the windows fragment was found and the skia fragment worked. 

image

Is this ok for you? 
Of course for building with tycho we need the p2.inf configuration. And for finding the skija jars, the target must be configured to find the skija jars at the build process.

merks:
Note that I used Advanced... to add a **folder**

image

I'm not sure why the bundle is unjarred as a folder.  We didn't specify via `Eclipse-BundleShape: dir` to do that.

DenisUngemach:
Oh, sorry... yes. This also works.

PR #3801, opened 2026-05-01
Upgrade jakarta.annotation-api to 3.0.0 in the target platform

merks:


merks:
There is an infrastructure problem with cloning that I've also seen on Oomph ci instance.

PR #3807, opened 2026-05-06
Remove jsvg-systemlogger

merks:


merks:
FYI @HeikoKlare 

This build does not appear to be using the latest SWT commit.

> Error:    Cannot satisfy dependency: org.eclipse.swt.svg 3.132.100.qualifier depends on: java.package; com.github.weisj.jsvg.view [2.0.0,3.0.0)

The actual current lower bound is 2.1.0:

https://github.com/eclipse-platform/eclipse.platform.swt/blob/c1fd7a0bcb45191bfdd5293834cb852fddf3662e/bundles/org.eclipse.swt.svg/META-INF/MANIFEST.MF#L11-L14

I guess we need a real successful I-build to make progress here.  Is that correct @HannesWell?

HeikoKlare:
Great. Should we just kick off another I-Build? Even though the one from today failed because the test jobs were not started, the artifacts were still deployed: https://download.eclipse.org/eclipse/updates/4.40-I-builds/I20260506-0618/
We may do that again so that we can proceed with this PR.

merks:
Hmmm. I did kick off a build but I see here  now that the verification build passed.  So only the license check uses the old content from an I-build. 

So I think I just merge this, abort the I-build and start a new one.

PR #3813, opened 2026-05-10
Make eclipse and equinox website folder names more clear

HannesWell:
Currently the names of the folders containing the sources for the Eclipse and Equinox websites are quite generic within the `sites` folder.
This change has the goal to make the folder intent and content more clear with their name, in the context of the p2-repositories also residing in the `sites` folder.

This is currently a draft since I'm contemplating to change the structure again, by just having one `websites` folder within the `sites` top-level folder and within that `websites` folder have the `eclipse` and `equinox` folder. This way shared files could just reside in the `websites` folder and the `eclipse` folder would not be special compared to the `equinox`. This would make the handling a bit simpler and would also keep nice URLs when testing the websites locally.

Yet another alternative would be a top-level `websites` folder co-located to the `sites` folder, which would separate p2-repos and websites entirely.


HannesWell:
> I'm contemplating to change the structure again, by just having one `websites` folder within the `sites` top-level folder and within that `websites` folder have the `eclipse` and `equinox` folder. This way shared files could just reside in the `websites` folder and the `eclipse` folder would not be special compared to the `equinox`. This would make the handling a bit simpler and would also keep nice URLs when testing the websites locally.

This strategy is now implemented.

Issue #2979, opened 2025-04-08
Streamline I-build test setup.


HannesWell:
Currently the setup of I-build tests is a very convoluted process that involves multiple levels of shell- and ANT-scripts. It's really hard to fully understand the whole setup and I assume nobody does it at the moment.
The starting point is:
https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/blob/0e81c4f903ec9e7c11bd92e85c106d96237c2341/JenkinsJobs/AutomatedTests/I_unit_tests.groovy#L68-L80

To streamline and simplify this, my suggestion is to replace the entire setup by a single (big) java class that is launched as a _Single-File Source-Code Programs_, possible since Java-11 with [JEP-330](https://openjdk.org/jeps/330). If we use Java-22 or later just to run the setup we can even split that in multiple java files, since [JEP-458](https://openjdk.org/jeps/458) and launch them directly.

My suggestion is to start with the `getEBuilder.xml` ANT script and successively move more of the setup into the java class, until nothing is left:
https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/blob/0e81c4f903ec9e7c11bd92e85c106d96237c2341/production/testScripts/bootstrap/getEBuilder.xml

We can even inline the properties files into the (or one of the) java script(s).

The goal would be to have only very few input arguments, mainly just the `buildId` and the `testSuite` and optionally the os/ws/arch (we can try to derive that from system-properties).

The advantages of this would be
- everyone of the active developers of Eclipse knows Java very well
  - not everyone knows ANT or the scripting languages for different OS 
- it would be very simple to launch locally as one just has to run it as plain Java-application from the IDE or CLI and has to specify the mentioned parameters (or just set them temporarily in code). No more need to set up ANT, etc.
  - Currently it's very difficult to run I-builds locally and at least on my Windows machine I never made it to run.  
- It would be easier to debug as the mentioned launch could simply be run in debug mode
- Modern day Java can be really compact when it comes to file-IO (which the setup mainly is), some things are still a bit more verbose to implement in java than in ANT or a shell-script, but some things are even easier.

akurtakov:
I like the proposed workflow. Only addition from my side would be to question the need for every bit we move. E.g. first question is what is Ebuilder and is it truly still needed?

HannesWell:
> I like the proposed workflow.

Great. :)

[...]

PR #3819, opened 2026-05-13
Save screenshots to Maven build output directory

basilevs:
Partial fix for eclipse-platform/eclipse.platform.swt#3303.

SWT does not use Platform, so existing heuristics do not work in GitHub Actions.

Once this more sensible screenshot location is set, SWT GHA workflows can be updated to archive screenshots to artifacts. Before the change they have been thrown away and could not be inspected.

### DIFF
--- a/eclipse-platform-parent/pom.xml
+++ b/eclipse-platform-parent/pom.xml
@@ -131,6 +131,7 @@
     <!-- system specific JVM args; if needed provided by system properties to the build command -->
     <surefire.systemProperties></surefire.systemProperties>
     <surefire.timeout>1200</surefire.timeout>
+    <surefire.screenshot_directory>${project.build.directory}/screenshots</surefire.screenshot_directory>
     <java.version>17</java.version>
     <printApiMessages>false</printApiMessages>
     <failOnJavadocErrors>false</failOnJavadocErrors>
@@ -343,6 +344,11 @@
           <groupId>org.apache.maven.plugins</groupId>
           <artifactId>maven-surefire-plugin</artifactId>
           <version>${surefire.version}</version>
+          <configuration>
+            <systemPropertyVariables>
+                <eclipse.test.screenshot_directory>${surefire.screenshot_directory}</eclipse.test.screenshot_directory>
+            </systemPropertyVariables>
+          </configuration>
         </plugin>
          <plugin>
           <groupId>org.apache.maven.plugins</groupId>
@@ -434,6 +440,9 @@
             <enableAssertions>true</enableAssertions>
             <argLine>${surefire.testArgLine} ${surefire.platformSystemProperties} ${surefire.systemProperties} ${surefire.moduleProperties}</argLine>
             <quiet>true</quiet>
+            <systemProperties>
+                <eclipse.test.screenshot_directory>${surefire.screenshot_directory}</eclipse.test.screenshot_directory>
+            </systemProperties>
           </configuration>
         </plugin>
         <plugin>
--- a/eclipse.platform.releng/bundles/org.eclipse.test/src/org/eclipse/test/Screenshots.java
+++ b/eclipse.platform.releng/bundles/org.eclipse.test/src/org/eclipse/test/Screenshots.java
@@ -103,27 +103,34 @@ public final class Screenshots {
         return filename;
     }
 
-    /**
-     * @return unspecified
-     * @noreference This method is not intended to be referenced by clients.
-     */
-    public static File getResultsDirectory() {
-        File resultsDir = getJunitReportOutput(); // ends up in testresults/linux.gtk.x86_6.0/<class>.<test>.png
+	/**
+	 * @return unspecified
+	 * @noreference This method is not intended to be referenced by clients.
+	 */
+	public static File getResultsDirectory() {
+		File resultsDir = discoverResultsDirectory();
+		resultsDir.mkdirs();
+		return resultsDir;
+	}
 
-        if (resultsDir == null) {
-            File eclipseDir = new File("").getAbsoluteFile();
-            if (isRunByGerritHudsonJob()) {
-                resultsDir = new File(eclipseDir, "/../").getAbsoluteFile(); // ends up in the workspace root
-            } else {
-                resultsDir = new File(System.getProperty("java.io.tmpdir"));
-            }
-        }
+	private static File discoverResultsDirectory() {
+		File resultsDir = getJunitReportOutput(); // ends up in testresults/linux.gtk.x86_6.0/<class>.<test>.png
+		if (resultsDir != null) {
+			return resultsDir;
+		}
+		// Set from "surefire.screenshot_directory" Maven property by
+		// eclipse-platform-parent/pom.xml
+		String propertyPath = System.getProperty("eclipse.test.screenshot_directory");
+		if (propertyPath != null) {
+			return new File(propertyPath);
+		}
 
-        resultsDir.mkdirs();
-        return resultsDir;
-    }
+		return new File(System.getProperty("java.io.tmpdir"));
+	}
 
-    private static File getJunitReportOutput() {
+	private static File getJunitReportOutput() {
+		// Current implementation only interfaces with Ant test launcher.
+		// Maven tests do not use "-junitReportOutput" argument.
 		boolean platformAvailable = false;
 		try {
 			Class.forName("org.eclipse.core.runtime.Platform", false, Screenshots.class.getClassLoader());
@@ -140,14 +147,6 @@ public final class Screenshots {
 			}
 		}
 		return null;
-    }
-
-    /**
-     * @return unspecified
-     * @noreference This method is not intended to be referenced by clients.
-     */
-    public static boolean isRunByGerritHudsonJob() {
-        return System.getProperty("user.dir").matches(".*/(?:eclipse|rt\\.equinox)\\.[^/]+-Gerrit/.*");
-    }
+	}
 
 }
--- a/.github/workflows/mavenBuild.yml
+++ b/.github/workflows/mavenBuild.yml
@@ -94,4 +94,4 @@ jobs:
         if-no-files-found: warn
         path: |
           ${{ github.workspace }}/**/target/surefire-reports/*.xml
-          ${{ github.workspace }}/**/results/**/*.png
+          ${{ github.workspace }}/**/target/screenshots/*.png


basilevs:
@merks @HeikoKlare could you suggest a better heuristic or another way to work with CI screenshots in SWT?

merks:
I have no knowledge in this area.  Maybe @HannesWell has a good idea.

HeikoKlare:
It sounds reasonable to have a properly defined place for storing the screenshots so that they can be archived during GHA runs. I remember that I did some tweaks when analyzing an issue only occurring in GHA runs for which I required screenshots. So having a general solution for this would be great. Thank you for putting effort into this.

I wonder if we couldn't just define a proper value for the existing system property `junitReportOutput` in the Maven build. Doing that in the parent might only require it to be properly defined/parameterized once.
In case we only want a specific folder to be used in GHA runs, as we align the GHA workflow configuration with specific code for defining the output folder, we may also simply guard this behavior with the GHA context, like we do in some utilities such as: https://github.com/eclipse-platform/eclipse.platform.swt/blob/f33b79ee65f5cedfa8f7ac4e1bf93994a904a445/tests/org.eclipse.swt.tests/JUnit%20Tests/org/eclipse/swt/tests/junit/SwtTestUtil.java#L121-L123

Regarind the proposed change: Unless the Eclipse platform is present and the system property for the output folder is specified, the results will now always be written into the surefire target repository (if existing). Is that what we want to happen? For example, when executing the tests locally in the IDE, do we want to have the screenshots placed in a Maven-build-specific folder? In case I have execute a local Maven build at some earlier point in time, that folder will be present and used.

Maybe @akurtakov can also help here, as he guarded the screenshot output folder extraction with the Platform being present recently:
- https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/pull/3577

basilevs:
> existing system property `junitReportOutput`

Is not it an application argument and not a system property?

HeikoKlare:
> Is not it an application argument and not a system property?

You're right. Then we maybe want to check for a GHA environment and then define a dedicated folder that is also used for artifact upload?

basilevs:
@HeikoKlare 
> You're right. Then we maybe want to check for a GHA environment and then define a dedicated folder that is also used for artifact upload?

The decision should be based on whether Maven is in use, not on which environment it is running on.
I've added a dedicated property and a suitable default via parent pom in the latest revision. Is it better?

As some projects reuse aggregator's workflow, and most use it's parent POM, the maintenance effort per-project should be minimal.

eclipse-platform-parent/pom.xml
```diff
@@ -135,6 +135,7 @@
     <printApiMessages>false</printApiMessages>
     <failOnJavadocErrors>false</failOnJavadocErrors>
     <compileLogDir>${project.build.directory}/compilelogs</compileLogDir>
+    <eclipse.screenshot_directory>${project.build.directory}/screenshots</eclipse.screenshot_directory>
```
HannesWell: Wouldn't the property be better named `surefire.screenshot_directory` (and be placed next to the other surefire related properties)?
```suggestion
    <surefire.screenshot_directory>${project.build.directory}/screenshots</surefire.screenshot_directory>
```

HannesWell (commented):
Generally I think this is fine.
Do you know if the screenshots are also saved for I-build tests and if some according adaption is needed then? I'm not aware of it, but might have missed it.

eclipse-platform-parent/pom.xml
basilevs: The JVM system property is naturally usable outside of surefire. For example, it could be configured via launch configuration. This makes naming it surefire specific confusing.

Maven property was named to exactly match the JVM system property to help with identification and search.

basilevs:
> Do you know if the screenshots are also saved for I-build tests

I[ did fix screenshots for platform.ui a few years back](https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/pull/1485). Those relied on `Platform` to be present.  I'm not sure if that fix applies to I-build.

eclipse-platform-parent/pom.xml
basilevs: Applied anyway.

HeikoKlare:
In general, the approach also looks fine for me now.
But I have to admit that I am a bit lost with the different configurations and how they are described in the code.

You mention this older PR that states to ensure that test screenshots are provided in the GH workflow artifacts:
- https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/pull/1485

That one did not affect I-Builds but just GH workflows and introduced archiving of screenshots placed here:
> ${{ github.workspace }}/**/results/**/*.png

I ran local Maven builds and explicitly made some tests fail. The produced screenshots are placed in the temp folder. So I wonder if this configuration is still valid?

Also note that there is further documentation (and outdated code relying on Gerrit) in the `Screenshots.getResultsDirectory()` method. I also think we should clean up the code in `Screenshots.getResultsDirectory()` and `Screenshots.getJunitReportOutput()` to clarify our current understanding of it. The Gerrit-related code can probably be removed. the `getJunitReportOutput()` is specific to the system argument that is used in Ant-based I-Builds and there will be the new configuration option based on a system property currently used by Maven (which is currently placed in `getJunitReportOutput()  which in my opinion is misleading). When they are all not present, we will fall back to using the temp folder.

basilevs:
> I ran local Maven builds and explicitly made some tests fail. The produced screenshots are placed in the temp folder. So I wonder if this configuration is still valid?

I can't find how Ant tests (which store data in `results/*`) are launched in CI, so I can't answer that. Maven tests do not use the relevant command line argument.

> Also note that there is further documentation (and outdated code relying on Gerrit) in the `Screenshots.getResultsDirectory()` method. I also think we should clean up the code in `Screenshots.getResultsDirectory()` and `Screenshots.getJunitReportOutput()` to clarify our current understanding of it. The Gerrit-related code can probably be removed.

Do you suggest to cleanup Gerrit code as part of this unrelated feature? While I can certainly do so, I have no means to reliably test it (without committer level access to all involved CI infra) and would like to avoid mixing concerns.

> the getJunitReportOutput() is specific to the system argument that is used in Ant-based I-Builds and there will be the new configuration option based on a system property currently used by Maven (which is currently placed in `getJunitReportOutput() which in my opinion is misleading). 

I see no indication that `org.eclipse.test.Screenshots.getJunitReportOutput()` is or should be limited to Ant. Indeed, all tests run using JUnit in both Ant and Maven. IMO, it using Ant specific argument is an implementation detail. I can move the new code out to `getResultsDirectory()`, but I do not believe this would improve clarity in any way.

HeikoKlare:
> I can't find how Ant tests (which store data in results/*) are launched in CI, so I can't answer that. Maven tests do not use the relevant command line argument.

Why do you assume Ant tests to store data in "result/*"?  From what I see, the configuration is only used in a GH workflow, so used by GHA runs. The GHA runs use Maven to execute tests and the Maven execution does not seem to place the screenshots at any reasonable place currently. The Ant tests set the `junitReportOutput` property in the according Ant scripts:
https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/blob/9c51d36c22163bf7c0c9b29ccada26369fdfde15/sites/eclipse-junit-tests/src/main/scripts/library.xml#L266-L267

> Do you suggest to cleanup Gerrit code as part of this unrelated feature? While I can certainly do so, I have no means to reliably test it (without committer level access to all involved CI infra) and would like to avoid mixing concerns.

I just suggest to follow boy scout rule and clean up the current implementation of how the path for storing the screenshots is defined instead of adding further cases at a "random" position. In addition:

> I see no indication that `org.eclipse.test.Screenshots.getJunitReportOutput()` is or should be limited to Ant. Indeed, all tests run using JUnit in both Ant and Maven. IMO, it using Ant specific argument is an implementation detail. I can move the new code out to `getResultsDirectory()`, but I do not believe this would improve clarity in any way.

The method is about processing the `junitReportOutput` argument (which is why it is named like that). This argument is currently only used when executing the tests via Ant. I just proposed to make the understanding that we gained in this discussion explicit in the code to avoid that we meet again in few months and have to extract all the semantics out of some grown code again.

That's why I propose to streamline the implementation by removing what is obsolete and cleaning up what needs to remaind or is added:
- Make the `getResultsDirectory()` implementation operate based on precedence:
  - If present, use the `junitReportOutput` argument (like done with the existing `getJunitReportOutput() already, maybe add a note that this is currently used by Ant-based test execution)
  - Otherwise if present, use the added sytem property `eclipse.screenshot_directory` for the screenshots (maybe add a note that this is currently used by Maven-based test execution)
  - Otherwise fall back to using temp dir
- Remove the Gerrit-related implementation, as it's unused
- Remove the obsolete "results/*" entry in the GH workflow

If you want, I can also provide a preparatory PR removing the obsolete code/configuration so that you can then add your new configuration on top.

basilevs:
> Why do you assume Ant tests to store data in "result/*"?

https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/blob/9c51d36c22163bf7c0c9b29ccada26369fdfde15/sites/eclipse-junit-tests/src/main/scripts/library.xml#L158-L161

> The method is about processing the junitReportOutput argument (which is why it is named like that). This argument is currently only used when executing the tests via Ant. 

The argument is named this way, because it manages JUnit output. Arguably, method name matches the argument, because they serve the same purpose, not because method implementation depends on argument. It does not matter though, I'll move stuff around to unblock.

> That's why I propose to streamline the implementation by removing what is obsolete and cleaning up what needs to remaind or is added:

I will apply these changes as requested, but am unable to test that.

> If you want, I can also provide a preparatory PR removing the obsolete code/configuration so that you can then add your new configuration on top.

I'll do the change, but it would be nice to somehow run all related CI's on it. I'm not sure if it even possible (given all sub-projects have their own configurations, some reusing `buildMaven.yml` and some not)

basilevs:
@HeikoKlare I've applied all suggestions.

>     * Remove the obsolete "results/*" entry in the GH workflow

When I was doing the old PR, I assumed that Maven somehow started Ant tests for platform.ui project. This does not seem to be the case anymore, but I can't verify that. A few of recent failed builds in platform.ui do not produce screenshots at all (no message about AWT screenshot in console). Do you happen to know what is going on there?

BTW:
Following aggregator submodules use `buildMaven.yml`:

- eclipse.pde
- eclipse.platform.ui
- eclipse.platform
- equinox.p2

Surprisingly not a lot of them. Still, we can't be sure, that projects out of aggregator would not break.

eclipse.platform.releng/bundles/org.eclipse.test/src/org/eclipse/test/Screenshots.java
```diff
@@ -140,14 +147,15 @@ private static File getJunitReportOutput() {
...
-     */
-    public static boolean isRunByGerritHudsonJob() {
-        return System.getProperty("user.dir").matches(".*/(?:eclipse|rt\\.equinox)\\.[^/]+-Gerrit/.*");
-    }
+	/**
+	 * @return unspecified
+	 * @noreference This method is not intended to be referenced by clients.
+	 */
+	@Deprecated
+	public static boolean isRunByGerritHudsonJob() {
```
HeikoKlare: This plugin is not public API, so we don't need to deprecate this but can just remove it.

HeikoKlare (commented):
> I've applied all suggestions.

Thank you! It looks fine to me and I am sure this will help us in the future to better understand the code. @HannesWell what do you think about the changes? 

> When I was doing the old PR, I assumed that Maven somehow started Ant tests for platform.ui project. This does not seem to be the case anymore, but I can't verify that. A few of recent failed builds in platform.ui do not produce screenshots at all (no message about AWT screenshot in console). Do you happen to know what is going on there?

I have researched in history for a while now, trying to understand why/how this could have ever worked (using the "target/results" directory), but I have no clue yet. To the best of my knowledge, the GH workflows always used pure Maven (with surefire for test execution) and Ant was only used for I-Build tests. Maybe there was a different configuration at some point in time (in particular the `Screenshots` class was heavily modified and moved around) that made this work. Are you sure that the artifacts upload has ever worked?

I also wondered where screenshots are actually taken. I assumed that the `ScreenshotOnFailure`/`Screenshots.onFailure()` rules is used by many UI tests in Platform UI, but that's not the case. Some usages have silently been removed during JUnit 5 migration, such as the one in `PartRenderingEngineTests` via https://github.com/eclipse-platform/eclipse.platform.ui/pull/3636.
@vogella do you know if there were more usages of this JUnit 4 rule that have been removed during JUnit 5 migration?

> BTW: Following aggregator subprojects projects use `buildMaven.yml`:
> 
> * eclipse.pde
> * eclipse.platform.ui
> * eclipse.platform
> * equinox.p2
> 
> Surprisingly not a lot of them. Still, we can't be sure, that projects out of aggregator would not break.

Projects out of aggregator are not supposed to use those workflows. But anyway, where do you see a chance that anything could break? The change in the workflow descriptions just affects an included folder in artifacts upload, so the worst I can expect is that some PNGs are not uploaded as expected?

eclipse.platform.releng/bundles/org.eclipse.test/src/org/eclipse/test/Screenshots.java
basilevs: For non-public API, it is used a lot:

Screenshot 2026-05-15 at 18 14 18

HeikoKlare: Sorry, my comment was a bit imprecise: the plugin is not part of the SDK but only used internally for tests inside the SDK. And within the SDK, there are no further consumers of this method:
image

basilevs:
> Projects out of aggregator are not supposed to use those workflows. But anyway, where do you see a chance that anything could break? The change in the workflow descriptions just affects an included folder in artifacts upload, so the worst I can expect is that some PNGs are not uploaded as expected?

Well, given this is a second time I'm fixing these screenshots archival I would not wish for anyone to do this yet again.

HeikoKlare:
> Well, given this is a second time I'm fixing these screenshots archival I would not wish for anyone to do this yet again.

I have to admit that I still don't get which risk we need to consider here. The archival is already broken (and probably never worked), so what do you think might become worse with this PR?

basilevs:
> > Well, given this is a second time I'm fixing these screenshots archival I would not wish for anyone to do this yet again.
> 
> I have to admit that I still don't get which risk we need to consider here. The archival is already broken (and probably never worked), so what do you think might become worse with this PR?

I have no visibility over usage of involved workflow and library. I'll defer to you for the decision.

eclipse.platform.releng/bundles/org.eclipse.test/src/org/eclipse/test/Screenshots.java
```diff
@@ -103,27 +103,34 @@ public static String takeScreenshot(Class<?> testClass, String name) {
...
-            }
-        }
+	private static File discoverResultsDirectory() {
+		File resultsDir = getJunitReportOutput(); // ends up in testresults/linux.gtk.x86_6.0/<class>.<test>.png
+		if (resultsDir != null) {
+			return resultsDir;
+		}
+		// Set from "surefire.screenshot_directory" Maven property by
+		// eclipse-platform-parent/pom.xml
+		String propertyPath = System.getProperty("eclipse.screenshot_directory");
```
HannesWell: In this case I think it's suitable to be a bit more specific and name the property `eclipse.test.screenshot_directory`
```suggestion
		String propertyPath = System.getProperty("eclipse.test.screenshot_directory");
```
And following up on the discussion from: https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/pull/3819#discussion_r3242781255:

I understand your point, still it can also make sense to name the same thing differently depending on the exact context. Furthermore another name could also allow to overwrite the general value with a more specific one in a more specific case, if necessary. Currently not necessary (and no clue if it ever will).

HannesWell (commented):
In general this looks good for me and I just have one nit for the name of the property (sorry for the second round).

> I'll do the change, but it would be nice to somehow run all related CI's on it. I'm not sure if it even possible (given all sub-projects have their own configurations, some reusing `buildMaven.yml` and some not)

While it's not ideal to have test-failures I wouldn't be too concerned and would accept some uncertainty if the effort to verify a change is too high. The testing framework is just used internally by the Eclipse TLP so we can fix or change anything at any time as we need it.

The verification builds running in GH workflows or Jenkins all use Maven/Tycho surefire. Only the I-build tests are launched using ANT, shell-scripts and the built eclipse version. It all starts at
https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/blob/4f490a54d58ff3eb925c4026ee6df7e6c365fcea/JenkinsJobs/AutomatedTests/integrationTests.jenkinsfile#L93

Currently it's sadly quite convoluted by I'll work on it to become simpler in the next development cycle. I've summarized the plan at (if you are interested):
- https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/issues/2979

This hopefully also simplifies local execution and testing.

eclipse.platform.releng/bundles/org.eclipse.test/src/org/eclipse/test/Screenshots.java
basilevs: Done

HannesWell (approved):
Thanks!

HeikoKlare (approved):
Thank you for the final update and patiently working on a good overall solution.

HeikoKlare:
Changes only affect tests, thus merging despite freeze period.

HeikoKlare:
Just verified that this works as expected: made a test that uses the screenshot functionality on failure explicitly fail and checked the artifacts of the GitHub workflow

This is the according PR:
- https://github.com/HeikoKlare/eclipse.platform.ui/pull/2

And this screenshot is present in the artifacts upoad of the MacOS build:
image

PR #3821, opened 2026-05-14
Improve downloads website styling with dark mode and wider layout

### LINKED ISSUE #3812: Update download page
The current download page looks IMHO dated. I suggest to modernize it, and will make suggestions for that. 

Current L&F

Overview (Landing page):

Image

Details for a build:

Image

vogella:
Extends `sites/eclipse/page.css` with a CSS-variable palette (Eclipse-blue accent), automatic dark mode via `prefers-color-scheme`, Inter and JetBrains Mono typography, card-style tables with larger headers, card-shaped details/accordions and a gradient-stripe popup dialog.

Replaces the section info marker emoji with a FontAwesome `fa-solid fa-circle-info` icon for a consistent, OS-independent look.

Widens the content container to 1600px and adjusts the main/TOC column split to 80/20 so wide tables have more room.

HTML markup, JSON schema and the build pipeline are untouched.

Fixes https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/issues/3812

### DIFF
--- a/sites/eclipse/page.css
+++ b/sites/eclipse/page.css
@@ -1,35 +1,217 @@
 @import "https://eclipse.dev/eclipse.org-common/themes/solstice/public/stylesheets/astro.min.css";
-@import "https://fonts.googleapis.com/css2?family=Libre+Franklin:ital,wght@0,100;0,200;0,300;0,400;0,500;0,600;0,700;0,800;0,900;1,100;1,200;1,300;1,400;1,500;1,600;1,700;1,800;1,900&family=Roboto:ital,wght@0,100;0,300;0,400;0,500;0,700;0,900;1,100;1,300;1,400;1,500;1,700;1,900&display=swap";
 
 :root {
 	interpolate-size: allow-keywords;
-	--marker-error: red;
-	--marker-warning: orange;
-	--marker-info: dodgerblue;
-	--marker-task: dodgerblue;
+
+	/* Brand / accent palette */
+	--color-accent: #2c5aa0;
+	--color-accent-hover: #1f4478;
+	--color-accent-soft: #eaf1fb;
+
+	/* Neutrals */
+	--color-text: #1f2933;
+	--color-text-muted: #52606d;
+	--color-surface: #ffffff;
+	--color-surface-alt: #f7f9fc;
+	--color-surface-hover: #eef2f7;
+	--color-border: #e1e6ec;
+	--color-border-strong: #c4ccd6;
+
+	/* Status */
+	--marker-error: #d64545;
+	--marker-warning: #e58e26;
+	--marker-info: #2c7be5;
+	--marker-task: #2c7be5;
+	--color-success: #2f855a;
+	--color-success-soft: #e6f4ec;
+	--color-error-soft: #fdecec;
+	--color-warning-soft: #fdf3e2;
+
+	/* Shape & motion */
+	--radius-sm: 4px;
+	--radius-md: 6px;
+	--radius-lg: 10px;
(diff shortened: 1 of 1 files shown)


HannesWell:
> Extends `sites/eclipse/page.css` with a CSS-variable palette (Eclipse-blue accent), automatic dark mode via `prefers-color-scheme`, Inter and JetBrains Mono typography, card-style tables with larger headers, card-shaped details/accordions and a gradient-stripe popup dialog.

Generally I don't think its good to more than double, almost triple the size of the CSS file. As mentioned in https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/issues/3812#issuecomment-4414527501 one goal is simplicity or at least to make it not too complex.
Would it be possible to leave out few new details to make this significantly simpler?

> Inter and JetBrains Mono typography

I have to say that I don't see a noticeable difference between the fonts. But generally it seems to be smaller, especially the titles, which now seem to be even smaller than the following paragraph text.

> 
> Replaces the section info marker emoji with a FontAwesome `fa-solid fa-circle-info` icon for a consistent, OS-independent look.

Please move this into a separate PR. In https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/issues/3812#issuecomment-4414527501 I described the drawbacks of using FontAwesome (of course it has advantages).
But I'm still not sure if this is the overall better solution.
In fact these icons where used in a previous version, which the git history should show. I changed that for the current solution, but that also isn't ideal in all aspects.

> 
> Widens the content container to 1600px and adjusts the main/TOC column split to 80/20 so wide tables have more room.

This should also discussed separately. With that applied the TOC looks squashed on long lines like the one for checksums. To solve that a mechanism could be implemented to show a shorter title in the TOC than at the actual section (in that case e.g. just `Checksums and Signatures`).

vogella:
> Would it be possible to leave out few new details to make this significantly simpler?

Done

Icon and Font changes also removed from the PR.

> Widens the content container to 1600px

Also removed from the chnage

vogella:
@HannesWell is your feedback addressed?

HannesWell:
> is your feedback addressed?

In my opinion this is still a lot of extra complexity and I would be in favor if it could be done simpler.
It doesn't have to be a net zero change, but in general I think that simplicity is preferred over visual beauty. This page targets rather technical people and is probably not visited by a wider audience of users (like for example https://eclipseide.org/ is).
I'm not even sure how one can reach https://download.eclipse.org/eclipse/ if one doesn't know it?
In that regard I think it's similar to pages like https://openjdk.org/ or https://jdk.java.net/26 and these sites are, while technically interesting and useful, visually not sophisticated.

That being said, if we still want to have this, I have to review this in detail and we should further fine tune this.
I assume you are mainly interested in the dark mode?
I did a quick check and for my taste the table headers are a bit too bright and my eyes try to find a difference with the background, which makes it unpleasent to look at. I know that's probably pretty subjective, but we are discussing visuals here. :)
I also checked the dark mode and headlines are impossible to read and the bright header doesn't match.

Have you tried out/tested the changes locally?
https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/tree/53c0ae09eb9a47860540f14786f6876c6b36819e/sites/eclipse#local-testing
We should not figure this out in production.

Maybe it's simpler to discuss and review if the changes for the dark mode and common rework of colors and styles are separated?

vogella:
Thanks for the feedback @HannesWell. Assuming that this page is rarely used and it sounds like a lot of work for you to review this, I close the PR. Also I looks to me that the styling in other areas need more attention, for example the test results look a bit broken these days (I do not plan to work on this).

image

Issue #3328, opened 2025-09-15
Required retention policy for Y-build artifacts


HannesWell:
In the context of unifying the clean-up jobs in the RelEng Jenkins instance (https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/pull/3322) and generally refining the preparation of a new release cycle I stumbled upon the question for how long Y-builds, i.e. the drops and the corresponding p2 repos are actually required and should be retained.

The release notes explain that the Y-builds should be retained if a new release happens shortly after the Eclipse release (usually for odd versions):
https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/blob/c9fd1cc9cfa2f75098372f1be1e5d67ce4591eb2/RELEASE.md?plain=1#L122-L123

Which matches what has been explained in https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/issues/3273#issuecomment-3244735110.
And given that P-builds are now handled by JDT directly in the JDT JIPP and download space, I think we can disregard P-builds from further discussions.

This leads me to the following more specific questions:
- For how long should are the Y-build drops at the download page are needed and should be retained?
For I-builds only all the drops younger than five days are retained and for older I-builds of the current release, only one per week (roughly summarized). Would something similar also fit for Y-build drops? At the moment I cannot tell what actually cleans-up the Y-build drops (but it could be aligned with https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/pull/3322).
  - https://download.eclipse.org/eclipse/downloads/YIndex.php
  - https://download.eclipse.org/eclipse/downloads/drops4/Y20250909-1000/

- How long should the Y-build p2-repos be retained?
  - According to the doc linked above and my understand of the process, generally some time after the Java-release (i.e. the composite)
  - But are all p2 repos of all builds required until after the release or are you just interested in the latest build's repos anyways (e.g. the last five or so)? Especially as you keep the relevant bundles in the jdt/P-builds repo any ways.
  - And for how long are they required after the release?
  In the past the Y-build repos have been deleted a year later.
- Are the Y-build drops for Equinox actually interesting for anybody or can they be removed?
  - https://download.eclipse.org/equinox/drops/Y20250911-1000/

And besides the general questions above, I wonder specifically if the old `4.XX-Y-builds`, `4.XX-P-builds` or `4.XX-P-builds-old` directories/composite repositories are still required in the Eclipse downloads area or can be deleted?
[...]

PR #3827, opened 2026-05-17
[Y-Build] Align retention policy for Y-build p2-repositories with drops

HannesWell:
Similar to the Y-build drops, retain only the p2-repositories of the latest nine Y-builds.

Completes https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/issues/3328

@stephan-herrmann, @jarthana, @mpalat from the linked issue, I assume this is fine for you, is that correct?


mpalat:
> Similar to the Y-build drops, retain only the p2-repositories of the latest nine Y-builds.
> @stephan-herrmann, @jarthana, @mpalat from the linked issue, I assume this is fine for you, is that correct?

Yes, it's find. Thanks @HannesWell!

PR #3829, opened 2026-05-20
Update junit 6.0.3 to junit 6.1.0

merks:


iloveeclipse:
Do we really want it for 4.40? Junit updates are often unpredictable :-)

merks:
Only adding junit 6 has been a problem for all the updates that I recall over the past years, so yes.  Also it's really hard to keep Orbit and the Platform aligned during these final steps, so another reason to push this forward.  In the worst case, we could back it out...

iloveeclipse:
> In the worst case, we could back it out...

OK, we have not past RC2a yet :-)

PR #3832, opened 2026-05-21
[RelEng] Add job to send RelEng announcement emails

HannesWell:
This adds a new RelEng job to send the announcements for the RC weeks, the opening of the master branch for a new development cycle and to send the sign-off requests as well as creating the corresponding issue for it in this repository.
Having this automated avoids the need to collect the dates and data manually.

@MohananRahul FYI, this should make it very simple to send these announcements. For the RC 2 reminder and the RC2 sign off next week this is hopefully already usable.
And do you have an idea for a better name (creating the sign-off issue is only an announcement in the wider sense) or do you think it's suitable?

~~This is currently a draft since I have to test this first and have a few pending TODOs.~~


MohananRahul:
> @MohananRahul FYI, this should make it very simple to send these announcements. For the RC 2 reminder and the RC2 sign off next week this is hopefully already usable. And do you have an idea for a better name (creating the sign-off issue is only an announcement in the wider sense) or do you think it's suitable?

Yes ,The current name seems reasonable as the sign-off issue creation.

JenkinsJobs/Releng/sendAnnouncement.jenkinsfile
```diff
@@ -0,0 +1,150 @@
...
+							- ${rcDate.minusDays(2).format(dateFormat)} 23:00 (UTC)
+							  API/feature freeze and RC${rcNumber} build
+							
+							- ${rcDate.format(dateFormat)}
+							  Submit RC${rcNumber} to SimRel
+							
+							Please note the following excerpt from the freeze plan:
+							All fixes submitted must have a project lead vote on the bug report, and the fix must be reviewed by an additional committer (any committer other than the one who made the fix).
+							A positive review from a project lead means implicit approval and no vote is needed on the bug report.
+							Ongoing changes to documentation, tests or examples do not require approval.
```
HannesWell, line 80: This text should be updated to the development workflow at GH, to me it seems like still from the time when Bugzilla and Gerrit was used.
But I'll leave that for later.

HannesWell (commented):
This is now ready for submission.

To be more international, I've also updated all times to be in UTC.
I translated the RC deadline to 09:00 UTC and the submission to 12:00 UTC.
@MohananRahul I hope thats still fine for you.
The latter It should be about 17:30 in India and 13:00 or 14:00 in central Europe (depending on Winter or Summer). And the sign-off deadline is three hours earlier.
I think that's all suitable. And in general I have the impression that the sign-off is not taken too seriously and it's actually only relevant when the submission is done (because then it's fixed).

But if necessary, we can still fine-tune this later. For a first step to automate this, I think this is now sufficient.

PR #3839, opened 2026-05-25
[RelEng] Create release tracking issue when sending RC2 sign-off request

HannesWell:
This automates the creation of the issue that tracks the final release steps and makes the manual execution and setup of the `GAReleasePrep.sh` obsolete.

Generally the release tracking issue could also be created in `publishPromotedBuild.jenkinsfile` when publishing RC2 , but since the `sendAnnouncement.jenkinsfile` job already automates other administrative tasks, I thought it would fit there better.
Furthermore it creates the tracking issue two days earlier, which could also be suitable.

This currently needs a test-run but besides that should be ready.


HannesWell:
This is now ready for submission and should be used for the sign-off request send tomorrow.

@MohananRahul do you want to do it or should I do it tomorrow (European) morning?

To send the request the [Send Announcement](https://ci.eclipse.org/releng/job/Releng/job/sendAnnouncement/) Jenkins job should be used, with the parameters
- `type`: `SIGNOFF_RC`
- `rcNumber`: `2`
- `buildId`: `I20260527-1800` (expected)

The `deadline` should be left blank since it's a default RC.
That job also creates the sign-off issue, so it doesn't have to be created manually.

PR #3679, opened 2026-02-26
Adapt Release and RelEng documentation to reworked release automation

HannesWell:
Update the main preparation PR to include all remaining steps and to include further instructions. The main preparation PR in this repository then also acts as replacement for the issue tracking the preparation.

PR #3863, opened 2026-05-31
[RelEng] Set publication time at release day to 10:00 UTC

HannesWell:
As proposed in
- https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/pull/3679#discussion_r3299369137

this moves the time of the final publication of the release at the release day to 12:00 UTC (which translate to 07AM EST, 14:00 CEST or 17:30 IST).
Doing it earlier relaxes the situation for SimRel which wants to release after the Eclipse project.
Since the only task is starting one job, not much preparation time is needed at that day.

Since all people involved in the RelEng work are actually in Europe or India, we could even move this to 10:00 UTC, which would be 12:00 CEST or 15:30 IST.

@merks is that time fine fine for SimRel.
CC @MohananRahul

### DIFF
--- a/JenkinsJobs/Releng/sendAnnouncement.jenkinsfile
+++ b/JenkinsJobs/Releng/sendAnnouncement.jenkinsfile
@@ -93,7 +93,7 @@ pipeline {
 								  - [ ] Submit the pending PRs in this repository to update build scripts to ${releaseVersion} GA on master and the ${releaseVersion}-maintenance branch
 								
 								### Release day
-								at ${eclipseReleaseDates.GA.format(dateFormat)} 15:00 UTC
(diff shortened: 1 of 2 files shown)


merks:
Yes that’s fine and good. I do many of the publishing activities early on the release day so I can test everything properly. Only the main download pages are made available in at 4:00pm European central time.

HannesWell:
> Since all people involeved in the RelEng work are actually in Europe or India, we could even move this to 10:00 UTC, which would be 12:00 CEST or 15:30 IST.

Then lets move it to even 10:00 UTC. The earlier the better.

MohananRahul:
> Then lets move it to even 10:00 UTC. The earlier the better.

Sure, that's fine

PR #3868, opened 2026-06-01
Install newer EGit to avoid SDK using a newer/different bouncy castle

merks:


merks:
The build doesn't test this so emerging immediately.

PR #3875, opened 2026-06-02
Remove unused JDBC/Derby dead code from performance test bundle

vogella:
The org.eclipse.test.internal.performance.db package still carried classes from the old Derby/Cloudscape-backed results database that is no longer used. SQL (the only java.sql.* consumer), Report, and SummaryEntry have no references anywhere in the bundle, so they are removed, along with two stale comments in StatisticsSession that referenced the gone DB code. Variations and TimeSeries live in the same package but are still in use and are kept untouched.


akurtakov:
I have tried hard to remove useless content from this bundle in the past but missed these.

PR #3881, opened 2026-06-04
Replace abandoned Wandalen/wretry.action in version increment workflow

HannesWell:
The `com/Wandalen/wretry.action` seems not to be updated anymore and is using the deprecated Node.js 20:
- https://github.com/Wandalen/wretry.action/issues/193

This replaces it by a simple loop in bash.


HannesWell:
Jenkins build failure is unrelated.

PR #3890, opened 2026-06-10
Remove unused com.ibm.icu icu4j dependency from SDK prereqs target

vogella:
Nothing in the Eclipse SDK source consumes `com.ibm.icu` anymore. No production bundle declares a `Require-Bundle` or `Import-Package` on it, and no Java source imports it; the only remaining references in the tree are test fixtures and old test data. Following the recent removal of the icu4j `extraRequirement` from the ISV doc builds, the icu4j entry in the `eclipse-sdk-prereqs` target is now dead weight, so this drops it.

This shrinks the target platform and removes a long-standing third-party dependency that the SDK no longer needs.

### DIFF
--- a/eclipse.platform.releng.prereqs.sdk/eclipse-sdk-prereqs.target
+++ b/eclipse.platform.releng.prereqs.sdk/eclipse-sdk-prereqs.target
@@ -113,12 +113,6 @@
 				  <type>jar</type>
 			  </dependency>
 			  <dependency>
-				  <groupId>com.ibm.icu</groupId>
-				  <artifactId>icu4j</artifactId>
-				  <version>78.3</version>
-				  <type>jar</type>
-			  </dependency>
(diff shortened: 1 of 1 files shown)


vogella:
cc @merks @akurtakov @HannesWell, not sure if the simrel will be affected by such a change.

akurtakov:
Now is the right time to merge it so there is enough time for projects to fix their target platforms if needed.

HannesWell:
It's really great to have this large dependency, it had 15MB in size, finally removed. Thanks for that effort.

E.g the Platform products now are only about 85MB in size. In the 4.38 release they had ~150MB and removing sources already shrinked them by 50MB).
Considering that we have two products for eight platforms and two p2 repos in each build drop, every reduction is multiplied by the factor of 18. So with this we save about 270-280MB of storage size per build.

PR #3891, opened 2026-06-10
Move I/Y-Build and RelEng jobs to build with Java-25

HannesWell:
SimRel starts to accept Java-25 bundles, and although the bundles of the Eclipse top level project still must still require only on Java-21, the build and RelEng infrastructure can already be moved to use Java-25 as a first step in that direction.
It also simplifies the build pipeline a bit.

Since the I/Y-build runs with the `bree-libs` profile activated, a JDK matching each bundle's BREE is selected from the toolchain for compilation anyway.

I/Y-build tests remain on their current JDK to test the Java-21 baseline.

@akurtakov this is what we discussed today at the PMC meeting.

### DIFF
--- a/JenkinsJobs/Builds/build.jenkinsfile
+++ b/JenkinsJobs/Builds/build.jenkinsfile
@@ -29,7 +29,7 @@ pipeline {
 		}
 	}
 	tools {
-		jdk 'temurin-jdk21-latest'
+		jdk 'temurin-jdk25-latest'
 		maven 'apache-maven-latest'
 	}
 	environment {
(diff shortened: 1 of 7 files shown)


akurtakov (approved):

HannesWell:
For reference: For the complete shift to Java-25 as baseline the following changes are required too
- https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/pull/4009
- https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/pull/4011
- https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/pull/4015
- https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/pull/4016 (not entirely required, only the javadoc source part).


Additionally the Jenkins pipelines and Oomph setups in all submodules have to be adapted to provide a Java-25 JDK.
The Jenkins pipelines and Github workflows in Equinox and SWT are more complex and special and correspondingly need corresponding adjustments too
- https://github.com/eclipse-equinox/equinox/pull/1350

PR #3895, opened 2026-06-11
Set ECF's httpclient5.win32 forbiddenReference to error

merks:


merks:
Doesn't affect the build.  Merging immediatley.

PR #3896, opened 2026-06-11
Enforces qualifier update on doc bundles due to build with Java-25

HannesWell:
This also enforces a new qualifier update on the doc bundles to clear them from comparator errors in the context of
- https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/pull/3891


HannesWell:
It turned out that the project's BREE is not considered in the selection of the JDK used to generate the JavaDoc, it seems to be simply the default JDK.
Therefore I reverted the set of missing BREEs (which are otherwise not useful) and changed it to a qualifier update of the doc bundles affected by change to build with Java-25.

PR #3901, opened 2026-06-13
[Build] Define build schedule and the build time in UTC timezone

HannesWell:
Using UTC as reference timezone makes the schedule more international and makes it simpler for those not located at the American east-coast to understand the build schedule.

Furthermore it aligns the time-zone of build timestamps with the time-zone used for git timestamps.

This should also fix a rarely occurring failures, when running two integration builds shortly after each other and a change was committed only shortly before the first build.
Since commit timestamps are in UTC, with the previous build timezone the qualifier of the changed artifact had a later timestamp (higher value) than the build timestamp was in that case (between 4 and 5 hours depending on the difference).
If that changed bundle is in included in a feature, together with a bundle that uses the build time as qualifier, the features timestamp was dominated by the bundle that's using the git-timestamp in that case.
In the second build the qualifier of the 'runtime-qualified' bundle of course reflects the second build's time, but the feature was still dominated by the 'git-qualified' bundle, consequently didn't change it's qualifier and was therefore baseline replaced.
But in the baseline version the feature requires the exact timestamp of the previous build for the other, 'runtime-qualified' bundle.
Eventually this lead to a build failure because the version of the 'runtime-qualified' bundle from the previous build isn't found in the current build.

Using UTC, which doesn't have rules for daylight saving, means that the build time will change by on hour for regions that do have such mechanisms, when entering or leaving a daylight saving time period.
But I don't think this is a problem.

### DIFF
--- a/JenkinsJobs/Releng/updateIndex.jenkinsfile
+++ b/JenkinsJobs/Releng/updateIndex.jenkinsfile
@@ -184,7 +184,6 @@ def toUTCdateTime(String dateTime) {
 	def int hour = Integer.parseInt(dateTime.substring(8, 10))
 	def int minute = Integer.parseInt(dateTime.substring(10, 12))
 	return java.time.LocalDateTime.of(year, month, day, hour, minute)
-		.atZone(java.time.ZoneId.of("America/New_York"))
-		.withZoneSameInstant(java.time.ZoneOffset.UTC)
-		.toLocalDateTime().toString() + 'Z';
+		.atZone(java.time.ZoneOffset.UTC)
+		.format(java.time.format.DateTimeFormatter.ISO_INSTANT); // formats in UTC, such as '2011-12-03T10:15:30Z'
 }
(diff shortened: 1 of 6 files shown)


JenkinsJobs/Releng/updateIndex.jenkinsfile
```diff
@@ -184,7 +184,6 @@ def toUTCdateTime(String dateTime) {
 	def int hour = Integer.parseInt(dateTime.substring(8, 10))
 	def int minute = Integer.parseInt(dateTime.substring(10, 12))
 	return java.time.LocalDateTime.of(year, month, day, hour, minute)
-		.atZone(java.time.ZoneId.of("America/New_York"))
-		.withZoneSameInstant(java.time.ZoneOffset.UTC)
-		.toLocalDateTime().toString() + 'Z';
```
HannesWell, line 189: The update index job now assumes that all build-times are in UTC, which is of course false for the existing build drops and will render a build time off by 4/5 hours.
But since in one, two weeks the vast majority of build drops will use the new time-zone the problem will sort itself out over time.
And I don't consider it a relevant problem if the build-time of a past release build is off by a few hours and therefore an exact translation is IMO not worth the overhead.

PR #3902, opened 2026-06-15
Exclude JNA and tests from deprecations report

akurtakov:


akurtakov:
There is https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/actions/runs/27527287198 :
```

Invalid workflow file: .github/workflows/pr-checks.yml#L19
The workflow is not valid. .github/workflows/pr-checks.yml (Line: 19, Col: 3): Error calling workflow 'eclipse-platform/eclipse.platform.releng.aggregator/.github/workflows/checkMergeCommits.yml@0850457b1332202f30c2b21aed76d07a355b57c6'. The nested job 'message-check' is requesting 'pull-requests: read', but is only allowed 'pull-requests: none'.
```

PR #3907, opened 2026-06-17
Remove formatter parameter in EclipseTestRunner

akurtakov:
The value was simply ignored and it was hiding the test-output parameter setting due to convoluted handling..


akurtakov:
@HannesWell Would you please test that this one works as expected?

HannesWell:
> @HannesWell Would you please test that this one works as expected?

Yes, I can do that tomorrow, but it's currently not easy as one first has to produce a new unofficial I-build in a side location and make the test use it. But it's doable.

Do you have a specific test-suite in mind to run with this?

akurtakov:
Nothing specific. I'm interested that tests still run and unit tests reports do not change - so probably a full build would be best.

HannesWell (approved):
My testing concluded and I can say at least that this doesn't break the execution.
From that testing I unfortunately can't see the final result files, but the printouts indicate that it should be fine.

Thanks for this.

Issue #3908, opened 2026-06-17
Doc bundle size


vogella:
I recently removed the ICU bundle from the Eclipse Platform, reducing the Eclipse SDK download by ~15MB.
While updating a client IDE in a slow environment (v-client) I noticed the docs bundles are huge.

JAR sizes of the shipped doc bundles:

| Bundle | JAR size |
|---|---|
| `org.eclipse.platform.doc.isv` | 42M |
| `org.eclipse.jdt.doc.user` | 9.1M |
| `org.eclipse.jdt.doc.isv` | 7.9M |
| `org.eclipse.pde.doc.user` | 6.0M |
| `org.eclipse.platform.doc.user` | 3.8M |

The working trees are far larger (600M+ for `platform.doc.isv`) but git-ignored.

For example, `org.eclipse.platform.doc.isv`:
the JAR is 42M compressed (~254M uncompressed) and is almost all generated Javadoc.

| HTML category | Uncompressed | Files | Share |
|---|---|---|---|
| Generated Javadoc (`reference/api/`) | 230.2M | 11,049 | 98.4% |
| Other HTML| 3.8M | 696 | 1.6% |

So ~90% of the entire uncompressed JAR is generated Javadoc.
The same concentration applies to `org.eclipse.jdt.doc.isv`.

How is API info used to my knowledge:

* AFAIK PDE uses the bundled Javadoc for hover/F1 only when the matching source bundle is *not* attached. This is offline; AFAIK if online it could point  to online location. The full SDK ships source bundles, so editor hover/F1 renders Javadoc from these bundles. For that setup the bundled HTML is redundant.
* The same reference is online at help.eclipse.org
* Maybe the generated API is used for the API baseline comparison (not sure about this)

Possible directions:

* Ship the API Javadoc as a separate, optional feature instead of bundling it by default.
* Point `org.eclipse.pde.core.javadoc` at the online reference instead of a bundle if that is possible

I do not plan to work on this myself, but AFAIK @HannesWell is actively optimizing the build and download size, so this might be of interest to him.

iloveeclipse:
Maybe we can keep org.eclipse.platform.doc.isv & org.eclipse.jdt.doc.isv in the SDK p2 repository but not add them to the SDK binary. So webhelp (or any other product) can include the documentation bundles from p2 update site and SDK itself doesn't need to include them? I personally never use API Javadocs from inside Eclipse, but google search time to time points to some reference pages.

[...]

PR #3917, opened 2026-06-26
Remove unused Batik from SDK target platform

vogella:
Nothing in the platform imports `org.apache.batik.*` anymore: the e4 CSS engine moved to its own internal parser and SVG rendering now uses JSVG, and `org.apache.batik.css` is not shipped in the SDK product. This drops the four Batik bundles from the prereqs target along with `org.apache.xmlgraphics`, which was only consumed by `batik.css`, plus the matching stale requirement in the doc.isv build.

`org.eclipse.orbit.xml-apis-ext` is intentionally kept: it provides the SAC API (`org.w3c.css.sac`) that `org.eclipse.pde.spy.css` still imports and that ships in the product. Its target comment is updated to reflect that, since it was previously labelled a Batik dependency.

### DIFF
--- a/eclipse.platform.common/bundles/org.eclipse.platform.doc.isv/pom.xml
+++ b/eclipse.platform.common/bundles/org.eclipse.platform.doc.isv/pom.xml
@@ -125,11 +125,6 @@
                   </requirement>
                   <requirement>
                     <type>eclipse-plugin</type>
-                    <id>org.apache.batik.css</id>
-                    <versionRange>0.0.0</versionRange>
-                  </requirement>
-                  <requirement>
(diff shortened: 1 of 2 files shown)


vogella:
@HannesWell @akurtakov WDYT?

akurtakov:
Let's do it.

vogella:
icu gone, Batik gone, sac is planned to be gone soon. If we implement also https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/issues/3908 our SDK download size should be significantly smaller.

merks:
Less is more. 😀

PR #3922, opened 2026-06-30
R4_35_maintenance pin cbi-ecj-version to 3.39.0

MohananRahul:
Pin the version to the ecj 3.39.0 to ensure reproducible builds.


noopur2507 (approved):
Looks good.

PR #3925, opened 2026-07-01
Remove unused SAC (xml-apis-ext) from the target platform

vogella:
The e4 CSS engine dropped SAC long ago, and its last consumer, org.eclipse.pde.spy.css, no longer references org.w3c.css.sac in code (the dead CSSParseException catch was removed in eclipse-pde#2385). The org.eclipse.orbit.xml-apis-ext bundle was pulled into the target solely to provide that API, so this drops it from the SDK prereqs target, from the doc.isv extraRequirements that forced it into the Javadoc build, and from the deploy-to-maven mappings. Removing a dependency that nothing uses keeps the target platform smaller and the build a little faster. The pde.spy.css Import-Package is cleaned up separately.

### DIFF
--- a/eclipse.platform.common/bundles/org.eclipse.platform.doc.isv/pom.xml
+++ b/eclipse.platform.common/bundles/org.eclipse.platform.doc.isv/pom.xml
@@ -220,11 +220,6 @@
                   </requirement>
                   <requirement>
                     <type>eclipse-plugin</type>
-                    <id>org.eclipse.orbit.xml-apis-ext</id>
-                    <versionRange>0.0.0</versionRange>
-                  </requirement>
-                  <requirement>
-                    <type>eclipse-plugin</type>
                     <id>org.eclipse.update.configurator</id>
                     <versionRange>0.0.0</versionRange>
                   </requirement>
--- a/eclipse.platform.releng.prereqs.sdk/eclipse-sdk-prereqs.target
+++ b/eclipse.platform.releng.prereqs.sdk/eclipse-sdk-prereqs.target
@@ -18,9 +18,6 @@
       <!-- This version contains with notarized *.jnilib -->
       <unit id="com.sun.jna" version="5.19.1.v20260612-1000"/>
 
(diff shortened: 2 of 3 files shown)


merks:
I thought maybe this would work now that the PDE change to remove the package import is merged.

merks:
No, I guess the PDE changes need to be in an I-build.   I though the PDE change was merged already.  Hard to keep track of everything.

vogella:
> No, I guess the PDE changes need to be in an I-build. I though the PDE change was merged already. Hard to keep track of everything.

PDE change was merged today by @akurtakov, that is why this is still in draft, I undraft it tomorrow.

merks:
I know patience is a virtue, but I kicked off an I-build now because I'd really like to see the impact of the changes sooner rather than later.. In particular I hope that no downstream SimRel projects are depending on the Platform providing things it will no longer be providing.

vogella:
> I know patience is a virtue, but I kicked off an I-build now because I'd really like to see the impact of the changes sooner rather than later.. In particular I hope that no downstream SimRel projects are depending on the Platform providing things it will no longer be providing.

Sounds good to me, thanks. I try to be more patient with my PRs but I like to see others pushing them fast. :-)

vogella:
Undrafted so that @merks can merge if he like.

vogella:
The failing "License vetting status check / call-license-check / check-licenses (pull_request)" does not make sense to me (we removed a lib, why should that fail the licence check), but I do not have the knowledge to investigate licence check failures.

merks:
I think the license check is a "fast check" and it does the checking of target platform resolution relative to the already-built-bundles so needs an up-to-date I-build.  I think anyway.  😁

merks:
I tried rebasing this time. I think it must literally be using an old state of the git repositories.  And yes, that looks better for the license test...

merks:
Thanks @vogella for the simplifications and for follow up in with target platform too!

vogella:
> Thanks @vogella for the simplifications and for follow up in with target platform too!

Thank you for your support and the positive feedback.

PR #3929, opened 2026-07-05
Fix findMilestoneNumber returning milestone object instead of number

MohananRahul:
findMilestoneNumber was returning the whole milestone Map returned by findElement() instead of extracting its `number` field. When interpolated into the PATCH URL in updateMilestone, Groovy's default Map.toString() rendered it as `[url:, id, number:79,]`, which curl then misinterpreted as glob syntax, failing with: curl: (3) bad range in URL position N.

### DIFF
--- a/JenkinsJobs/shared/githubAPI.groovy
+++ b/JenkinsJobs/shared/githubAPI.groovy
@@ -80,7 +80,7 @@ def findMilestoneNumber(String orgaRepo, String title) {
 	if (milestone == null) {
 		error "Milestone '${title}' not found among the most recent ~1000 milestones"
 	}
-	return milestone
+	return milestone.number
 }
 
 private Object findElement(String query, Closure predicate) {


HannesWell (approved):
Looks good. 
Thanks for the fix and sorry for the trouble!

This issue is why strictly typed languages are better...

MohananRahul:
> Looks good. Thanks for the fix and sorry for the trouble!
> 
> This issue is why strictly typed languages are better...

No problem @HannesWell 
PR build failure seems due to https://www.eclipsestatus.io issues.
and I suspect the another problem is that the renameBuildDrop target directory (S-4.41M1-202607031319) already exists from an earlier promotion run. Removing it and running the job again should fix it.

HannesWell:
All changes to these Jenkins jobs are (sadly) not verified in PR builds.

> and I suspect the another problem is that the renameBuildDrop target directory (S-4.41M1-202607031319) already exists from an earlier promotion run. Removing it and running the job again should fix it.

That could be a problem, but I currently don't have computer access and therefore can't verify or fix this.

If you need to delete the folders I suggest to create a new Jenkins pipeline job and fill the pipeline with the `hello world` example and then adjust the step to start an SSH agent and then to delete the problematic folder via ssh.

When I have computer access again, I can check to make the promotion pipeline more robust in this regard and maybe to add means to enforce the deletion of specific folders to one of the cleanup jobs.

MohananRahul:
> If you need to delete the folders I suggest to create a new Jenkins pipeline job and fill the pipeline with the `hello world` example and then adjust the step to start an SSH agent and then to delete the problematic folder via ssh.

Yes, I had been attempting to complete this for the past two days, but due to service outages I was unable to do so. It has now been completed successfully. Thanks.

PR #3960, opened 2026-08-05
Run API tools check and version baseline comparison only on Linux

### LINKED ISSUE #3959: Run API tools check and baseline version comparison only on Linux in GitHub validation
`.github/workflows/mavenBuild.yml` passes `-Papi-check` and
`-Dcompare-version-with-baselines.skip=false` unconditionally, and the `build` job
is a 3-way matrix (Linux, Windows, macOS).
Both checks therefore run three times per pull request.

API-checks are platform-independent: API tools compares class files against a p2 baseline,
and the baseline version comparison compares built artifacts against the previous
release repository.

Proposal: gate both flags on the Linux matrix entry, for example
`${{ matrix.config.name == 'Linux' && '-Papi-check' || '' }}`.
This cuts two thirds of that work with no loss of coverage.

This especially helps as Mac and Windows verification runs are much longer than Linux.

vogella:
The validation build runs on a Linux/Windows/macOS matrix, but `-Papi-check` and `-Dcompare-version-with-baselines.skip=false` were passed unconditionally, so both checks ran three times per pull request.

Both are platform-independent: API tools compares compiled class files against a p2 baseline, and the version baseline comparison compares built artifacts against the previous release repository. The Windows and macOS runs therefore produce the same result as the Linux run and add no signal. Gating both on the Linux matrix entry removes that duplicated work and shortens the two slowest matrix legs, which is where it helps most.

Fixes https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/issues/3959

### DIFF
--- a/.github/workflows/mavenBuild.yml
+++ b/.github/workflows/mavenBuild.yml
@@ -81,9 +81,9 @@ jobs:
       run: >-
         ${{ runner.os == 'Linux' && 'xvfb-run' || '' }}
         mvn --show-version --update-snapshots --errors
-        -Dcompare-version-with-baselines.skip=false
+        ${{ runner.os == 'Linux' && '-Dcompare-version-with-baselines.skip=false' || '' }}
         -Pbree-libs
-        -Papi-check
+        ${{ runner.os == 'Linux' && '-Papi-check' || '' }}
         --fail-at-end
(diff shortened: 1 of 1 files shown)


akurtakov (approved):
Sounds good. Infra is never free so we shouldn't act like that.

PR #3962, opened 2026-08-05
[RelEng] Fix stale build result paths in README

vogella:
The README still points to `eclipse.platform.releng.tychoeclipsebuilder/eclipse.platform.repository/target/products` for the output of a local SDK build. That project no longer exists, its last remaining script was moved to `scripts` in e519ac96ec, and the README was the only place still referencing the name.

Replaced it with the directories the Tycho build actually writes to: the per-platform product archives under `products/eclipse-sdk` and `products/eclipse-platform`, and the p2 update site under `sites/eclipse-platform-repository`. These match the paths already used in the Jenkinsfile archive step.

### DIFF
--- a/README.md
+++ b/README.md
@@ -37,7 +37,9 @@ git submodule update
 mvn clean verify  -DskipTests=true
 
 # find the results in
-# eclipse.platform.releng.tychoeclipsebuilder/eclipse.platform.repository/target/products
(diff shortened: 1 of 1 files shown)


akurtakov:
As this is readme change it will not be verified. Merging.

PR #3963, opened 2026-08-05
Provision the PDE schema converter from the stream being built

vogella:
The schema-to-html goal of tycho-document-bundle-plugin provisions a headless Eclipse containing org.eclipse.pde.core in order to run ConvertSchemaToHTML. With no pdeToolsRepository configured, Tycho falls back to a repository URL hardcoded in TychoConstants, currently the 2025-12 simultaneous release. The converter was therefore taken from an unrelated release rather than from the stream being built, and drifts further out of date with every release until someone bumps that constant.

Configure pdeToolsRepository in eclipse-platform-parent to ${eclipse-sdk-repo}, next to tycho-eclipse-plugin, which already pins the same repository for the same purpose. All three consumers of schema-to-html inherit this parent.

### DIFF
--- a/eclipse-platform-parent/pom.xml
+++ b/eclipse-platform-parent/pom.xml
@@ -578,6 +578,18 @@
           </executions>
         </plugin>
         <plugin>
+          <groupId>org.eclipse.tycho.extras</groupId>
+          <artifactId>tycho-document-bundle-plugin</artifactId>
+          <version>${tycho.version}</version>
+          <configuration>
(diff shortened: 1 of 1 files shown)


vogella:
From memory: The additional download took ~ 2 mins in the PDE verification flow. This will be gone with this this PR

akurtakov:
Tycho master/5.0x branch should use 2026-06 now (https://github.com/eclipse-tycho/tycho/pull/6213) thanks for spotting it.  SDK build using it's own repo is good fix here too.

PR #3965, opened 2026-08-06
[RelEng] Add a startup time benchmark for the Eclipse IDE

vogella:
The platform already instruments IDE startup, but nothing turns it into a repeatable measurement. `UIStats` prints `Startup complete: <n>ms` only when the platform runs with `org.eclipse.core.runtime/debug=true`, and since the results database behind `org.eclipse.test.performance` was retired there is nothing to compare a number against. The only startup benchmark in the tree, `EmptyWorkbenchPerfTest`, measures an empty RCP workbench rather than the IDE.

`StartupTimeBenchmark` launches a given Eclipse installation repeatedly against the same workspace and reports the median wall-clock and CPU time, along with the observed spread so that a measurement taken on too busy a machine is visible as such rather than being quietly believed. Optional CSV output makes the results archivable and plottable from a build job.

Both wall-clock and CPU time are reported because they fail in opposite ways. Wall-clock time is what users perceive but also counts everything else the machine was doing, so it degrades badly on a shared agent. CPU time is insensitive to that competition but cannot see I/O waits or work moved onto background threads, and it spreads wider run to run because it sums all threads.

One trap is worth calling out, because measuring it wrong is the default. A benchmark has to stop each launch, so the launch never shuts down in an orderly way and `ExtensionRegistry.stop()` never runs, which means the extension registry cache is never written. Every launch then re-parses all `plugin.xml` files. Three interleaved A/B pairs on an SDK, alternating the arms so that drift in machine load hits both equally:

| | registry activator | all activators | total |
|---|---|---|---|
| no cache | 1245 ms | 3250 ms | 7649 ms |
| primed cache | 14 ms | 1531 ms | 4890 ms |

So the benchmark primes the installation with `eclipse -initialize` and prints whether the cache is in place. `--no-initialize` measures the re-parse path deliberately, which is what users get on the first start after an install or an update.

It is a plain source-code program under `scripts` with no dependencies beyond the JDK, so it changes no product code and nothing in the build.

### DIFF
--- a/scripts/README.md
+++ b/scripts/README.md
@@ -4,3 +4,5 @@ This folder contains scripts that are used in the Eclipse build, mainly for RelE
 
 The contained `Java` classes are intended to be launched as single- or mutli-file _Source-Code Programs_ respectively as plain _Java scripts_.
 This means the main source file is launched directly without being compiled and therefore this project must not have any external dependencies except the JDK.
+
+`releng/StartupTimeBenchmark.java` measures how long an installed Eclipse IDE takes to start, see the script for usage.
--- a/scripts/releng/StartupTimeBenchmark.java
+++ b/scripts/releng/StartupTimeBenchmark.java
@@ -1,0 +1,330 @@
+/*******************************************************************************
+ *  Copyright (c) 2026 Vogella GmbH and others.
+ *
+ *  This program and the accompanying materials
+ *  are made available under the terms of the Eclipse Public License 2.0
+ *  which accompanies this distribution, and is available at
+ *  https://www.eclipse.org/legal/epl-2.0/
+ *
+ *  SPDX-License-Identifier: EPL-2.0
+ *
+ *  Contributors:
+ *     Lars Vogel - initial API and implementation
+ *******************************************************************************/
+
+/// Measures how long an installed Eclipse IDE takes to start, repeatedly, and
+/// reports wall-clock and CPU time per run.
+///
+/// ```
+/// xvfb-run -a java StartupTimeBenchmark.java --eclipse <launcher> --runs 10 --csv results.csv
+/// ```
+///
+/// Options:
+///
+/// - `--eclipse <path>` -- Eclipse launcher executable (required)
+/// - `--runs <n>` -- number of launches, default 10
+/// - `--workdir <dir>` -- scratch directory, default a temporary directory
+/// - `--csv <file>` -- additionally write the results as CSV
+/// - `--timeout <sec>` -- per-launch timeout, default 300
+/// - `--no-initialize` -- skip priming, see below
+///
+/// A display is required. On a headless Linux machine wrap the whole benchmark,
+/// not the individual launches, as shown above.
+///
+/// ## What is measured
+///
+/// Wall-clock time is read from the `Startup complete: <n>ms` line that
+/// `org.eclipse.ui.internal.misc.UIStats` prints when the platform runs with
+/// `org.eclipse.core.runtime/debug=true`. It spans from the timestamp the native
+/// launcher records in `eclipse.startTime` until the workbench event loop starts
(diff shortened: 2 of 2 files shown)


scripts/README.md
```diff
@@ -4,3 +4,28 @@ This folder contains scripts that are used in the Eclipse build, mainly for RelE
...
 This means the main source file is launched directly without being compiled and therefore this project must not have any external dependencies except the JDK.
+
+## Startup time benchmark
+
+`releng/StartupTimeBenchmark.java` measures how long an installed Eclipse IDE takes to start.
+It launches the given Eclipse repeatedly against the same workspace and reports median wall-clock and CPU time.
+
+```
+xvfb-run -a java StartupTimeBenchmark.java --eclipse <path-to-eclipse-launcher> --runs 10 --csv results.csv
+```
```
HannesWell: Implementation and usage details should not be located in this README, instead that should be in the script itself.
At most the existence of the script should be mentioned here with a very rough summary.

scripts/releng/StartupTimeBenchmark.java
```diff
@@ -0,0 +1,344 @@
...
+ * what users get on the first start after an install or an update.
+ * <p>
+ * A display is required. On a headless Linux machine wrap the whole benchmark,
+ * not the individual launches:
+ *
+ * <pre>
+ * xvfb-run -a java StartupTimeBenchmark.java --eclipse /path/to/eclipse --runs 10
+ * </pre>
+ */
+public class StartupTimeBenchmark {
```
HannesWell: Since this script can target Java-25 (or later) I suggest to use the latest and greatest java features like Compact Source files and instance main methods, etc:

- https://docs.oracle.com/en/java/javase/25/language/compact-source-files-instance-main-methods.html#GUID-35544A22-61AB-4928-99BB-A9DD1CA062FF

If you prefer it, you can also use markdown javadoc.

You can use the `BuildDropDataGenerator.java` as an example.

HannesWell (commented):
This is nice to have.

> No Jenkins job is included. Where such a job should run, and against which build, seems worth deciding separately from landing the tool.

If we should run this regularly in a Jenkins job is to be discussed ( in general, at least for IDE users, I assume startup time is often not the most important performance measure. At least for the performance during the work is much more important). In case we generally revive performance tests, this should certainly discussed.

Nevertheless, if two remarks below.

vogella:
Thanks for the feedback.

We can already use Java 25 für Eclipse?

scripts/README.md
vogella: Moved. The README now only names the script and points at it, everything else lives in the doc comment on `main`.

scripts/releng/StartupTimeBenchmark.java
vogella: Converted to a compact source file with an instance `main`, following `BuildDropDataGenerator.java`, and switched the doc comments to markdown javadoc. The implicit `java.base` import means the file now needs no imports at all. Verified it still runs: 3 launches against a locally built SDK, median 2991 ms wall / 9855 ms CPU, CSV output unchanged.

scripts/README.md
HannesWell: That's better. Thanks.

But IMO the mentioning of the new script should be in a dedicated section to make the context more clear, e.g. named `Descriptions` or `Scrip descriptions` or maybe `Summary`.
Could also prepend a dash before the description  so we get a listing that could be extended later, if suitable.

HannesWell:
> We can already use Java 25 für Eclipse?

It depends. For Code of the platform/IDE: Not yet, but soon.
For RelEng tasks to build and test Eclipse, of course yes (as long as it e.g. doesn't interfere with the actual test runtime).
In fact most of the builds already runs on Java-25 and some of the RelEng scripts already use the latest Java-25 features.
If it helps we could use even more recent Java versions to build Eclipse.

PR #3857, opened 2026-05-29
[RelEng] Create RelEng calendar entries in preparation pipeline

HannesWell:
The creation of corresponding credentials is still pending, but already requested in
- https://gitlab.eclipse.org/eclipsefdn/helpdesk/-/issues/7532

Additionally some more methods used in the pipeline have to be approved first, which is requested in
- https://gitlab.eclipse.org/eclipsefdn/helpdesk/-/issues/7546

PR #3974, opened 2026-08-11
[RelEng] Add self-managed RelEng/build calendar to overview site

HannesWell:
The new calendar is
- simpler to discover and at a more prominent location
- easier to maintain, especially automatically, but also everyone can
contribute further improvements like code changes
- Events have more detailed descriptions
- Y-builds are contained in the calendar too

Furthermore the logic of build schedules, used for Jenkins cron triggers and the calendar, is implemented at one central location in the script style `releaseSchedule.java` Java file.

This is an alternative to https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/pull/3857, which is a dead-end:
- https://gitlab.eclipse.org/eclipsefdn/helpdesk/-/work_items/7532#note_8156710


HannesWell:
This is now ready for submission.

@MohananRahul this means for the preparation of the next development cycle the existing calendar doesn't have to be updated anymore. This PR updates the Release doc accordingly.

I suggested that we delete the old calendar soon, once the new one is established.

PR #3982, opened 2026-08-18
[RelEng] Generalize and unify Jenkins job folder names

HannesWell:
Make the folder name less specific and use lower case with dash-separation consistently.
This matches names of generated jobs like builds and tests so that all segments of the job-id have a consistent style.

This also results in nicer URLs for jobs.

To get even nicer URLs, the jenkins files could be changed to use dash to separate words instead of CamelCase.
But that leads to file names that seem uncommon to me (for folders it's semi common).
It would also be possible to have different job id casing (which determine the URL) and file name casing. But I think it's more important to have the job ID to match the file name (and parent folder) exactly. 
Edit: I've now changed the Jenkins files to use dash-case too.


HannesWell:
Submitting this now despite the freeze-period since this is a RelEng only change which is best to be used in the upcoming release phase.

PR #3984, opened 2026-08-20
Update Y-build tests to run on latest OpenJDK JDK 27 Release-Candidate

HannesWell:
The current RC is Build 35 (2026/8/20).

@jarthana, @mpalat are you interested in that update now, so the final builds can run on the latest RC?


mpalat:
@HannesWell Thanks for the pro-active check. Yes, 35 would be good since it's RC.

PR #3996, opened 2026-08-25
Update the target platform to the Orbit 4.41.0 release update site

merks:


merks:
This is a necessary change for the release in order to support the maintain branch forked from master.  Moreover, it doesn't change the contents of the target platform but rather provides a permanently stable repository URL.

PR #3997, opened 2026-08-26
[RelEng] Automate creation of Migration guide in release preparation

HannesWell:
Automate the creation of Migration guide in release preparation pipeline.

This automates changes like the following during release preparation
- https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/pull/3991

Furthermore this updates the Release documentation to consider the automation added here and in
- https://github.com/eclipse-platform/www.eclipse.org-eclipse/pull/601

The automation of the corresponding JDT change is added in
- https://github.com/eclipse-jdt/eclipse.jdt/pull/163

PR #4000, opened 2026-08-27
[RelEng] Limit included porting guides to twelve and fix JDT porting TOC

HannesWell:
In order to prevent an unrestricted growth, limit the number of included porting guides in the org.eclipse.platform.doc.isv and org.eclipse.jdt.doc.isv bundles to twelve. This corresponds to three years of releases.
The last four guides are considered older, the first eight are 'recent'. The latter corresponds to two years of releases, which matches the deprecation policy of the Eclipse SDK.

The deletion performed in this PR would also happen during the release preparation after the submission of
- https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/pull/3997

In order to avoid additional changes, this is extract together with the mentioned TOC fix.
In the future, the number of porting guides will be limited to twelve and the deletion of older entries happens automatically when new entries are added during release preparation.


HannesWell:
This is a releng related change and will only appear in the builds of the upcoming 4.42 development cycle, which will be prepared soon.

PR #4009, opened 2026-09-02
Update GHAs to support Java 25

akurtakov:
Prereq for https://github.com/eclipse-equinox/equinox/pull/1348


akurtakov:
Failing Jenkins is unrelated to GHAs change. Merging.

PR #4010, opened 2026-09-02
Move targetJRE to Java 25

akurtakov:
It's a prereq to allow using Java 25 in
https://github.com/eclipse-equinox/equinox/pull/1348 .

### DIFF
--- a/eclipse.platform.releng.prereqs.sdk/eclipse-sdk-prereqs.target
+++ b/eclipse.platform.releng.prereqs.sdk/eclipse-sdk-prereqs.target
@@ -947,5 +947,5 @@
 		  </dependencies>
 	  </location>
   </locations>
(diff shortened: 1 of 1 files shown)


akurtakov:
My plan is to move builds and verifications in all the repos before merging this one sometime next week. @HannesWell any concerns?

HannesWell:
I had the intention (and already worked on it locally) to do this as part of a greater move of all missing pieces to Java-25 (like tests and more). I intended to do that soon after the preparation of the new cycle but until now was distracted by the hectic after the new cycle started.
If you don't mind, I'd just include this change then.

akurtakov:
I'm always fine with someone else doing the work :).

merks:
Oomph setups will need updating.

HannesWell:
This was included in
- https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/pull/4015

Thus this is effectively empty and I'm closing this. Nevertheless thanks for your help.

> Oomph setups will need updating.

For cross references, that was done in
- https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/pull/4011

PR #4014, opened 2026-09-03
Enable JVM native access

akurtakov:
Silences warnings on startup like
```
WARNING: Use --enable-native-access=ALL-UNNAMED to avoid a warning for
callers in this module
WARNING: Restricted methods will be blocked in a future release unless
native access is enabled
```
caused by https://openjdk.org/jeps/472

### DIFF
--- a/products/eclipse-platform/platform.product
+++ b/products/eclipse-platform/platform.product
@@ -15,6 +15,7 @@
 -Dorg.slf4j.simpleLogger.defaultLogLevel=off
 -Declipse.platform.mergeTrust=true
 --add-modules=ALL-SYSTEM
+--enable-native-access=ALL-UNNAMED
       </vmArgs>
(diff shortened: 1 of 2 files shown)


akurtakov:
No dependency changes in this PR and the failure is "[INFO] Querying Eclipse Foundation for license data for 500 items.
  [ERROR] Error response from the Eclipse Foundation HTTP 504
" . Merging.

products/eclipse-platform/platform.product
```diff
@@ -15,6 +15,7 @@
 -Dorg.slf4j.simpleLogger.defaultLogLevel=off
 -Declipse.platform.mergeTrust=true
 --add-modules=ALL-SYSTEM
+--enable-native-access=ALL-UNNAMED
```
HannesWell, line 18: I was wondering if we should make this more specific? But then we'd have to list all the bundles that use JNI or FFM and all installed external third-party plugins would have to do the same.

akurtakov: I don't believe we will have the chance to remove that at all with current knowledge.

PR #4017, opened 2026-09-03
[RelEng] Make RC test in freeze check more specific

### LINKED ISSUE #4013: Unexpected code freeze detected in JDT core
See the discussion here: https://github.com/eclipse-jdt/eclipse.jdt.core/pull/5322#issuecomment-5516159528

It looks like this milestone is not parsed quite right:

```
Checking milestone BETA_JAVA27_RC2
Error: Active freeze period for 
```

And so the check reported a code freeze when there was none.

HannesWell:
Makes the test if a milestone is a common Eclipse RC milestone more specific to avoid false positive freeze checks, for example on JDT's `BETA_JAVA `RC milestones.

Fixes https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/issues/4013


HannesWell:
Tests looked good and the remaining verifications build doesn't test this. Submitting.

PR #4028, opened 2026-09-05
Remove o.e.ant.optional.junit

### LINKED ISSUE #4022: Remove the org.eclipse.ant.optional.junit fragment from the Eclipse Test Framework
## Summary

`org.eclipse.ant.optional.junit` is a code-free fragment of `org.apache.ant` in `eclipse.platform.releng/bundles`. Its only content is `Require-Bundle: org.junit`.
It was created so that Ant's JUnit task and result formatters, which the Eclipse Test Framework (`org.eclipse.test`) drives, could see JUnit classes. That job is obsolete and the fragment can be removed.

## Why it is no longer needed

- The Orbit `org.apache.ant` bundle (1.10.17 in the current target) declares  optional `Import-Package` entries for `junit.framework`, `org.junit.*` and all  `org.junit.platform.*` packages. PDE resolves every one of them against the  target platform, so Ant reaches JUnit 4 and the JUnit Platform on its own.
- The Eclipse test runner has driven tests through Ant's `junitlauncher` since  JUnit 5 support was added in 2018 (bug 531057). The fragment never provided  any JUnit Platform package, so for years Ant has been resolving JUnit 5  through its own imports. The JUnit 4 packages resolve the same way.
- `org.eclipse.test` requires `org.junit` directly, so JUnit 4 is installed into  a test installation with or without the fragment.
- The fragment therefore adds one redundant wire to Ant and nothing else.

akurtakov:
Eclipse test runner gets JUnit 5.x dependencies in another ways as described already so this bundle is simply remnant from the past. Fixes https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/issues/4022


production/testScripts/configuration/sdk.tests/testScripts/test.xml
```diff
@@ -448,7 +448,7 @@
       <arg value="tooling" />
       <arg value="-installIUs" />
       <arg
-        value="${testPlugin},org.eclipse.test,org.eclipse.ant.optional.junit,org.eclipse.test.performance,org.eclipse.test.performance.win32" />
+        value="${testPlugin},org.eclipse.test.performance,org.eclipse.test.performance.win32" />
```
HannesWell: Is the removal of the `org.eclipse.test` really wanted here? I don't see another way how it would be installed without this, but maybe I'm missing something.
```suggestion
        value="${testPlugin},org.eclipse.test,org.eclipse.test.performance,org.eclipse.test.performance.win32" />
```

HannesWell (changes requested):
Generally this looks good, but the removals in `testScripts/test.xml‎` seem too much.

production/testScripts/configuration/sdk.tests/testScripts/test.xml
akurtakov: I have deleted too much. Fixed now.

HannesWell (approved):
Looks good. Thank you.

Issue #4031, opened 2026-09-07
Publish 4.41 to Maven central


elsazac:

elsazac:
@HannesWell @MohananRahul 

I only see JDT and PDE under deployments. Could you check the Jenkins Job?
<img width="481" height="602" alt="Image" src="https://github.com/user-attachments/assets/f351c473-5db5-4b3f-95f6-b5815b360f69" />

merks:
Maybe we should run another build at 

https://ci.eclipse.org/releng/job/releng/job/deploy-maven/build?delay=0sec

with parameter

https://download.eclipse.org/eclipse/updates/4.41/R-4.41-202608281142/

and monitor what the publishing step does...

elsazac:
https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/issues/3951#issuecomment-5565961020

HannesWell:
For now I've replayed the build with just the platform branch enabled and (with the maximal possible polling interval)
- https://ci.eclipse.org/releng/job/releng/job/deploy-maven/465/replay/diff

See also larger the work to further increase the polling intervals:
- https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/pull/4029

If that fails again, I'd replay it again and disable the status check. It can still be done in the UI.

HannesWell:
> For now I've replayed the build with just the platform branch enabled and (with the maximal possible polling interval)
> 
>     * https://ci.eclipse.org/releng/job/releng/job/deploy-maven/465/replay/diff

That run was successful, @elsazac please check the Central Portal:
```
 Upload to Maven Central Portal successful:
 {
     "deploymentId": "4a28d373-1704-4223-9c32-5271b593f8d3",
     "deploymentName": "PLATFORM_R-4.41-202608281142",
     "deploymentState": "VALIDATED",
     "purls": [
         "pkg:maven/org.eclipse.platform/org.eclipse.core.jobs@3.15.900",
         "pkg:maven/org.eclipse.platform/org.eclipse.terminal.connector.telnet@1.1.100",
         "pkg:maven/org.eclipse.platform/org.eclipse.swt.gtk.linux.ppc64le@3.135.0",
         "pkg:maven/org.eclipse.platform/org.eclipse.tips.ui@0.4.200",
         "pkg:maven/org.eclipse.platform/org.eclipse.e4.core.di.extensions.supplier@0.17.1200",
[...]

PR #4033, opened 2026-09-07
Add projectsToDeploy parameter to selectively deploy platform/jdt/pde

MohananRahul:
Adds a projectsToDeploy choice parameter (all / platform / jdt / pde, defaulting to all)

### DIFF
--- a/jobs/releng/deploy-maven.jenkinsfile
+++ b/jobs/releng/deploy-maven.jenkinsfile
@@ -39,6 +39,8 @@ pipeline {
 			</li>
 			</ul>
 		''')
+		choice(name: 'projectsToDeploy', choices: ['all', 'platform', 'jdt', 'pde'],
+			description: 'Which project(s) to deploy to Maven. "all" runs platform, jdt, and pde.')
 	}
 	agent {
 		label 'basic'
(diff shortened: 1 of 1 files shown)


jobs/releng/deploy-maven.jenkinsfile
```diff
@@ -128,6 +130,12 @@ pipeline {
 				}
 				stages {
 					stage('Deploy project to Maven') {
+					when {
+							anyOf {
+								expression { params.projectsToDeploy == 'all' }
+								expression { params.projectsToDeploy == env.PROJECT }
+							}
+						 }
```
HannesWell, line 138: You probably also have to consider the null case here, in case this job is invoked without that parameter set (like it's currently done from I-builds).
Furthermore, since you are already in a expression context, you could just use the 'Java' style and or the expressions together to make it a bit more compact:
```suggestion
						when {
							expression { params.projectsToDeploy == null || params.projectsToDeploy == 'all' || params.projectsToDeploy == env.PROJECT }
						}
```
And ideally please make sure to use consistent indentation (tabs only).

HannesWell (commented):
Thank you Rahul for that one.
Most of it looks good, I just have one point below.

However generally I'm not sure we really need this. I hope that the failure that triggered this doesn't happen anymore and in case we selectivly want to deploy one project again, we can still just `Replay` a previous attempt and modify the matrix axis to our needs
https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/blob/2fbdc58233a2c963b5e35a5718abb9710d3f7c95/jobs/releng/deploy-maven.jenkinsfile#L124-L127

like I did in https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/issues/4031#issuecomment-5566328030

Nevertheless thanks again.

PR #4037, opened 2026-09-09
Update to EMF 2.48 milestone

merks:


merks:
@HannesWell  @akurtakov 

I don't think this can be fixed by me....

iloveeclipse:
> I don't think this can be fixed by me....

Do you mean build error? I assume we just need SDK build, see https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/issues/3987#issuecomment-5604079913

HannesWell:
> I don't think this can be fixed by me....

What exactly are you referring to? The latest builds succeeded as far as I can tell.
And I see you rebased on top of the release change and I assume that fixed your build?

merks:
The build before that failed. It was using the deleted  older I build. The one might delete those a day later. Or people could be more patient.

PR #4039, opened 2026-09-10
Define a version for the exec-maven-plugin plug-in in parent pom.xml

trancexpress:
The `exec-maven-plugin` is required by SWT and will soon be required by JDT UI. So define its version in the `eclipse-platform-parent/pom.xml` file.


trancexpress:
3.6.3 is the latest I see here: https://mvnrepository.com/artifact/org.codehaus.mojo/exec-maven-plugin

akurtakov (approved):
Looks good.

PR #4042, opened 2026-09-11
Replace abandoned com.jcraft.jsch with the maintained mwiede fork

akurtakov:
com.github.mwiede.jsch exports the same com.jcraft.jsch package, so no changes are needed in the consumers, when they use Import-Package.

https://github.com/eclipse-platform/eclipse.platform/issues/958

Assisted-by: Anthropic Claude Code (claude-opus-5[1m])


akurtakov:
Merging it in and starting I-build so hopefully there are test results today.

PR #4043, opened 2026-09-14
Remove orphaned tasks-103.htm left over from CVS cleanup

BeckerWdf:
Commit [50e40dba57](https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/commit/50e40dba5739a1ca285e6d84547cb4ddd34d63d4) removed CVS content from the docs in Feb 2022, including the <topic> entry for tasks-103.htm from topics_Tasks.xml and deleting the other CVS task files that linked to it. However, tasks-103.htm itself was only modified (CVS-specific content pruned) but never deleted, leaving it as an unreachable orphan with no TOC entry and no inbound links anywhere in the bundle.


akurtakov (commented):
I trust you to do the right thing here but can't deep dive into it soon.

PR #4046, opened 2026-09-14
Remove outdated screenshots from compare editor and toolbar docs

BeckerWdf:
tasks-68bg.htm: removed a numbered-callout screenshot of the Compare editor and rewrote the surrounding text so the explanation of color coding and difference bands stands on its own.

ttoolbars.htm: removed two outdated toolbar screenshots and rewrote the first step to be clearer: instead of asking the user to "make sure the toolbar is unlocked", the text now directly says to unlock it if locked and how to verify the unlocked state.

help_scope.htm: removed four outdated screenshots and consolidated the surrounding step-by-step captions into prose that stands without the images. Also rewrote the unclear sentence "in the help view help the active scope only affects search results" to make the distinction between the embedded Help view and the standalone help browser explicit.


akurtakov:
One problem is that you're pushing changes in submodules which shouldn't happen.

BeckerWdf:
> One problem is that you're pushing changes in submodules which shouldn't happen.

Sorry. Fixed now.

PR #4057, opened 2026-09-21
I-build: add Java 27 tests 

MohananRahul:
As of this change, Eclipse Temurin has not published a Java 27 build (GA or EA).To unblock Java 27 automated testing in the meantime, this configuration uses OpenJDK for Mac.


merks:
Yes, I'm still waiting for Java 27 release builds from Temurin as well, but there are Java 27 Beta/EA builds available

https://github.com/eclipse-justj/justj/blob/2b1ac9136d0b774d253141b659040ac7ef5d7da8/releng/org.eclipse.justj.releng/Jenkinsfile#L180-L211

And, for what it's worth, Java 28 Beta/EA as well:

https://github.com/eclipse-justj/justj/blob/2b1ac9136d0b774d253141b659040ac7ef5d7da8/releng/org.eclipse.justj.releng/Jenkinsfile#L213-L243

MohananRahul:
> Yes, I'm still waiting for Java 27 release builds from Temurin as well, but there are Java 27 Beta/EA builds available

Thanks. Since type: "temurin" currently requests [GA by default ](https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/blob/master/jobs/shared/utilities.groovy#L276)and Temurin 27 GA isn't available yet, I think we should use type: "install" with the available Java 27 openjdk build for now.

HannesWell:
> > Yes, I'm still waiting for Java 27 release builds from Temurin as well, but there are Java 27 Beta/EA builds available
> 
> Thanks. Since type: "temurin" currently requests [GA by default ](https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/blob/master/jobs/shared/utilities.groovy#L276)and Temurin 27 GA isn't available yet, I think we should use type: "install" with the available Java 27 openjdk build for now.

I'm fine either way. But you also also add the line `"releaseType": "ga",` to the macos configuration. Mayb that's a bit simpler?

Or we just wait until the Temurin Release is GA. I assume it won't take too long until it becomes available?
- https://github.com/adoptium/temurin/issues/301

MohananRahul:
> I'm fine either way. But you also also add the line `"releaseType": "ga",` to the macos configuration. Mayb that's a bit simpler?

Just want to double check I'm not missing something right now getJDKInstallationPath's 'temurin' case only reads test.jdk.ea (test.jdk.ea ? 'ea' : 'ga'), not a releaseType field, so adding "releaseType": "ga" on its own wouldn't change behavior , be ignored, and the pipeline would still default to 'ga' regardless.

Did you mean we should also update the pipeline to read releaseType explicitly instead of the ea boolean? otherwise I think keeping "type": "install" for now, and switching back to "type": "temurin" once #[301](https://github.com/adoptium/temurin/issues/301) shows movement.

HannesWell:
> > I'm fine either way. But you also also add the line `"releaseType": "ga",` to the macos configuration. Mayb that's a bit simpler?
> 
> Just want to double check I'm not missing something right now getJDKInstallationPath's 'temurin' case only reads test.jdk.ea (test.jdk.ea ? 'ea' : 'ga'), not a releaseType field, so adding "releaseType": "ga" on its own wouldn't change behavior , be ignored, and the pipeline would still default to 'ga' regardless.

You're right, my bad. I didn't looked into the code. In fact using Temurin EA builds should be enabled by adding `"ea": "true",`.

> Did you mean we should also update the pipeline to read releaseType explicitly instead of the ea boolean?

No IMO there is no need for a change in that part.

> otherwise I think keeping "type": "install" for now, and switching back to "type": "temurin" once #[301](https://github.com/adoptium/temurin/issues/301) shows movement.

Actually I wanted to suggest to use the Temurin EA builds in the meantime.
But in the meantime Temurin seems to have released JDK 27 GA for macOS aarch64. So now the final configuration can be used.

In general I'm usually in favor of waiting a few days until something missing becomes available instead of introducing temporary solutions, unless its urgent and there's a compelling reasons for doing something immediately.

HannesWell:
> But in the meantime Temurin seems to have released JDK 27 GA for macOS aarch64. So now the final configuration can be used.

@MohananRahul the Temurin JDKs should be available now. Do you want to update this PR so it can be submitted?

